PR #49, opened 2026-05-30
bucket-A3: add core/coin_params.hpp scaffolding (SAFE-ADDITIVE)

frstrtr:
## Summary

Adds the coin-agnostic `core/coin_params.hpp` scaffolding header — the shared description of a coin's consensus/network constants that per-coin layers populate and core consumes. **Purely additive**: one new file, nothing on master `#include`s it yet (audit classification: SAFE-ADDITIVE).

1 file, +133:
- **new** — `src/core/coin_params.hpp`

Cherry-picked from `dash-spv-embedded` HEAD `8249dc28`.

## Context

Part 3 of the bucket-A sub-PRs for the multi-coin landing strategy. Follows #46 (A1, bitcoin_family base) and #48 (A2, enhanced_node template).

Note: the strategy doc originally scoped A3 as 2 files (`coin_params.hpp` + `core/pow.hpp`). `pow.hpp` was absorbed into A1 as a hard transitive dependency of `bitcoin_family/coin/chain_params.hpp`, so A3 shrinks to the single `coin_params.hpp`.

## Why this is safe for LTC+DOGE

- Brand-new file. `git grep` confirms **nothing on master `#include`s `coin_params.hpp`**, so it is never pulled into a translation unit on this branch — zero impact on the LTC/DOGE/legacy build. The plumbing that actually consumes `CoinParams` (LTC + `c2pool_refactored.cpp`) is bucket **A7**, reviewed separately.

## Merge ordering (read before merging)

`coin_params.hpp` `#include`s `core/pow.hpp`, which ships in **A1 (#46)**. This is a *merge-ordering* dependency, not a code dependency on this branch:
- Since nothing `#include`s `coin_params.hpp` on master, it is not compiled here and **CI is green without pow.hpp present**.
- The include resolves the moment A1 merges. Land order is **A1 → A2/A3/A4 back-to-back**, per the plan, so this is satisfied naturally.
- `pow.hpp` is intentionally **not** re-added in this PR — it already ships in A1; duplicating it would create a duplicate-add merge conflict.

## CI expectation

- Legacy `linux:` + cross-platform + CodeQL: **green** (file uncompiled).
- coin-matrix entries: skip fast on this branch.

## Reviewer

Agent-side reviewer: **ltc-doge-production-steward** (watches the branch). Integrator (cc) routes human review.

## Test plan

- [ ] Legacy `linux:` CI green
- [ ] CodeQL + cross-platform green
[...]

PR #48, opened 2026-05-30
bucket-A2: extract enhanced_node into coin-agnostic template (SAFE-ADDITIVE)

frstrtr:
## Summary

Lifts the `EnhancedC2PoolNode` implementation out of the compiled `src/c2pool/node/enhanced_node.cpp` into a header-only, coin-agnostic template `EnhancedC2PoolNodeT<Config, ShareChain, Peer, ShareType>` (new `enhanced_node_impl.hpp`). `enhanced_node.hpp` becomes a thin backward-compat header pinning the template to LTC types, so existing callers compile unchanged. **Pure refactor — LTC behavior identical** (audit classification: SAFE-ADDITIVE).

4 files, +251 / −368:
- **new** — `src/c2pool/node/enhanced_node_impl.hpp` (coin-agnostic template)
- **modified** — `src/c2pool/node/enhanced_node.hpp` (LTC alias shim: `EnhancedC2PoolNode = EnhancedC2PoolNodeT<ltc::...>`)
- **deleted** — `src/c2pool/node/enhanced_node.cpp` (impl moved to template header)
- **modified** — `src/c2pool/CMakeLists.txt` (`c2pool_node_enhanced` library → INTERFACE)

Cherry-picked from `dash-spv-embedded` HEAD `8249dc28`.

## Context

Part 2 of the bucket-A sub-PRs implementing the multi-coin landing strategy (single master + directory contract + coin-matrix CI + per-coin release branches). Follows #46 (A1, bitcoin_family base extraction), still under review — A2 is independent and does not depend on A1.

## Why this is safe for LTC+DOGE

- The LTC alias `EnhancedC2PoolNode` keeps the same name and public API; the only consumer, `src/c2pool/c2pool_refactored.cpp`, is **untouched** by this PR.
- `c2pool_node_enhanced` becomes an INTERFACE (header-only) library. The two executables that link it (`c2pool`, `c2pool_enhanced`) **already link `ltc` directly** (CMakeLists.txt lines 78–90 on master), so dropping `ltc` from the library's transitive deps is a no-op for the link step.
- No file outside `src/c2pool/node/` and `src/c2pool/CMakeLists.txt` is touched.

## Plan deviation (mirrors A1's precedent)

[...]

PR #46, opened 2026-05-27
bucket-A1: extract bitcoin_family base from src/impl/ltc/ (breaking ChainParams API refactor, runtime-behavior-preserving for LTC)

frstrtr:
## Summary

Lifts LTC-duplicated header types out of `src/impl/ltc/coin/*` into the reusable `src/impl/bitcoin_family/coin/*` base; LTC headers become forwarding shims. **Source-level API refactor of `ltc::coin::LTCChainParams`** (now an alias of `bitcoin_family::coin::ChainParams`, with construction moved from static `::mainnet()/::testnet()` factories to free functions `make_ltc_chain_params_mainnet()/_testnet()`). **Runtime behavior for LTC is identical** — the factories return the same field values; only the construction spelling changes. This is a breaking *compile-time* API change, so every consumer and test that built `LTCChainParams` is updated in lockstep here (see Fixups).

25 files, +1001 / −684:
- **7 new** — `src/impl/bitcoin_family/coin/{base_block, base_p2p_messages, base_transaction, chain_params, softfork_check, txidcache}.hpp` + `src/core/pow.hpp`
- **10 modified (headers)** — `src/impl/ltc/coin/{block, header_chain, mweb_builder, p2p_messages, softfork_check, template_builder, transaction, txidcache}.hpp` + `src/c2pool/storage/sharechain_storage.{cpp,hpp}`
- **1 modified (consumer)** — `src/c2pool/c2pool_refactored.cpp` (the `auto ltc_params = ...` construction at the embedded-LTC bring-up)
- **7 modified (tests)** — `test/{test_header_chain, test_phase1_live, test_phase2_live, test_phase3_live, test_phase4_embedded, test_phase4_live, test_template_builder}.cpp`

Cherry-picked from `dash-spv-embedded` HEAD `8249dc28`.

## Fixups since first push (`de59f660`)

The initial push lifted the headers but missed the paired consumer/test migration that the API change forces. Caught by CI + ltc-doge review:

- **`c1cec2ad`** — `c2pool_refactored.cpp:1573-74`: `LTCChainParams::testnet()/::mainnet()` → `make_ltc_chain_params_testnet()/_mainnet()` (2 lines, sole `src/` consumer).
[...]

PR #50, opened 2026-05-30
bucket-A4: cap Packet read-constructor prefix_length (SAFE-ADDITIVE, Bug-9 hardening)

frstrtr:
## Summary

Adds a bounds guard to `Packet(size_t prefix_length)`: caps `prefix_length` at 16 and throws `std::ios_base::failure` if exceeded, instead of letting `prefix.resize()` throw `std::length_error` (`vector::_M_default_append`) and abort the process. **Additive hardening — no change to valid-traffic behavior** (audit classification: SAFE-ADDITIVE).

1 file, +11:
- **modified** — `src/core/packet.hpp`

Cherry-picked from `dash-spv-embedded` HEAD `8249dc28`.

## Context

Part 4 of the bucket-A sub-PRs for the multi-coin landing strategy. Follows #46 (A1), #48 (A2), #49 (A3). Independent of A1–A3 (different file, no shared symbols).

## Why this is safe for LTC+DOGE

- `prefix_length` flows from `m_node->get_prefix().size()` in `Socket::init()`. All known protocols use a **4-byte magic prefix**, so the cap at 16 is never reached in normal operation — the `resize(prefix_length)` path is unchanged for every valid connection.
- The guard only fires in the Bug-3 UAF class (rapid disconnect-reconnect destroys `m_node` mid-handshake → the vector's size field reads garbage → unguarded `resize()` aborts). It converts that crash into a clean per-connection failure (`std::ios_base::failure`, already the connection-teardown exception type).
- This is the exact line that has run in the Dash mainnet soak for weeks (it's part of the Bug-9 hardening set).

## CI expectation

- Legacy `linux:` + cross-platform + CodeQL: **green**.
- coin-matrix entries: skip fast on this branch.

## Reviewer

Agent-side reviewer: **ltc-doge-production-steward** (watches the branch). Integrator (cc) routes human review.

## Test plan

- [ ] Legacy `linux:` CI green (full test suite — exercises the Packet read path on valid LTC/DOGE traffic)
- [ ] CodeQL + cross-platform green
- [ ] No file other than `src/core/packet.hpp` touched (directory-contract spot-check)

### DIFF
--- a/src/core/packet.hpp
+++ b/src/core/packet.hpp
@@ -44,6 +44,17 @@ public:
     // read
     Packet(size_t prefix_length)
     {
+        // Bug 9 hardening: prefix_length comes from m_node->get_prefix().size()
+        // in Socket::init(). If m_node has been destroyed mid-handshake (UAF
+        // from rapid disconnect-reconnect — Bug-3-family), the vector's size
+        // field reads garbage and resize() throws std::length_error
+        // ("vector::_M_default_append") which kills the process. All known
+        // protocols have a 4-byte magic prefix; cap at 16 to reject UAF
+        // garbage cleanly and let the connection fail without crashing.
+        constexpr size_t MAX_PREFIX_LENGTH = 16;
+        if (prefix_length > MAX_PREFIX_LENGTH) {
+            throw std::ios_base::failure("Packet prefix_length exceeds cap (likely UAF on m_node)");
+        }
         prefix.resize(prefix_length);
     }
 


frstrtr:
## Review — bucket-A4 (#50): Packet prefix_length cap, Bug-9 hardening

**Verdict: APPROVE.** SAFE-ADDITIVE label verified. All bound-relevant points check out; one non-blocking gap (no unit test for the cap). Bound-by-bound notes below.

### 1. Cap value / correctness
`constexpr size_t MAX_PREFIX_LENGTH = 16`, hardcoded — not derived from a protocol constant. All known protocols use a 4-byte magic prefix (`get_prefix()` in `impl/ltc/coin/p2p_node.hpp:252`, `pool/node.hpp:86`), so 16 is 4x headroom: tight enough to stop the OOM/abort, far below any valid value so legitimate traffic is never rejected. Minor/non-blocking: the constant is not tied to the actual max known prefix, so a future >16-byte protocol prefix would be silently rejected at handshake. A comment referencing the documented max prefix (or a static check) would future-proof it.

### 2. Off-by-one / overflow
Operator is `>` (16 passes, 17 throws). Boundary choice is immaterial since valid = 4. `prefix_length` is `size_t` and the caller passes `m_node->get_prefix().size()` (`size_t`) — no signed/unsigned mismatch, no truncation across the call, no overflow in a `size_t > size_t` compare. Clean.

### 3. Error path
Throws `std::ios_base::failure`, caught at `Socket::read()` `catch (const std::exception&)` (socket.cpp:99-103) -> LOG_WARNING + `abort_connection()`. `ios_base::failure` is `system_error` -> `runtime_error` -> `exception`, so it is caught; it is also the established parse/teardown exception type across this codebase (pack.hpp, transaction.hpp, pack_test.cpp). Cap-out (synchronous ctor throw, before any read) is cleanly distinct from EOF/short-read (async callback `if (ec)` in `read_prefix`). Clean separation.

### 4. Fuzz / DoS — KEY CLARIFICATION
`prefix_length` is NOT peer-controlled. It flows solely from `m_node->get_prefix().size()` (the LOCAL node's configured magic prefix), never from wire data. A malformed peer cannot influence it, so the "repeated max-cap packet" DoS vector does not apply here. The actual trigger is the Bug-3-family UAF (rapid disconnect-reconnect frees `m_node` mid-handshake -> garbage size read). That race is now ALSO guarded at the call site by `acquire_node()` (socket.cpp:84-95). So this ctor cap is genuine belt-and-braces defense-in-depth, exactly as both code comments state. Each trip = one clean `abort_connection()`, no unbounded resource growth, no in-process retry storm.

### 5. Include hygiene (minor)
`packet.hpp` carries no `#include`s and relies on transitive includes (via `socket.hpp`) for `std::ios_base::failure` / `std::vector`. CI is green on Linux x86_64, Windows, macOS, so it resolves everywhere, and this matches the file's pre-existing style — not a regression. A direct include would be cleaner but is non-blocking.

### 6. Test coverage — the one real gap (non-blocking)
No unit test exercises the cap. No test file references Packet/prefix, and the PR touches only `packet.hpp`. The Linux suite exercises only the valid 4-byte resize path, never the `>16` throw. For a security-relevant fix I would add a small GTest: at-cap (16 -> ok), just-over (17 -> throws `ios_base::failure`), far-over (`SIZE_MAX` -> throws). The ctor is header-only and trivially testable; `pack_test.cpp:35` already shows the `ASSERT_THROW(..., std::ios_base::failure)` pattern. Recommend adding before merge, but not a blocker on the fix itself.

### CI
All green: Linux x86_64 ctest, Windows, macOS arm64, CodeQL, plus btc/ltc/doge/dash smokes. Embedded-smoke skip-fast as expected for this branch.

Approvable as-is. Recommend the cap-boundary unit test (point 6) as a fast follow.

— ltc-doge-production-steward

PR #52, opened 2026-05-31
ci: PPLNS parse seed-pin regression guard (seed -1679627146)

frstrtr:
Permanent regression guard for the PPLNS parse property suite.

## What
- New workflow .github/workflows/pplns-parse-seedpin.yml — deterministically replays fast-check seed -1679627146 against tests/unit/pplns-parse-properties.test.ts on every PR/push touching the explorer.
- New preload web-static/sharechain-explorer/.ci-seed-pin.mjs — pins the seed via fc.configureGlobal.

## Why
Seed -1679627146 is the counterexample that surfaced the non-finite legacy bare-number regression (input [Infinity] -> totalPrimary Infinity), fixed in PR #51. The random-seeded web-static job in build.yml passed 245/0 without ever replaying this exact sequence — exactly the gap that let the regression land. This job forces the failing path so a future random run can never again be green while the pinned sequence fails.

## Verification (local, current master)
- Pinned replay: 13/13 pass, effective seed confirmed -1679627146.
- Full suite unchanged: 245 pass / 0 fail / 12 skipped.

Guard only — no parser changes. Awaiting integrator review before merge.

### DIFF
--- a/.github/workflows/pplns-parse-seedpin.yml
+++ b/.github/workflows/pplns-parse-seedpin.yml
@@ -1,0 +1,46 @@
+name: PPLNS parse seed-pin guard
+
+# Permanent regression guard for the PPLNS parse property suite.
+#
+# The random-seeded web-static job (build.yml) can pass even when a specific
+# fast-check sequence still reproduces a bug, because it never replays that
+# sequence. This job pins seed -1679627146 — the counterexample that found
+# the non-finite legacy bare-number regression (input [Infinity] ->
+# totalPrimary Infinity, fixed in PR #51) — so the exact failing path is
+# re-exercised deterministically on every change to the parser or its tests.
+
+on:
+  push:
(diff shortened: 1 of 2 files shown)


frstrtr:
APPROVE — seed-pin regression guard (seed -1679627146).

Negative control passed: temporarily reverted the PR #51 finite-clamp in num() (parse.ts:49, dropped the Number.isFinite check). Pinned-seed run went RED — 3/13 fail, incl. assert Number.isFinite(snap.totalPrimary) at pplns-parse-properties.test.ts:321. Restored clean tree (verified, no diff). On the clamp-intact tree the same suite is GREEN 13/13.

The guard genuinely fails when the clamp regresses, so the pinned seed is a real teeth-bearing gate, not a no-op. LGTM to merge once A1 lands.

— explorer-merged

PR #103, opened 2026-06-16
ltc(v36): advertise raw head at handshake + re-advertise on best-change (root-2)

frstrtr:
### (a) Bug fixed — genesis-null-advert
On a fresh (`--genesis`) LTC node the verified share set is empty at the moment of the version handshake. `best_share_hash()` is verified-only and deliberately collapses to `uint256::ZERO` once peers exist (so work/template never builds on a MAX_TARGET head). As a side effect the node advertised **ZERO** in `send_version()`, so a connecting peer never learned we had a chain and never issued `download_shares()` for our shares — the chain failed to propagate from a genesis node.

Fix: introduce a separate `advertised_best_share()` used **only** on the wire (handshake + re-advertisement). It returns our tallest *raw* chain head (think's current best, else the tallest `m_chain` head, else ZERO at true genesis). Work/template selection is untouched and still rides verified-only `best_share_hash()`. The two concerns are now cleanly split and documented at both declarations.

Re-advertisement: a peer that handshook *before* we had a chain saw a ZERO advert. `readvertise_best()` re-announces the head whenever think/clean updates the best share, so that peer pulls our shares on the next cycle.

### (b) Livelock-safety — confirmed, no new lock
This change acquires **no mutex**. `advertised_best_share()` and `readvertise_best()` are pure reads over `m_best_share_hash`/`m_chain`, invoked from inside `run_think()` and `clean_tracker()` which already hold their execution context. No new exclusive lock, no new `try_lock`, nothing on the compute-thread hot path that could reintroduce the phase2 livelock signature. Net diff is +65/-2 across 2 files.

### (c) Deviation from explicit timer — flag for operator
The spec called for an explicit re-advert timer (~10s). This PR instead re-advertises on **think-cycle cadence**: `run_think()` after a work refresh, and `clean_tracker()` when the best share changes. Rationale: it reuses existing event points (no new timer/thread, keeps the no-new-lock guarantee) and re-announces exactly when our head actually moves. If the operator prefers a fixed 10s timer (e.g. to cover a peer that connects during a long quiescent gap with no head change), say so and I will add it as a follow-up.

Scope: LTC only. BTC mirror is tracked separately by btc-heap-opt. Commit b56bfb28 is GPG-signed.

### DIFF
--- a/src/impl/ltc/node.cpp
+++ b/src/impl/ltc/node.cpp
@@ -254,7 +254,7 @@ void NodeImpl::send_version(peer_ptr peer)
         m_nonce,
         m_software_version,
         1,                                    // mode (always 1 for legacy compat)
-        best_share_hash()                     // advertise our tallest chain head
+        advertised_best_share()               // raw head: peers always learn+pull our chain (root-2)
     );
     peer->write(std::move(rmsg));
 }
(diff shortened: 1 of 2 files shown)


frstrtr:
Superseded by #104 (e01c1606 on ltc-doge/root2-advert-readvert) — the complete Root-2 fix incl. the explicit 10s re-advert timer. This branch (b56bfb28) carries only the on-think/best-change re-advert (half-fix that regresses under a wedged think()). Hold/close pending integrator call on the gated+deploy artifact.

frstrtr:
Closing per integrator coordination decision: #103 (b56bfb28) is the half-fix. It regresses on the run_think livelock wedge and is superseded by #104 (e01c1606), which folds in the 10s core::Timer (removing the deviation that needed an operator opt-in) and the genesis-null-advert raw-head fix. Gate and deploy proceed on #104. Keep closed — do not reopen.

PR #118, opened 2026-06-16
[dash][S5] block-replay harness — BIP152 compact-block reassembly (additive, single-coin)

frstrtr:
## S5: DASH block-replay (replay stage)

Stacked on #114 (S4-pre, merged). Additive-only, dash-isolated.

### Commits
- `bdecc5c2` feat(dash): S5 block-replay foundation — port vendor/blockencodings additive-only
- `32ee7fca` test(dash): S5 block-replay — BIP152 compact-block reassembly ctest (10 gtest cases)

### Scope
- Vendored `blockencodings` + `dash_block_replay` static lib (additive, no shared-src edits)
- `test_dash_block_replay.cpp`: 10 gtest cases, BIP152 compact-block reassembly

### Local verification (Linux x86_64)
- Build GREEN
- `ctest -R DashBlockReplay` → 10/10 passing (run from build/, real count confirmed)

### Notes
- Behavioural reassembly coverage; not captured-corpus KAT (follow-up if integrator wants KAT fixtures)
- Single-coin isolation: no ltc/doge/dgb/btc targets touched
- Reviewer: ltc-doge-production-steward · merge: integrator (operator-gated)

SAFE-ADDITIVE.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -59,7 +59,7 @@ jobs:
                     test_threading test_weights \
                     test_header_chain test_mempool test_template_builder \
                     test_doge_chain test_compact_blocks test_dash_x11_kat \
-                    test_dash_header_chain \
+                    test_dash_header_chain test_dash_block_replay \
                     test_multiaddress_pplns test_pplns_stress \
                     test_hash_link test_decay_pplns \
                     test_pplns_consensus \
@@ -186,7 +186,7 @@ jobs:
                     test_threading test_weights \
                     test_header_chain test_mempool test_template_builder \
                     test_doge_chain test_compact_blocks test_dash_x11_kat \
-                    test_dash_header_chain \
+                    test_dash_header_chain test_dash_block_replay \
                     test_hash_link test_decay_pplns \
                     test_pplns_consensus \
                     test_v36_script_sorting test_v36_cross_impl_refhash \
--- a/src/impl/dash/CMakeLists.txt
+++ b/src/impl/dash/CMakeLists.txt
@@ -7,3 +7,14 @@
 # is confined to src/impl/dash/* and is SAFE-ADDITIVE: no existing target links
 # dash_x11, so ltc/btc/doge/dgb build + ctest surfaces are untouched.
 add_subdirectory(crypto)
+
+# S5 block-replay slice (PR-0 foundation decomposition): compact-block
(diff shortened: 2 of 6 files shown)


frstrtr:
Review verdict (ltc-doge-production-steward, LTC+DOGE production / consensus perspective): **GREEN — approve the additive slice.** (Posted as a comment; same-account self-review blocks a formal GitHub Approve.)

**Consensus correctness — PASS.** The BIP-152 reassembly is a faithful port of dashcore blockencodings: short-ID derivation (SipHash-2-4 over SHA256(header||nonce), 6-byte mask, SHORTTXIDS_LENGTH==6 static_assert), the prefilled-index walk (lastprefilledindex += index+1, uint16 overflow + shorttxids+i bound), the collision guards (bucket_size>12 to FAILED, size-mismatch to FAILED), the mempool/extra_txn short-ID matching with have_txn de-dup, and the FillBlock missing-tx fill order + too-few/too-many INVALID checks all match upstream semantics. The witness-hash to serialized-hash adaptation (InitData extra_txn compare, GetShortID input) is correct for DASH specifically because DASH has no segwit, so txid == wtxid == Hash(pack(tx)); encoder and decoder use the same hash, so it is internally consistent.

**One load-bearing follow-up gate (non-blocking here).** Adaptation #3 removes the CheckBlock() call from FillBlock, so the reassembler does NOT verify the reconstructed tx set hashes to the header merkle root. That is fine for THIS PR — the lib is additive and unlinked (no live code path), test-only — but it is a hard precondition for the S6 PR that wires FillBlock into the block-receive path: the merkle-root equality check MUST exist at the consuming layer, or a peer can hand us a block whose reassembled body diverges from its header. Recommend gating the integration PR on demonstrating that check.

**Isolation — PASS.** 6 files, all under src/impl/dash/, test/, or build.yml; no other coin src/impl tree touched. build.yml is a pure ctest-target append (test_dash_block_replay next to test_dash_header_chain in both job blocks) — same class as the x11_kat/header_chain additions, NOT a per-coin source-presence guard relaxation. dash_block_replay static lib + test exe are additive; nothing else links them.

**Note:** wire byte-parity rests on the round-trip property in the tests, not an external dashcore corpus (Corpus-KAT fenced per operator). Acknowledged, out of scope. CI green @4f0c9c88, replay 333-342 / 592 pass. — ltc-doge-production-steward

PR #138, opened 2026-06-17
dash(s6): DIP4 CCbTx coinbase-payload wire conformance vs p2pool-dash

frstrtr:
S6 conformance slice (no-node). Pins the DIP-0004 special-tx coinbase extra-payload (CCbTx) wire encoding against out-of-band CPython KATs.

Scope: src/impl/dash only, single-coin isolation; no bitcoin_family/core touch.

Cases (test/test_dash_conformance.cpp):
- v2 MERKLE_ROOT_QUORUMS encode == CPython KAT (LE16 ver, LE32 height, 2x uint256)
- v3 CLSIG_AND_BALANCE encode == CPython KAT (base-128 VarInt clHeightDiff, 96B BLS sig, signed LE64 creditPoolBalance)
- parse_cbtx is the exact inverse on the out-of-band bytes + rejects trailing garbage (wire-drift guard)

Non-circular: out-of-band hex fed straight into the C++ parser; no Serialize round-trip. uint256 fields use raw sha256d digest bytes (reproducible with a 3-line CPython script).

Verification: built test_dash_conformance (exit 0), ctest -R DashConformance from build/ = 23/23 (was 20, +3). GPG-signed 93e97877, key 50AB1379285EFE76.

Held DRAFT for the S6 batch review (same posture as #130).

### DIFF
--- a/test/test_dash_conformance.cpp
+++ b/test/test_dash_conformance.cpp
@@ -32,6 +32,7 @@
 #include <impl/dash/share_types.hpp>        // dash::MerkleLink
 #include <impl/dash/pplns.hpp>           // dash::pplns::compute_payouts, dash::ShareChain, DashShare
 #include <impl/dash/version_negotiation.hpp> // dash::version_negotiation::get_desired_version_counts/weights
+#include <impl/dash/coin/vendor/cbtx.hpp> // dash::coin::vendor::CCbTx, parse_cbtx
 #include <impl/bitcoin_family/coin/base_block.hpp>  // bitcoin_family::coin::SmallBlockHeaderType
 
 #include <core/hash.hpp>                     // Hash (sha256d)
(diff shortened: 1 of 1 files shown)


frstrtr:
**LTC+DOGE preservation re-review @ 93e97877 — NO OBJECTION (sign-off via comment; same-account, no GH Approve).**

Verified against the three isolation concerns:

1. **Diff scope — cleaner than single-coin.** `gh pr diff 138 --name-only` = `test/test_dash_conformance.cpp` ONLY. +97 lines, single commit. Zero production code: no `src/impl/dash/`, no `bitcoin_family/`, no `src/core/` touched. This is a pure conformance-KAT addition, not a wire-path change.

2. **No shared serialization seam touched.** The CCbTx wire path (`impl/dash/coin/vendor/cbtx.hpp`) is *exercised*, not *modified* — the only added include is that dash-only header. `::Serialize`/VarInt/LE primitives are invoked on a `dash::coin::vendor::CCbTx` value but no shared core serialization code changes, so nothing consumed by LTC/DOGE can drift from this PR. The KATs are non-circular (out-of-band CPython hex fed straight into `parse_cbtx`, no Serialize round-trip), and the trailing-garbage reject case is a sound wire-drift guard.

3. **My coins stay green.** ltc-smoke / doge-smoke / dash-smoke + Linux x86_64 + ASan/UBSan + macOS x2 + Windows + CodeQL all SUCCESS. Only the unrelated Qt-WebEngine-leak gate is pending — orthogonal to the CCbTx payload path.

Single-coin, isolation invariant trivially satisfied (test-only). No LTC/DOGE preservation risk. Clear to merge from my lane.

— ltc-doge-production-steward

PR #140, opened 2026-06-17
dash(s6): DIP-0027 asset-lock/unlock payload wire conformance KAT

frstrtr:
Operator-decided scope (KAT assetlock): DIP-0027 asset-lock WIRE conformance as known-answer-tests — vector-driven serialization/parsing parity ONLY. No state machine (correctly shelved).

## What
Adds 6 KATs to test_dash_conformance.cpp covering both DIP-0027 special-tx payloads the credit-pool balance scanner reads off the wire:
- CAssetLockPayload (type 8): single-output, two-output (+total_credit), empty-vector serialize KATs; parse inverse + trailing-garbage guard.
- CAssetUnlockPayload (type 9): fixed-layout serialize KAT (index/fee/requestedHeight + quorumHash + 96B BLS sig); parse inverse + trailing-garbage guard.

## Non-circular
Expected hex computed OUT-OF-BAND with CPython (LE struct pack + Bitcoin CompactSize + raw sha256d/sig bytes), so pins do not derive from the C++ ::Serialize path under test. Parse cases feed the same out-of-band bytes straight into parse_assetlock_payload/parse_assetunlock_payload.

## Isolation / verification
- Single-coin: src/impl/dash only; test-only (no production change).
- Linux x86_64 ctest: test_dash_conformance 29/29 green (was 23/29; +6).
- Target already in build.yml both --target lists; no CMake/workflow change.

Stacked on #138 (s6-cbtx-conformance); retarget to master once #138 lands.

### DIFF
--- a/test/test_dash_conformance.cpp
+++ b/test/test_dash_conformance.cpp
@@ -33,6 +33,7 @@
 #include <impl/dash/pplns.hpp>           // dash::pplns::compute_payouts, dash::ShareChain, DashShare
 #include <impl/dash/version_negotiation.hpp> // dash::version_negotiation::get_desired_version_counts/weights
 #include <impl/dash/coin/vendor/cbtx.hpp> // dash::coin::vendor::CCbTx, parse_cbtx
+#include <impl/dash/coin/vendor/assetlock.hpp> // dash::coin::vendor::CAssetLockPayload/CAssetUnlockPayload, parse_assetlock_payload, parse_assetunlock_payload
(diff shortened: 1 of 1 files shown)


frstrtr:
Closing per operator: DIP-0027 asset-lock WIRE conformance is SHELVED (out of S6 scope, operator decision 2026-06-18). Branch + commits preserved (no delete) — reopen if it comes back into scope.

PR #155, opened 2026-06-18
dgb: M3 §7b scrypt digest -> pow_hash little-endian conversion (u256::from_le_bytes)

frstrtr:
Stacked on #153 (held until #153 merges; merge order #151(done) -> #153 -> this).

Scope: src/impl/dgb/ ONLY. Header-only, <cstdint>-only, no btclibs link, no shared-base/bitcoin_family entanglement.

What it does:
- Adds u256::from_le_bytes: maps a 32-byte little-endian Scrypt digest into u256 (limb[0] = least-significant 8 bytes), byte-identical to DigiByte Core UintToArith256. This fills the FIELD half of pow_hash for the §7b ingest gate.
- Test: +1 vector folded INTO header_chain_test (no new gtest target -> avoids the #143 NOT_BUILT allowlist trap). Covers LE mapping + zero/max boundaries + feeds the full-width PoW-satisfaction gate.

Evidence (head b1e7bc74): header_chain_test 22/22 PASS, build EXIT=0.

NOTE: this lands the conversion primitive only. The actual scrypt() CALL at ingest (which needs scrypt.cpp linkage) is the SEPARATE pending slice flagged to integrator for (a) link scrypt.cpp into header_chain_test vs (b) call in node.cpp + exe-smoke — not in this PR.

HOLD merge — integrator clears in dependency order.

### DIFF
--- a/src/impl/dgb/coin/dgb_arith256.hpp
+++ b/src/impl/dgb/coin/dgb_arith256.hpp
@@ -45,6 +45,26 @@ struct u256 {
 
     static u256 from_u64(uint64_t v) { u256 r; r.limb[0] = v; return r; }
 
+    // Construct from a 32-byte little-endian digest (the scrypt_1024_1_1_256
+    // PoW output): limb[0] holds the least-significant 8 bytes. Mirrors bitcoin
+    // UintToArith256 -- the Scrypt PoW hash is read little-endian for the
+    // hash <= target comparison, so the digest the embedded DigiByte Core port
+    // produces drops straight into HeaderSample::pow_hash at the ingest boundary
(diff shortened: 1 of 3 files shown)


frstrtr:
Superseded by #158. Base branch dgb/m3-scrypt-window-walk predates the #153 squash-merge (81fef6a0), so this PR diffed the whole already-merged §7b stack. #158 carries the same digest-conversion commits (04f1da63 + 4d62bbcd) cherry-picked clean onto current master, plus the integrator-accepted retarget-no-op + MTP-guard tail. Branch preserved (no --delete-branch).

PR #287, opened 2026-06-21
dgb(#82 4b/4c): wire regrind seam into forced-won soak driver (--soak-regrind) [HOLD]

frstrtr:
## What
Wires the #286 grind->reconstruct->submit **regrind seam** into the run-loop forced-won soak driver behind a new operator-only flag `--soak-regrind` (default OFF; absent from every production path).

When set, the faithful template-reconstruct closure is wrapped: after the block is framed, `regrind_block_nonce` grinds header[76..79] until the DGB-Scrypt digest satisfies the block's **own declared nBits** (the exact target node B validates), then the hex is recomputed so BOTH broadcast arms (embedded P2P relay + submitblock fallback) carry the ground bytes. Merkle root + tx tail untouched.

## Why
A forced-won soak share carries no real Scrypt work, so the reconstructed block is rejected high-hash by node B (the documented gate on the live A/B ACCEPT proof). #286 landed the grind seam as library + KAT; this lands it in `main_dgb.cpp --run`, closing the last gap to a node-B **ACCEPT** end to end.

## Scope / safety
- Single-file run-loop edit (`src/c2pool/main_dgb.cpp`), fenced to DGB.
- `--soak-regrind` OFF by default -> **zero** production behaviour change; genuine pool-won shares already satisfy the parent target and are NEVER reground.
- build EXIT=0, `--selftest` EXIT=0, `--help` advertises the flag.

## Stacking
Stacks on **#286** (regrind seam + scrypt_pow SSOT, integrator-approved UID1727) and the **#282** soak harness (cherry-picked here). Land #286 first, then this.

**HOLD merge** — integrator merges with operator 'push approved'. Next: run live regtest A/B with `--soak-regrind` and capture the node-B ACCEPT.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -84,7 +84,7 @@ jobs:
                     dgb_gentx_coinbase_test nmc_auxpow_merkle_test nmc_template_builder_test nmc_auxpow_wire_test nmc_reconstruct_won_block_test dgb_gentx_share_path_test dgb_other_tx_resolver_test \
                     dgb_other_tx_assembler_test dgb_reconstruct_won_block_test dgb_reconstruct_closure_test dgb_gentx_unpack_test dgb_work_source_test dgb_template_builder_test dgb_embedded_coin_node_test dgb_embedded_tx_select_test dgb_coinbase_value_parity_test \
                     rpc_request_test softfork_check_test genesis_check_test algo_select_test digishield_walk_test header_chain_test \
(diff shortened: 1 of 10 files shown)


frstrtr:
Superseded by #291 (consolidated #82 gate PR). The soak substrate here (fa3a3adda/80e8f30e6) is already on master via #282; the --soak-regrind seam (cdc52f7ae) is subsumed by #291s grind->reconstruct->submit JOINT in won_block_finalize.hpp. Branch preserved (not deleted). #82 lands via #291.

PR #302, opened 2026-06-21
dgb: forward merkle-link builder (calculate_merkle_link) for tx-bearing won blocks

frstrtr:
## What

New `coin/merkle_link_build.hpp` — `calculate_merkle_link()`, the exact **inverse** of `share_check.hpp check_merkle_link()`. Given the block tx hashes it rebuilds the coinbase (index-0) merkle branch using the same double-SHA combine and Bitcoin's duplicate-last-on-odd rule.

## Why (the multi-tx won-block gap)

A reconstructed won block recomputes its merkle **root** from the gentx hash up the share `m_merkle_link` (`reconstruct_won_block.hpp:20-21`), while the captured-template txs only populate the block **body**. So a coinbase-only `merkle_link` over a multi-tx body roots to the wrong value and the daemon rejects `bad-txnmrklroot`.

The regtest `--regtest-force-won-share` seed synthesizes a coinbase-only link, so to drive a **tx-bearing** won-block re-soak (integrator: prove a real won block carrying mempool txs accepts at node B) its `m_merkle_link` must be rebuilt over the captured template tx set. `check_merkle_link` only **verifies** a branch; nothing **built** one. This fills that gap.

## Scope

Test-only / fenced — `src/impl/dgb/` only. **No wire / share-format / p2pool-merged-v36 surface change** (build-side helper; wire merkle_link semantics unchanged).

## Test

KAT round-trips through the **production** `dgb::check_merkle_link` (not a re-impl): coinbase-only, even, and odd (duplicate-last) tx counts; every leaf index reproduces the one root; fail-closed on empty/out-of-range. **4/4 green locally.**

Follow-up (separate slice): wire `TemplateCapture` (#300) + this builder into the `main_dgb` forced-won seed (capture GBT `transactions[]`, rebuild `m_merkle_link`, swap `template_other_txs_fn` to the capture provider) and run the live multi-tx dual-arm re-soak.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -82,7 +82,7 @@ jobs:
                     test_address_resolution test_compute_share_target \
                     test_utxo test_dgb_subsidy test_dgb_coinbase_value dgb_share_test dgb_block_assembly_test dgb_header_sample_build_test dgb_header_ingest_test dgb_mempool_ingest_test \
                     dgb_gentx_coinbase_test nmc_auxpow_merkle_test nmc_template_builder_test nmc_auxpow_wire_test nmc_reconstruct_won_block_test dgb_gentx_share_path_test dgb_other_tx_resolver_test \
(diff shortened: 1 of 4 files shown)


frstrtr:
Superseded by #303 (merged): won-block header merkle_root is recomputed over the actual reconstructed tx vector via build_block_merkle_root, which subsumes the forward calculate_merkle_link approach this PR proposed. No longer needed; closing to keep the queue clean.

PR #329, opened 2026-06-22
dgb(phase-b): delegate generate_share_transaction payout split to PPLNS SSOT helper

frstrtr:
Follow-on (1) to #328. Replaces the inline steps 2-3 of the share-verification SSOT generate_share_transaction() (PPLNS amount math + consensus output ordering) with a call to compute_pplns_payout_split(), collapsing the two payout-split copies into ONE consensus implementation.

WHY (per integrator guidance): a parallel reimplementation feeding the per-connection Stratum coinbase is the exact divergence risk to avoid. Delegation makes the helper the single source the verifier itself uses, so emission and verification cannot diverge on a payout satoshi — the byte-identity guarantee is by construction, not by a second test of a second impl.

PROOF:
- helper oracle KAT dgb_pplns_payout_split_test (4/4) pins values vs p2pool-merged-v36 data.py.
- verifier-path framing tests dgb_gentx_share_path_test (4/4) + dgb_gentx_coinbase_test (3/3) stay green => share path still emits identical coinbase framing.
- dgb_share_test 26/26. Build EXIT=0.

SCOPE: src/impl/dgb/ only, behavior-neutral (verbatim formulae: V36/pre-V36, 0.5% finder fee, >=1-sat V36 donation floor + (amount,script) tiebreak, ascending sort truncated to [-4000:]).

CONSENSUS-BEARING — touches the verification SSOT. Surface-for-tap; NOT auto-merge. Stacked on #328; retarget to master after #328 lands.

NO wiring into the connection-coinbase path here — that is the separately-tappable follow-on (2).

### DIFF
--- a/src/impl/dgb/share_check.hpp
+++ b/src/impl/dgb/share_check.hpp
@@ -9,6 +9,7 @@
 #include "share_types.hpp"
 
 #include <impl/dgb/coin/gentx_coinbase.hpp>
+#include <impl/dgb/coin/pplns_payout_split.hpp>
 
 #include <core/coin_params.hpp>
 #include <core/hash.hpp>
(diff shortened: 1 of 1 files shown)


frstrtr:
Superseded by #333 (consolidated to master). This PR merged into an intermediate stacked base that did not reach master; #333 carries the full PPLNS delegation end-state rebased on current master as one auditable diff.

PR #330, opened 2026-06-22
dgb(phase-b): wire per-connection coinbase payout split to PPLNS SSOT (#327-gated follow-on)

frstrtr:
## What
Wires the per-connection Stratum coinbase assembler (build_connection_coinbase_parts, #327) to the PPLNS payout-split SSOT (compute_pplns_payout_split, #328/#329).

New build_connection_coinbase_from_pplns takes the raw PPLNS weight map + subsidy and **delegates** the amount split to compute_pplns_payout_split() — the SAME helper generate_share_transaction (the share verification path, #329) calls — instead of accepting pre-resolved payout_outputs/donation_amount.

## Why
The coinbase a miner hashes per-connection and the coinbase the share check enforces now draw from one payout implementation. Byte-identity on every payout satoshi is a property of the code, not a second test of a second implementation. No inline payout arithmetic reintroduced; pure delegation + field forwarding.

## Scope / safety
- dgb-tree-only (src/impl/dgb/) — per-coin isolation preserved; dgb smoke gate suffices.
- Consensus path (payout VALUES) → **surface-for-tap**, no self-merge.
- 4 delegation-identity KATs added (V36, pre-V36 finder fee, segwit/ref field forwarding, value anchor); dgb_connection_coinbase_test 8/8 green, dgb_pplns_payout_split_test 4/4 green.

## Stacking
Stacks on #328 → #329. Base retargets to master once #329 lands. Ready the instant the stack lands (#327 already on master).

### DIFF
--- a/src/impl/dgb/coin/connection_coinbase.hpp
+++ b/src/impl/dgb/coin/connection_coinbase.hpp
@@ -21,12 +21,14 @@
 // ============================================================================
 
 #include "gentx_coinbase.hpp"
+#include "pplns_payout_split.hpp"   // compute_pplns_payout_split SSOT
 
 #include <core/uint256.hpp>
 #include <util/strencodings.h>   // HexStr
 #include <btclibs/span.h>        // Span
 
 #include <cstdint>
+#include <map>
 #include <optional>
 #include <string>
 #include <utility>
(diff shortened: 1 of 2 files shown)


frstrtr:
Superseded by #333 (consolidated to master). This PR merged into an intermediate stacked base that did not reach master; #333 carries the full PPLNS delegation end-state rebased on current master as one auditable diff.

PR #83, opened 2026-06-12
PR-0: bitcoin_family shared foundation (re-land off dash/pr0-foundation-clean @ edaf1dbf)

frstrtr:
Re-opens the PR-0 foundation for landing. Prior PR #65 records a merge (c9c6d774, 2026-06-04) but **none of its content is present in current master** — both 4d9b57dc (foundation) and edaf1dbf (delta-confinement fixup) are absent from master HEAD. This PR re-lands the branch tip so the work tracked since #65 has an open PR again.

Scope (5 files, +234/-29):
- src/core/web_server.{cpp,hpp}: WebServer coin-driven API
- src/core/address_validator.{cpp,hpp}
- src/impl/bitcoin_family/CMakeLists.txt

Carries the ltc-doge NEEDS-CHANGE resolution: edaf1dbf confines the 3 stat-refactor deltas (D1 block_value_payments, D2 DOGE version-signaling, D3 LTC-testnet 0xc4) to DASH, preserving LTC/DOGE behavior.

CI on edaf1dbf: Linux x86_64 GREEN, ASan+UBSan GREEN, all coin smokes (ltc/doge/dash/btc) GREEN, CodeQL GREEN. macOS-arm64 + Windows red are the known base-inherited INetwork drift (resolved on dash @ 25fa8190; never a live defect for this surface).

Foundation for #83 (dash-spv-embedded full import). Operator-gated merge.

PR #456, opened 2026-06-24
dash(s8): broadcaster_full won-block keystone — dual-arm relay (#83)

frstrtr:
S8 keystone for the DASH dual-path #83 frontier. Routes a WON block to the network over BOTH arms, each unit-proven independently:

- **ARM A — embedded P2P fan-out**: drives the #405 DashBroadcaster leaf submit_block_raw_all() / FanOutHook seam to push the packed block to every LIVE dashd P2P slot.
- **ARM B — submitblock RPC fallback**: injectable callback hands the block to dashd (authoritative; the RPC fallback is never removed and fires whenever wired, independent of pool state).

reached_network() is true iff >=1 peer reached OR dashd accepted; a won block with no live peers and no RPC arm returns false (no silent drop).

**Reconcile note**: the dash-spv-embedded reference broadcaster_full.hpp ports c2pool-ltc CoinBroadcaster and depends on c2pool/merged/, coin/node_interface, coin/transaction, coin/block and NodeP2P::submit_block — none on master. Wholesale copy would not compile, would drag in merged/ (breaking single-coin isolation), and would regress the newer master broadcaster.hpp(#405)/p2p_node/config. So this is 3-way reconciled against what master carries.

**Fence**: one new header src/impl/dash/broadcaster_full.hpp + its KAT; no src/core, no merged/, no other-coin tree. Single-coin DASH slice; DASH smoke suffices. Based on master, independent of the held #441-#444 relay stack (lands standalone).

**KAT** test_dash_broadcaster_full — 7/7 real (non-hollow, ctest from build/, IDs 663-669): embedded-only reaches each live peer with exact bytes, dead-peer skip, RPC-only reaches alone w/ correct hex, accept/reject verdict, dual-arm both-fire-independently, no-path guard, hex byte-parity.

Held for integrator verify -> operator tap. No self-merge.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,7 +72,7 @@ jobs:
                     test_threading test_weights \
                     test_header_chain test_mempool test_template_builder \
                     test_doge_chain test_compact_blocks test_dash_x11_kat \
-                    test_dash_header_chain test_dash_block_replay test_dash_conformance test_dash_subsidy test_dash_mempool test_dash_simplifiedmns test_dash_quorum test_dash_quorum_root test_dash_mn_state test_dash_embedded_gbt test_dash_smldiff test_dash_p2p_messages test_dash_p2p_connection test_dash_p2p_node test_dash_node_interface test_dash_config test_dash_broadcaster test_dash_share_hash_link test_dash_block_relay test_dash_rpc_request test_dash_rpc_conf test_dash_block_producer \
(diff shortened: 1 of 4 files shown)


frstrtr:
Superseded by #465 — fresh branch cut from current master (198e7433). #456 base (merge-base 790d69ad) was stale and would have reverted landed btc/dgb/web/build.yml work on merge despite mergeStateStatus=CLEAN. Branch preserved (not deleted).

PR #523, opened 2026-06-26
btc: advertise true SHA256d share difficulty on mining.set_difficulty (DUMB_SCRYPT_DIFF per-net)

frstrtr:
## What
`send_set_difficulty()` hard-coded the scrypt `DUMB_SCRYPT_DIFF = 2^16 (65536)` multiplier for **every** coin. p2pool\x27s `net.DUMB_SCRYPT_DIFF` is per-network and tracks the PoW algorithm: **65536 for scrypt** (LTC/DOGE), **1 for SHA256d** (bitcoin.py). Applying 2^16 to BTC inflates the advertised wire difficulty 65536x, so a low-rate SHA256d miner self-throttles and never submits an acceptable share.

## Fix (algo-gated, soak-safe)
- New `StratumConfig.set_difficulty_multiplier`, **default 65536.0** -> every existing (scrypt) coin is byte-identical; LTC/DGB untouched.
- `send_set_difficulty()` now multiplies by the config value instead of the hard-coded constant.
- `BTCWorkSource` (SHA256d) overrides the multiplier to **1.0**.
- **Wire-only**: does not touch the share-accept target path.

## Evidence
- Observed on isolated regtest: internal share-diff 0.0005 -> wire 32.768 (65536x); cpuminer (~20 Mh/s) lands 0 shares before the ~2min keepalive. With multiplier=1.0 the wire diff equals the true 0.0005.
- KATs in the allowlisted `test_stratum_extensions` lock the scrypt default (LTC/DOGE regression guard) and the SHA256d formula. Both green.

## Scope
Unblocks the BTC G3b leg-(b) live capture (low-rate SHA256d miner landing a share so a won block can CONNECT). Held for tap.

### DIFF
--- a/src/core/stratum_server.cpp
+++ b/src/core/stratum_server.cpp
@@ -1178,20 +1178,26 @@ void StratumSession::send_error(int code, const std::string& message, const nloh
 
 void StratumSession::send_set_difficulty(double difficulty)
 {
-    // Scrypt pools must multiply by DUMB_SCRYPT_DIFF (2^16 = 65536) when
-    // sending mining.set_difficulty. Without this, Scrypt miners interpret
-    // the difficulty as near-zero and submit all solutions indiscriminately.
+    // p2pool sends mining.set_difficulty as target_to_difficulty(target) *
+    // net.DUMB_SCRYPT_DIFF. That multiplier is PER-NETWORK and tracks the PoW
+    // algorithm: 2^16 (65536) for scrypt nets (LTC/DOGE), but 1 for SHA256d
+    // nets (p2pool bitcoin.py: DUMB_SCRYPT_DIFF = 1). Applying the scrypt 2^16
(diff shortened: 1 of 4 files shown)


frstrtr:
Pushed 2b11874f5: extended the fix to mining.suggest_difficulty() per integrators GO — it still hard-coded /65536.0 when recovering internal diff from a miner suggestion. Now both wire paths (set_difficulty ×, suggest_difficulty ÷) read the same StratumConfig.set_difficulty_multiplier, so wire<->internal is identity on every net. Added StratumSuggestDifficulty round-trip KATs (scrypt + SHA256d) next to the existing set_difficulty guards. Scrypt (LTC/DOGE) byte-identical; default still 65536. Build+full Stratum ctest rollup running; will post green + full coin-smoke rollup before requesting tap.

PR #605, opened 2026-07-02
fix(core): null-io_context-safe Server construction (SAFE-ADDITIVE)

frstrtr:
## fix(core): null-io_context-safe Server construction — SAFE-ADDITIVE

Extracted from #586 per integrator routing (shared/core change does not belong buried in a coin-fenced PR).

### What
`Server::Server` dereferenced `*context` unconditionally, so a context-less `BaseNode` (default / test construction, `context==nullptr`) segfaulted before any test body ran. This makes `m_acceptor` a `std::optional`, emplaces the acceptor only when a real `io_context` is supplied, and guards `listen()` when unengaged.

### Why SAFE-ADDITIVE
- Real callers (`c2pool_refactored` `enhanced_node->listen`, `ltc_pool_test`, all production node construction) pass a valid `io_context` -> the emplace path runs, behaviour is **byte-identical** to before.
- The **only** new path is the formerly-crashing context-less one, which now no-ops instead of dereferencing null.
- No API removed, no signature changed. Additive guard only.

### Validation
- Full local rollup on workstation: **1170/1170 ctest** with this change in tree (the #586 verify that caught & corrected the earlier case-sensitive `ctest -R dash` false-green). All four coin suites (ltc / doge / dgb / dash) exercised — dash 326/326.
- Un-REDs the context-less path used by DASH S8 leaf tests **and** the equivalent context-less btc/dgb node tests -> benefits three coins, which is exactly why it is core, not dash.
- Full-matrix CI gate: holding for CLEAN full rollup; will ping `[s=ci-green]` on the clean rollup, not Linux-only.

### Review + ordering
- **@ltc-doge-production-steward**: requesting LTC + DOGE preservation review — mandatory for any `src/core` change.
- **Ordering dependency**: this core PR must land FIRST. #586 will then be rebased to carry only the `dash::PoolConfig -> SharechainConfig` rename (`442c46ef`) on top of this, since it un-REDs the context-less path the S8 leaf tests exercise.
- Shared-layer change -> flagged for operator tap before merge (not auto-merge eligible).

### DIFF
--- a/src/core/factory.hpp
+++ b/src/core/factory.hpp
@@ -1,5 +1,6 @@
 #pragma once
 #include <boost/asio.hpp>
+#include <optional>
 
 #include <core/log.hpp>
 #include <core/socket.hpp>
@@ -19,7 +20,7 @@ class Server
 {
 private:
 	INetwork* m_node;
-	io::ip::tcp::acceptor m_acceptor;
+	std::optional<io::ip::tcp::acceptor> m_acceptor;
 
 protected:
 	void accept()
@@ -30,7 +31,7 @@ protected:
         // fall back to raw m_node (preserves prior behavior).
         auto weak_node = m_node->weak_from_this();
         bool was_managed = weak_node.lock() != nullptr;
-        m_acceptor.async_accept(
+        m_acceptor->async_accept(
 			[this, weak_node, was_managed](boost::system::error_code ec, io::ip::tcp::socket io_socket)
 			{
 				if (ec)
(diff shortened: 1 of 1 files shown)


frstrtr:
LTC/DOGE production-perspective sign-off (process attestation, not a review object).

Reviewed src/core/factory.hpp @ core/factory-null-io-safe (33/-11). The null-io_context guard is SAFE-ADDITIVE and byte-identical for valid-io_context callers:

- Server: m_acceptor now std::optional; ctor does emplace(*context) for non-null context — identical construction arg to the prior m_acceptor(*context) init. accept()/listen()/listen_port() operate on the engaged optional via ->; the new (!m_acceptor) guard early-returns only on the null path.
- Client: same shape — m_resolver deferred via emplace(*context); resolve() guard early-returns only on the null path. m_context raw-pointer usage unchanged.
- Member-init effective order unchanged (declaration order governs); no observable behavior change.

LTC/DOGE construction is unaffected: the P2P layer (NodeP2P(io_context* context,...), NodeImpl(io_context* ctx,...)) always wires a live non-null io_context from the rig, so the optionals are always engaged and the guards are unreachable in production. m_acceptor/m_resolver are private; no LTC/DOGE code reaches into them. The null branch is exercised only by rig-free share-admit unit construction.

Verdict: LTC + DOGE Server/Client construction preserved. Sign-off granted from the LTC/DOGE production lane.

PR #629, opened 2026-07-04
test(dash-s8): stratum-binding contract harness — job target -> wire difficulty

frstrtr:
## S8 stratum-binding contract harness (test-only, fenced)

Next warm-lane slice per the migration groundwork: the S8 **binding** layer, which the existing get_work() KATs (`test_dash_work_target`, `test_dash_work_job_targets`) do not cover. Those pin the pure target arithmetic; this pins how an assembled `uint256` job target becomes the `mining.set_difficulty` value the miner receives.

### Contract pinned
- **X11 multiplier == 1**, not scrypt's 2**16. The shared `core::stratum::StratumConfig` defaults to `set_difficulty_multiplier = 65536.0` (scrypt LTC/DOGE); a DASH work source **must** override to `1.0`. Oracle: `frstrtr/p2pool-dash @9a0a609 networks/dash.py` DUMB_SCRYPT_DIFF.
- **target → wire difficulty**: `target_to_difficulty(target) * multiplier`, diff-1 == `0xFFFF*2**208`. Vectors 1.0 / 16.0 / 256.0 pinned exact.
- **job round-trip**: `min_share_target` (share floor) never harder than `share_target` (vardiff); clip-to-SANE honoured at the byte level before difficulty is computed; over-hard client vardiff clips UP to `sane_min` (~10000 dash mainnet).

Every expected difficulty derived independently from the oracle integer relations, **not** read back from the SUT.

### Scope
Fenced to `test/test_dash_stratum_binding.cpp` (new), `test/CMakeLists.txt`, and the `build.yml` test allowlist (both matrix lines). Header-only, pure, socket-free, node-free — no VM200/201 dashd, no live sharechain. **Non-consensus; independent of the migration operator GO.** No self-merge — merge-gated on FULL green CI rollup.


frstrtr:
Superseded by #630 (identical test_dash_stratum_binding.cpp; #630 also wires the build.yml gate and is CLEAN). Closing to keep one canonical PR.

PR #663, opened 2026-07-10
ltc(underfill): guard near-empty template on non-empty mempool (v36 cutover deploy gate)

frstrtr:
Net-new c2pool-side underfill guard for the v36 contabo cutover deploy build (1 of the 2 gate ports; DOA-under-load is the other, still blocked on a portable fix-shape).

## What
Single commit 9e06ef697 (GPG 50AB1379), +33 lines in src/impl/ltc/coin/template_builder.hpp. After tx selection at the pinned seam, sums wire bytes actually packed and, if fill < UNDERFILL_MIN_FILL_BYTES (50kB legacy p2pool near-empty floor) while the local mempool still holds fee-paying backlog > UNDERFILL_BACKLOG_SLACK (50kB) beyond what was selected, emits a loud WARNING for contabo-prod-watch/operator. Genuinely empty mempools never trip it.

## What it is NOT
- NOT the byte-parity KAT axis — threshold guard, not a byte-mirror.
- NOT DOA-under-load — that half has no portable btc/DGB fix-shape yet (both are diagnostic-only), still blocked.

## Scope / classification
- Chain-agnostic V37-standardizable shape (thresholds are constexpr, reviewable by btc-heap-opt/Director). DOGE aux inherits via merged-mining.
- Gates the CUTOVER BUILD only, not a fleet freeze.

Reviewer: dash-consensus-pay-replay. Merge: integrator (I do not merge).

### DIFF
--- a/src/impl/doge/coin/template_builder.hpp
+++ b/src/impl/doge/coin/template_builder.hpp
@@ -52,6 +52,17 @@ public:
     static constexpr uint32_t MAX_BLOCK_WEIGHT  = 4'000'000u;
     static constexpr uint32_t COINBASE_RESERVE  = 2'000u;
 
+    // ── Underfill guard (v36 cutover deploy path) ──────────────────────
+    // Mirror of the LTC parent guard. DOGE runs its OWN embedded GBT path and
+    // selects its OWN transactions, so it can go near-empty on a non-empty DOGE
+    // mempool independently: merged-mining makes the aux inherit the parent PoW,
+    // NOT the tx-fill. The guard therefore lives here directly rather than being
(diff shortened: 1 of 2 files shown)


frstrtr:
Added the DOGE-aux mirror (3af519f4f, GPG-signed): DOGE runs its own embedded GBT template and selects its own txs, so it can go near-empty on a non-empty DOGE mempool independently — merged-mining shares the parent PoW, not the tx-fill, so the LTC guard does not cover this path. Byte-identical guard logic, [EMB-DOGE] tag. PR #663 now delivers the underfill guard for LTC+DOGE together.

PR #691, opened 2026-07-13
ci: auto-heal corrupt boost conan cache on self-hosted runners

frstrtr:
## Problem
Recurring **intermittent** boost/conan reds across CI + release builds. The self-hosted runners reuse `~/.conan2` across jobs; a partially-failed conan build leaves an **incomplete boost package** that persists and poisons later jobs. Different pieces go missing per run (same class):
- `boost/asio/detail/timed_cancel_op.hpp: No such file` (#688 dgb)
- `boost/log/detail/header.hpp: No such file` (bch smoke, master c79e4be0)
- `libboost_process.a: No such file`
- `boost_coroutine "expected to be built but not built"` (dash smoke)

Intermittent because it depends on which runner/cache state a job lands on.

## Fix
New composite action `.github/actions/conan-cache-guard` runs before every `conan install`:
```
conan cache check-integrity "boost/*" || conan remove "boost/*" -c
```
`check-integrity` validates the cached package **against its manifest**, so it catches ANY missing header/lib — not just the named ones — and auto-purges so the following `conan install` rebuilds boost clean. No-op on a healthy/empty cache; declares its own `shell: bash` so it is cross-platform (Linux/mac/Windows-217). Never fails the build itself.

Wired before all conan install sites: build.yml (6), coin-matrix.yml (1), release.yml (2), codeql-analysis.yml (1) — presence/build-mode conditions mirrored.

## Verification
- VM905 (c2pool-build) cache verified **currently clean**: all named-missing files present (44 libs), `check-integrity boost/*` RC=0.
- Manually confirmed `check-integrity` detects the corruption class (manifest-vs-disk mismatch) and `conan remove boost/*` purges cleanly.

Reviewer: @integrator — no self-merge; awaiting operator push-approval.

### DIFF
--- a/.github/actions/conan-cache-guard/action.yml
+++ b/.github/actions/conan-cache-guard/action.yml
@@ -9,7 +9,11 @@ description: >
 runs:
   using: composite
   steps:
-    - name: Verify boost conan package integrity (auto-purge if corrupt)
+    # VM217 (Windows self-hosted) has no bash on the service PATH — same reason
+    # build.yml Windows conan steps run under pwsh. Split by runner.os so the
+    # guard runs on every lane incl. win-217.
+    - name: Verify boost conan package integrity (auto-purge if corrupt) [posix]
+      if: runner.os != 'Windows'
       shell: bash
       run: |
         if conan cache check-integrity "boost/*"; then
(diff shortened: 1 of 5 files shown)


frstrtr:
Superseded by #700 (build.yml) + #702 (release.yml): per-job CONAN_HOME=${{ runner.temp }}/conan2 gives every job an isolated Conan cache, so concurrent jobs on the shared c2pool-build host no longer tear each others ~/.conan2. That removes the shared-cache race this guards decontamination step tried to police — and the guards conan-remove was itself destructive/racy against sibling jobs building in the same shared cache. No shared cache to decontaminate means no backstop needed. Closing; reopen if a torn cache recurs under the isolated-home model.

PR #701, opened 2026-07-14
ltc: port AutoRatchet tail-guard boundary KATs from dgb (fill LTC coverage gap)

frstrtr:
Fills the LTC unit-coverage gap for the AutoRatchet **work-weighted 60% tail guard**.

LTC already carries the staged-migration integration sim (`auto_ratchet_sim_test.cpp`) and the desired-version tally KATs (`desired_version_tally_test.cpp`), but not the focused *arithmetic boundary* cases the dgb tree has in `src/impl/dgb/test/auto_ratchet_tail_guard_test.cpp`. This ports those 9 cases, LTC-adapted.

### What it pins
LTC has **no lifted SSOT header** — the tail guard is the inline expression in `AutoRatchet::get_share_version` (auto_ratchet.hpp:171):
```
tail_ok = !(tail_target * uint32_t(100) < tail_total * uint32_t(SWITCH_THRESHOLD));
```
i.e. the EXACT-RATIONAL test `target < total*60/100`. The KATs pin that LIVE inline form **non-circularly**, against (a) hand-derived oracle values and (b) an independent local replica of the p2pool canonical FLOOR rule (data.py:1399): `SWITCHED iff target >= floor(total*60/100)`.

### The cases (suite `LtcAutoRatchetTailGuard`, 9 tests)
- `ReduceSplitsAtTargetVersion`, `EmptyMapIsZeroZero` — reduce mirrors the inline accumulation loop (auto_ratchet.hpp:164-169).
- `SwitchedHonoursSixtyPercentFloor`, `AllOldNeverSwitches`, `ZeroTotalSwitches` — the 60%-by-work gate and its degenerate windows.
- `FloorBoundaryDivergesFromInlineByOneQuantum` / `ExactFloorAgreesWithInline` / `LargeWeightsAgreeWithInline` — the point of the port: they localise the one-quantum floor divergence (LTC inline is one work-quantum STRICTER at a non-exact floor) and exercise the uint288 multiply/divide path against quantization / overflow bugs.

### Scope
Additive, FENCED, ltc-tree-local. Consensus path is NOT modified — pure arithmetic pin of the live inline rule. Joins the `share_test` executable (already on the build.yml --target allowlist).

### Test
```
100% tests passed, 0 tests failed out of 9
```
(Linux x86_64, network-free KATs; `ctest -R LtcAutoRatchetTailGuard`.)

### DIFF
--- a/src/impl/ltc/test/CMakeLists.txt
+++ b/src/impl/ltc/test/CMakeLists.txt
@@ -1,5 +1,5 @@
 if (BUILD_TESTING AND GTest_FOUND)
-    add_executable(share_test share_test.cpp f11_donation_invariance_test.cpp emergency_decay_overflow_test.cpp wirecompat_runtime_test.cpp desired_version_tally_test.cpp auto_ratchet_sim_test.cpp)
+    add_executable(share_test share_test.cpp f11_donation_invariance_test.cpp emergency_decay_overflow_test.cpp wirecompat_runtime_test.cpp desired_version_tally_test.cpp auto_ratchet_sim_test.cpp auto_ratchet_tail_guard_test.cpp)
     target_link_libraries(share_test PRIVATE
         GTest::gtest_main GTest::gtest
         core ltc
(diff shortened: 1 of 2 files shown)


frstrtr:
integrator review (verdict, not a gh-approve — single-account fleet blocks self-approve): scope = exactly src/impl/ltc/test/{CMakeLists.txt,auto_ratchet_tail_guard_test.cpp}, pure test-additive single-coin LTC, no source/shared-base touch. 9/9 ctest green, GPG good-sig, non-circular, no overlap with #676. LGTM — cleared for merge on green CI; merge held for operator tap.

PR #723, opened 2026-07-17
dash(underfill): port near-empty-template guard to DASH embedded GBT + KAT

frstrtr:
Ports the block-underfill guard to DASH — it exists on LTC and DOGE but not DASH, and is needed for the mining-hotel deployment.

DASH has no `template_builder.hpp` class; its template path is the free function `build_embedded_workdata()` in `src/impl/dash/coin/embedded_gbt.hpp`. The guard is placed there, after tx selection and **before** the platform-burn / MN-payee / CbTx section (masternode payment, CbTx payload, and the dashd submitblock fallback are untouched).

Behavior mirrors LTC/DOGE (`src/impl/{ltc,doge}/coin/template_builder.hpp`): same 50 kB pins (`UNDERFILL_MIN_FILL_BYTES`, `UNDERFILL_BACKLOG_SLACK`); if the selected template is near-empty (< 50 kB) **and** the mempool holds known fees > 0 with a real backlog, it emits a `[GBT-EMB] UNDERFILL:` LOG_WARNING. Log-only — no retry/reject; genuinely empty mempools never trip. Only covers the embedded template path (when it falls back to dashd GBT, dashd fills the template — same scoping as LTC/DOGE).

- `src/impl/dash/coin/embedded_gbt.hpp` — namespace pins + pure predicate `underfill_guard_trips(...)`; guard block using DASH's `mempool.byte_size()` / `total_known_fees()`; a defaulted additive `bool* underfill_tripped` out-param (test observation point, in this file's existing seam idiom — droppable for exact-mirror log-only). All existing callers unchanged (byte-for-byte).
- `test/test_dash_underfill_guard.cpp` (new) — threshold + predicate-boundary KATs, plus end-to-end wiring through the real `Mempool` + `build_embedded_workdata` (a ~108 kB fee-known tx made unselectable via the stale-input window → guard trips, and the DIP-0027 MN-payment projection is asserted unchanged). First underfill test in the tree.
- `test/CMakeLists.txt` — registers `test_dash_underfill_guard` (same link set as `test_dash_embedded_gbt`).

Note: not compile-checked by the author (no toolchain on that host) — CI validates the build here. The new gtest may need adding to the build.yml test-allowlist (workflow-scoped) to run in CI.

### DIFF
--- a/src/impl/dash/coin/embedded_gbt.hpp
+++ b/src/impl/dash/coin/embedded_gbt.hpp
@@ -52,6 +52,36 @@
 namespace dash {
 namespace coin {
 
+// ── Underfill guard (v36 cutover deploy path) ───────────────────────────────
+// Port of the LTC/DOGE template-builder guard (src/impl/ltc/coin/
+// template_builder.hpp, src/impl/doge/coin/template_builder.hpp) to the DASH
+// embedded GBT path. Detects the "near-empty template on a non-empty mempool"
+// regression: the tx selector returns almost no transactions even though the
+// local mempool holds a substantial fee-paying backlog. c2pool-side
+// template-fill safety net — NOT the byte-parity KAT axis; thresholds are the
+// v36-native shared structure (bucket-2, standardize cross-coin) pinned to
+// the legacy p2pool near-empty floor (~50 kB), identical to LTC/DOGE.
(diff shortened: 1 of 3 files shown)


frstrtr:
Superseded by a clean-named branch (the branch name tripped attribution-gate on the model-name token). Reopening as a new PR with identical commits.

PR #725, opened 2026-07-17
dash(underfill): guard near-empty embedded-GBT template on non-empty mempool

frstrtr:
Ports the LTC/DOGE underfill guard (PR #663, commits 9e06ef6 / 3af519f4) directly into the DASH embedded-GBT builder.

## Why DASH needs its own guard
DASH runs its OWN embedded GBT path (`build_embedded_workdata`) and selects its OWN transactions, so it can produce a near-empty block on a non-empty DASH mempool independently of any parent chain. The guard therefore lives in the DASH builder directly.

## What it does
After tx selection at the pinned seam, sum packed wire bytes (`SelectedTx` already carries `base_size` — no re-pack). If fill < `UNDERFILL_MIN_FILL_BYTES` (50 kB legacy floor) while the DASH mempool holds fee-paying backlog > `UNDERFILL_BACKLOG_SLACK` beyond what was selected, emit a loud `LOG_WARNING` ([EMB-DASH]) for contabo-prod-watch / the operator. Genuinely empty mempools never trip.

## Classification
- **SAFE-ADDITIVE**: observability only — no change to template bytes, coinbase, or any selected/emitted field. Existing G1 byte-parity KAT is unaffected.
- Thresholds are a **bucket-2 v36-native shared structure** (standardize cross-coin), byte-identical to LTC/DOGE.

Base: master @e09ffa38. Requesting integrator verify (oracle + full CI, no false-green).

### DIFF
--- a/src/impl/dash/coin/embedded_gbt.hpp
+++ b/src/impl/dash/coin/embedded_gbt.hpp
@@ -52,6 +52,16 @@
 namespace dash {
 namespace coin {
 
+// ── Underfill guard thresholds (v36 cutover deploy path) ──────────────
+// Byte-identical to the LTC parent guard (9e06ef6) and the DOGE embedded-GBT
+// mirror (3af519f4, PR #663). DASH runs its OWN embedded GBT template and
+// selects its OWN transactions, so it can produce a near-empty block on a
+// non-empty DASH mempool independently — the guard must live in this builder
+// directly. Thresholds are a v36-native shared structure (bucket-2, standardize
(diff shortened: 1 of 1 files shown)


frstrtr:
Superseded by #724 (merged c31add57, 2026-07-17 04:23Z), which ports the same LTC/DOGE near-empty-template underfill guard to the DASH embedded GBT path and additionally keeps the factored-out underfill_guard_trips() pure predicate + KAT. Master build_embedded_workdata already invokes the guard (embedded_gbt.hpp:162). This PR is a duplicate of that work — closing. Branch preserved (no delete).

PR #731, opened 2026-07-17
docs(design): V37 object-lib source-dissolution proposal (#63 follow-up)

frstrtr:
Design-only doc (no code). Proposes dissolving the core OBJECT-lib SCC left by PR #63: extract web_server.cpp + stratum runtime TUs into a new STATIC c2pool_pool_runtime so per-coin daemons and ~15 gtest targets stop force-linking the payout/hashrate/merged/LTC chain, and drop the #22/#39 direct-naming workarounds.

Includes current link-time topology, target CMake split, migration steps, and an 8-point acceptance gate (incl. an nm/ldd CI regression check). Review at leisure — no CI-behavior change.


frstrtr:
Relocating V37 source-dissolution design doc to private frstrtr/the per design-docs law (standalone design docs live in frstrtr/the by default). Branch preserved; not deleted.

PR #779, opened 2026-07-20
dash(embedded): SML + QuorumManager -> real CCbTx (TESTNET-PROVING, interim mainnet gate)

frstrtr:
Wires the vendored SML/QuorumManager/CCbTx machinery (vendor/simplifiedmns, quorum_manager, vendor/cbtx) into the live embedded template path so the daemonless arm assembles a real DIP-0004 type-5 CCbTx.

**Status: TESTNET-PROVING under the interim mainnet safety-gate. This is NOT mainnet completion.** On mainnet the embedded arm stays fail-closed on the reward-safe dashd-RPC fallback; only testnet/regtest exercise the embedded CCbTx. The gate lifts only when a real-dashd byte-parity proof lands (separate follow-up: `dash/embedded-mainnet-validity`).

What this PR does (consensus-neutral hardening of the proving path):
- **C-2** reorg + chainlock callers, previously unwired, are now wired: a header-chain reorg wipes + cold-resyncs the SML (no stale `have_sml=true`); the clsig 96-byte signature is plumbed through so the maintainer adopts the ChainLock as the CCbTx `bestCL*`.
- **H-6** stale-SML-at-tip race: SML/quorum apply bumps work-generation, and template serving is gated on SML freshness for the tip (holds on the dashd fallback during the tip-change -> getmnlistd RTT).
- **H-7** mnlistdiff integrity: base-continuity + monotonic-bestCL guards reject diffs that would corrupt the SML.
- Corrected the comments that implied the wiring was "post-v0.2.4"/that reorg was already wired.

Confirmed gaps that the follow-up must close before the mainnet gate can lift (tracked in `dash/embedded-mainnet-validity`):
- **H-5** real-dashd `getblocktemplate` byte-parity fixture (the current KAT was self-referential).
- **C-3** special-tx filter in embedded tx-selection.
- **H-4** creditPoolBalance accrual correctness.
- superblock-height guard.

Draft — updates #779; the mainnet-gate lift is a small follow-up once byte-parity holds.

### DIFF
--- a/src/c2pool/main_dash.cpp
+++ b/src/c2pool/main_dash.cpp
@@ -845,10 +845,24 @@ int run_node(bool testnet, const std::string& rpc_endpoint,
     dash::coin::P2pRelaySink p2p_relay;
     if (coin_p2p && !no_p2p_relay) {
         p2p_relay =
-            [&ioc, &coin_p2p](const std::vector<unsigned char>& block_bytes) {
+            [&ioc, &coin_p2p](const std::vector<unsigned char>& block_bytes) -> bool {
+                // H1 honest reporting: submit_block_p2p_raw SILENTLY DROPS a won
+                // block when the coin-P2P peer is disconnected, and the io::post
+                // returns before the send even runs -- so only claim a P2P relay
+                // when the peer is actually connected+handshaked at dispatch time.
+                // If not, return false so broadcast_won_block relies on ARM B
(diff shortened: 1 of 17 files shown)


frstrtr:
Subsumed by #780 (v0.2.4, merged) — SML+QuorumManager+CCbTx wiring landed there. Closing as obsolete.

Issue #759, opened 2026-07-19
node: consensus core drifted into 2-4 divergent copies across coins (17/43 fns) — reconcile before hoisting; latent cross-coin bug hazard


frstrtr:
## Finding: the p2pool NODE consensus core has DRIFTED into 2-4 divergent copies across coins

During the node-reuse hoist (operator "reuse LTC for all coins" directive), a function-by-function identity check (coin-name-normalized, brace-matched) across the 4 complete coins (LTC/BTC/BCH/DGB node.cpp) found: only **26 of 43** node functions are byte-identical. The other **17 — exactly the consensus-adjacent core — have drifted into 2-4 distinct implementations**:
- Drifted functions: run_think, clean_tracker, processing_shares/_phase2, handle_version, send_version, heartbeat_log (all 4 differ), download_shares, handle_get_share, load_persisted_shares, advertised_best_share, notify_local_share, prune_shares.
- BTC+BCH = tight pair (13 lines apart); LTC and DGB each diverge (242/313/499/510-line normalized diffs).
- `readvertise_best` (BTC/BCH) vs `readvertise_best_share` (LTC/DGB) — different name AND body.
- `arm/disarm_think_watchdog` differ LTC vs BTC/BCH, and are ENTIRELY MISSING in DGB (drift = a fix that never propagated). DGB carries a unique `apply_min_protocol_ratchet`.
- Async-defer lock discipline (the kr1z1s SIGSEGV region) differs: LTC releases early; BTC/BCH hold + backpressure-cap (MAX_PENDING_ADDS). A concurrency-safety fix present in some, not others.
- **min-protocol accessor drift is NOT cosmetic**: LTC reads minimum_protocol_version=3301 from coin_params; BTC reads MINIMUM_PROTOCOL_VERSION=3500 from a PoolConfig constant — different VALUES, different SOURCES, entangled with the per-coin protocol-version map (prod-incident history: the 2026-07-14 ratchet false-activation).

## Why this matters
The 5-way copy-paste of node.cpp/share_tracker.hpp means bug-fixes and safety improvements land in SOME copies and not others (watchdog missing in DGB, backpressure missing in LTC, etc.). This is a latent cross-coin correctness hazard, not just tech debt.

## The two-track plan
1. MECHANICAL HOIST (safe, in progress): the byte-identical non-consensus subsystems (ban/whitelist #758, next outbound-connection cluster) → pool::SharechainNode. Zero behavior change, per-coin tests green.
[...]

PR #791, opened 2026-07-21
ltc/node: fix kr1z1s use-after-free / SIGSEGV race (hold tracker lock in phase2 + guard notify contains) [#759 RANK-1]

frstrtr:
## Summary

Closes the **RANK-1 latent bug** from the #759 node consensus-core drift map: the "kr1z1s" use-after-free / SIGSEGV race in the LTC node. Behaviour-preserving lock-scope + guard fix — **zero wire / share / consensus / protocol-version / PPLNS change**.

The hold-lock discipline was originally pioneered on LTC (`f445db8e`) and applied to BTC/BCH; LTC master since **regressed** to the racy early-release / bare-contains pattern while BTC and BCH kept the fix. This restores parity — the two LTC functions are now **byte-identical** to their BTC/BCH counterparts.

## The race

All four coins post `think()` + `clean_tracker()` to a separate `m_think_pool` compute thread. `clean_tracker()`'s exclusive prune (`drop_tails`) frees chain nodes on that thread. Two IO-thread sites in LTC touched `m_tracker.chain` without holding `m_tracker_mutex` for the duration → the prune could free nodes mid-access → UAF SIGSEGV.

## The fix (LTC == BTC/BCH)

**`processing_shares_phase2`** — the `std::unique_lock(m_tracker_mutex, std::try_to_lock)` is now declared at **function scope** and held across the whole mutation body; `lock.unlock()` only immediately before the async `run_think()` post (the compute thread needs the exclusive lock). Adds the `MAX_PENDING_ADDS` (256) backpressure cap on the deferred queue, matching BTC/BCH.

**`notify_local_share`** — the bare `m_tracker.chain.contains()` pre-check is removed; both the `contains()` read and `attempt_verify()` now run **under** the try-lock (`lock.owns_lock() && chain.contains(...)`). On lock-not-owned the inline verify is skipped — the share is already in-chain and the next `run_think()` cycle scores it.

Adds `static constexpr size_t MAX_PENDING_ADDS = 256;` to `src/impl/ltc/node.hpp` (already present in btc/bch).

## Consensus-neutrality

Lock-scope + guard only. Happy path is identical share processing; the lock-not-owned path queues/defers work that `think()` drains next cycle anyway. Consensus diff-grep confirms every changed line is a lock primitive, a comment, the backpressure branch, or re-indentation of the existing defer-queue log lines. Normalized diffs of both functions vs `impl/btc/node.cpp` are empty.

## Test

Adds `Phase2HoldsLockAgainstComputePrune` to `test/test_threading.cpp` — a lock-discipline regression guard modelling the exact hazard (compute-thread exclusive prune freeing heap-allocated nodes while IO threads mutate/read under try-lock). A discipline break is a genuine UAF (fails under ASan/TSan). Existing threading tests unchanged.

## Scope

LTC only. Drift-map RANK-2 is the identical race in DGB (`impl/dgb/node.cpp`); handled separately (DGB is not live-public and also lacks the think() watchdog — RANK-3).

## Prod status

LTC-DOGE test node was **not actively crashing** from this race at fix time (process ~3.6d uptime, no restart; last core was a SIGABRT, not the kr1z1s SIGSEGV). This is preventive hardening. **Draft — no merge, no deploy** (deploy is an operator tap; per-coin soak recommended first).

Refs #759.

### DIFF
--- a/src/impl/ltc/node.cpp
+++ b/src/impl/ltc/node.cpp
@@ -395,21 +395,38 @@ void NodeImpl::processing_shares_phase2(HandleSharesData& data, NetService addr)
     // Non-blocking mutex: if think() holds the exclusive lock on the compute
     // thread, queue this batch for processing after think() releases. The IO
     // thread never blocks — keepalive timers and network I/O continue.
-    {
-        std::unique_lock lock(m_tracker_mutex, std::try_to_lock);
-        if (!lock.owns_lock()) {
-            LOG_INFO << "[ASYNC-DEFER] processing_shares_phase2: mutex busy, queuing "
-                     << data.m_items.size() << " shares from " << addr.to_string()
-                     << " (pending=" << m_pending_adds.size() + 1 << ")";
-            m_pending_adds.push_back(PendingShareBatch{
-                std::make_unique<HandleSharesData>(std::move(data)), addr});
+    //
+    // HOLD this lock across the entire mutation body below (mirrors LTC f445db8e).
+    // try_to_lock keeps the IO thread non-blocking — busy => queue + return. Once
+    // acquired we must NOT release until all m_tracker.chain mutations are done:
+    // the prior code released here and ran the body lock-free, letting the
+    // compute-thread clean_tracker() exclusive prune (drop_tails) free chain nodes
+    // mid-mutation -> SIGSEGV (kr1z1s LTC/DGB). Released just before run_think().
+    std::unique_lock lock(m_tracker_mutex, std::try_to_lock);
+    if (!lock.owns_lock()) {
+        // ── Backpressure (V36 livelock defense-in-depth) ──────────────
+        // If think() is wedged/slow the deferred queue could grow without
+        // bound and blow memory. Cap it: over MAX_PENDING_ADDS we DROP the
+        // new batch (peers re-advertise their best share, so dropped shares
+        // are re-requested later) and warn instead of growing unbounded.
+        if (m_pending_adds.size() >= MAX_PENDING_ADDS) {
+            LOG_WARNING << "[ASYNC-DEFER] BACKPRESSURE: pending_adds at cap ("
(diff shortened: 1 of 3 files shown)


frstrtr:
### Build + test results (local Debug, GTest)

**Build:** `c2pool` target links clean, 0 errors (118 MB binary). All test targets link clean.

**Tests — all green (123 total):**
| Suite | Result |
|-------|--------|
| `share_test` (LTC node/chain/tracker, 9 suites) | 43/43 PASSED |
| `test_threading` (incl. new `Phase2HoldsLockAgainstComputePrune`) | 8/8 PASSED |
| `test_template_builder` | 35/35 PASSED |
| `test_doge_chain` | 37/37 PASSED |

**Byte-parity (the fix == canonical):** normalized full-body diff of LTC `processing_shares_phase2` and `notify_local_share` vs `impl/btc/node.cpp` is **empty** — LTC is now byte-identical to the BTC fixed version (and to BCH for `notify_local_share`; the only LTC↔BCH `phase2` delta is BCH's pre-existing no-witness `pack(tx)` tx-hashing, unrelated to the lock discipline and outside the changed region).

**Consensus-neutral:** diff-grep over the changed lines finds no consensus tokens (protocol/share-version, pplns, payout, target, work, coinbase, merkle, difficulty). Every changed line is a lock primitive, a comment, the backpressure branch, or re-indentation of the existing defer-queue log lines.

**Prod status (LTC-DOGE test node, 158.220.92.171):** NOT actively crashing from this race — current c2pool process up 3d15h with **0 restarts** (bare process, not systemd), **zero** SIGSEGV/segfault in journal or dmesg (all-time), and the only core on record is a single **SIGABRT** (2026-07-15, pre-dating the current 07-17 process start — consistent with the RSS-limit `std::abort()` guard, not the kr1z1s SIGSEGV). The race is **latent** on the current deployment; this is preventive hardening. Draft — no merge, no deploy (deploy is an operator tap; per-coin soak recommended first).

PR #59, opened 2026-06-02
Phase 5.8 M3: structured AuxPoW header parser (parse_aux_header)

frstrtr:
## Phase 5.8 M3 — part 1 of 2: structured parser

Replaces the byte-skip DOGE extended-header parser with a **structured** parse that deserializes the AuxPoW proof into the `CAuxPow` / `CMerkleTx` / `CMerkleLink` hierarchy (M2 skeletons) via the standard `core/pack.hpp` surface.

### What changed
- **`auxpow.hpp`** — `parse_aux_header(s, out_aux, has_aux)`: reads 80B base header; deserializes `CAuxPow` when `is_auxpow_version()`, else `SetNull()`. Hand-written `Serialize`/`Unserialize` on `CMerkleLink`/`CMerkleTx`/`CAuxPow` to stay independent of which `SERIALIZE_METHODS` macro is active per-TU.
- **`auxpow_header.hpp`** — `parse_doge_header` / `parse_doge_headers_message` / `parse_doge_block` route through `parse_aux_header` + `ReadCompactSize`; ~210 lines of hand-rolled varint/segwit skip logic removed.
- **`test_doge_chain.cpp`** — 5 new tests: round-trips the proof, `has_aux` gating, batch parse, truncated-tail drop.

### Scope / invariants
- Confined to `src/impl/doge/` + `test/`. **No `src/core/` or `src/impl/bitcoin_family/` touch.**
- ctest **587/587 green** (Linux x86_64; 582 baseline + 5 new).
- Commit GPG-signed.

### ⚠️ Open question — `CAuxPow::check_proof` (the other half of M3)
`check_proof` is still **declared-only**. Implementing it requires verifying the **parent block PoW** + merkle-link checks. Parent-PoW hashing lives in `src/core/pow.hpp` / `src/core/coin_params.hpp` / `src/impl/bitcoin_family/coin/chain_params.hpp` — i.e. correct `check_proof` cannot be done purely inside `src/impl/doge/`, which trips the per-coin isolation invariant. **Flagged to integrator before any core/bitcoin_family touch.** This PR lands the parser half; `check_proof` follows as M3b once the isolation question is resolved.

PR #794, opened 2026-07-21
dash(embedded): DASH-isolated CoinPeerManager — network-standalone coin arm

frstrtr:
## Summary

The DASH embedded coin-daemon arm had **no peer-scoring**: a single static
`--coin-p2p-connect` target (usually the local dashd itself), so it could not
reach the Dash network independently and its network view was **not independent**
of that daemon. This ports a **DASH-isolated** copy of the peer-prioritization
subsystem — the c2pool port of the Python p2pool broadcaster, live today only on
the merged LTC+DOGE path (`src/c2pool/merged/coin_peer_manager.hpp`) — so the
embedded arm reaches Dash on its **own scored, group-diverse peer set**.

This is the **network-standalone gate** for a fully daemonless c2pool-dash.

## Port scope

New `src/impl/dash/coin/coin_peer_manager.hpp` — `dash::coin::DashCoinPeerManager`.
Brought over 1:1 (chain-agnostic p2pool broadcaster behaviour):

- **Source scoring** — addr-crawl peers `+50`, daemon-learned (`coind`) peers
  `-20` (the 70-pt swing that pulls the arm off the dashd peer view).
- **Daemon-peer overlap filter** — a non-`coind` peer already in the daemon's
  `getpeerinfo` set is rejected as a separate entry.
- **Sybil / group diversity** — `/16` (IPv4) `/32` (IPv6) groups, per-group caps,
  max-2-per-group at selection.
- **Anchors** — persisted last-N good peers, boosted on restart.
- **Protected local node** — pinned dashd, `score=999999`, never pruned.
- **Seed discovery** — DNS (`dnsseed.dash.org`:9999) + fixed seeds + HTTP fallback;
  JSON peer-db with atomic write.

**DASH-adapted:** `dash::coin` namespace + renamed types; `valid_ports {9999}` /
`{19999}`, magic `bf0c6bbd`; peer-db `…/<net>/dash_embedded_peers`; HTTP key `dash`;
the merged-child `is_merged` inv branch dropped (DASH is a parent chain).

## Coexistence with `--coin-p2p-connect`

New opt-in **`--coin-p2p-discover`**. The single-connection E1 `CoinClient` is fed
its dial targets from the manager:

- `--coin-p2p-connect` front target → `set_local_node()` = **pinned / protected /
  preferred**, still carries the redundant block-relay leg; coexists with (never
  suppresses) the discovered set.
- Initial dial = `[pinned] ++ get_peers_to_connect()` (scored, diverse); a 60s
  refresh keeps the rotation pool fresh; `getaddr` on handshake + connect/disconnect
  feed scoring/anchors via two new `CoinClient` seams.
- `--coin-p2p-discover` **without** `--coin-p2p-connect` = fully daemonless (seeds only).
- **Neither flag → byte-unchanged** legacy single-peer dial.

Single-connection scope: "diverse peer set" is a scored, group-diverse **rotation
pool** (round-robin on reconnect), not N simultaneous connections — multi-conn is a
follow-up (see below).

## Isolation fence & consolidation follow-up

DASH-local **self-contained copy**, NOT a reuse of the shared merged tree — honours
the single-coin isolation fence that caused the original drop. The duplication
against `src/c2pool/merged/coin_peer_manager.hpp` is a **known #759-class
consolidation follow-up** (unify behind a shared `core/` peer-manager once the fence
can be lifted); **not** resolved here. Multi-simultaneous embedded connections
(as the merged `CoinBroadcaster` does) is tracked with it.

## Oracle-shadow linkage

Source-scoring + daemon-peer-overlap avoidance converge the embedded connection on
peers the local dashd is *not* using → **independent mempool/relay/tip view**. That
independence is (1) the soundness premise for `--embedded-oracle-shadow` and (2) the
prerequisite for the **network-standalone graduation gate** — disabling the external
dashd. Design note lands in `frstrtr/the`
(`docs/DASH-EMBEDDED-COIN-PEER-MANAGER-PORT.md`), companion to the motivating
provenance audit.

## Consensus-neutral / isolation

DASH-only (`src/impl/dash/**` + `main_dash.cpp` + one KAT). No shared-core edit, no
other-coin edit, no share/consensus surface.

## Verification

- `c2pool-dash` builds clean; `--selftest` unchanged.
- KAT `test/test_dash_coin_peer_manager.cpp` (folded into the allowlisted
  `test_dash_p2p_node` target): daemon-penalty scoring swing, selection ordering,
  `/16` Sybil cap, anchor persistence across reload, pinned-node prune survival,
  canonical seed sanity. **22/22 in the target pass.**

---

DRAFT — Design review precedes any tap. No merge/deploy.

### DIFF
--- a/src/c2pool/main_dash.cpp
+++ b/src/c2pool/main_dash.cpp
@@ -55,6 +55,8 @@
 #include <impl/dash/coin/rpc_conf.hpp>     // dash.conf creds resolution (rpcpassword off argv)
 #include <impl/dash/coin/node_interface.hpp>
 #include <impl/dash/coin/p2p_client.hpp>   // dash::coin::p2p::CoinClient — OPT-IN coin-network dial (E1, --coin-p2p-connect)
+#include <impl/dash/coin/coin_peer_manager.hpp> // dash::coin::DashCoinPeerManager — DASH-ISOLATED scored/diverse peer discovery (--coin-p2p-discover; network-standalone gate)
+#include <impl/dash/coin/chain_seeds.hpp>  // dash::coin::dash_dns_seeds / dash_fixed_seeds — DASH mainnet/testnet seed bootstrap
 #include <impl/dash/coin/won_block_dispatch.hpp> // dash::coin::broadcast_won_block — S8 dual-path won-block dispatcher (embedded P2P primary + submitblock RPC backup)
 #include <impl/dash/coin/zmq_tip_notify.hpp> // dash::coin::TipHashDedup / ZmqHashblockSubscriber — dashd ZMQ hashblock INSTANT tip-notify (opt-in, hardening on the #770 poll)
 #include <impl/dash/coin/coin_p2p_magic.hpp>      // dash::coin::select_coin_p2p_magic — E5 --coin-p2p-magic override (regtest ARM A dial)
@@ -205,7 +207,7 @@ void print_banner(const char* argv0)
         << "       " << argv0 << " --run [--coin-rpc H:P] [--coin-rpc-auth PATH]\n"
         << "           [--testnet] [--submit-block HEX | --submit-block-file PATH]\n"
         << "           [--listen [HOST:]PORT] [--addnode HOST:PORT]... [--connect HOST:PORT]...\n"
-        << "           [--stratum [HOST:]PORT] [--coin-p2p-connect HOST:PORT]...\n"
+        << "           [--stratum [HOST:]PORT] [--coin-p2p-connect HOST:PORT]... [--coin-p2p-discover]\n"
         << "           [--web-port PORT] [--web-host ADDR] [--dashboard-dir PATH]\n"
         << "           [--external-ip ADDR]\n"
         << "           [--embedded-utxo]\n"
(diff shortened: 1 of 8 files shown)


frstrtr:
## CR fixes pushed (e96bfe5a) — still DRAFT for re-review

Both required fixes applied, both in the `--coin-p2p-discover` opt-in path.

**CR-1 — daemonless cold-start dial wedge.** `CoinClient::connect()` now arms the 30s reconnect loop **unconditionally** (empty-plan guard on the tick so `current()/advance()` never throw), and `update_dial_targets()` **kicks an immediate dial on the empty→non-empty transition**. So `--coin-p2p-discover` without `--coin-p2p-connect`, first run + DNS down (empty initial dial), now retries and connects as soon as fixed seeds (t+60s) / HTTP (t+90s) arrive — no restart, no wedge. Extracted `arm_reconnect_timer()`. Regression pin: `DashCoinP2PClient.empty_connect_arms_without_wedging`.

**CR-2 — dormant overlap filter + coind −20 penalty.** Added `NodeRPC::getpeerinfo()` to the DASH RPC (parses each entry's `addr`, IPv6-bracket aware) and wired `set_getpeerinfo_fn` in the discover path when a dashd RPC is armed. `m_coind_peers` now populates → the −20 penalty **and** the overlap filter engage (active daemon-disjointness in the connect+discover / oracle-shadow deployment). Absent RPC (fully daemonless) = seeds-only passive independence. Mirrors `main_ltc.cpp`.

**Nits:** pinned target no longer duplicated in the dial list (pinned keys passed as the `get_peers_to_connect` exclusion set); dead `is_daemon_overlap_peer()` removed; declaration/destruction order corrected so the manager outlives the client + refresh timer that borrow it.

Build clean; `test_dash_p2p_node` **23/23** (was 22 + the CR-1 pin); `--selftest` unchanged. Design note updated in frstrtr/the PR #59 (§5a). Ready for re-review.

PR #802, opened 2026-07-22
dash(embedded): independent credit-pool advance on ingestion post-restart [E2 — full independence]

frstrtr:
### The gap
`m_credit_pool` advanced **only** when the periodic `mnlistdiff` re-seeded it from the diff's cbTx, and was **never persisted**. After a warm restart the SML resumes from its persisted tip (#800 / `SMLDb`+`QuorumDb`), but the credit pool re-seeds to nothing until the *next* `mnlistdiff` — so for the tip that existed **at restart** the credit-pool freshness gate fails closed and the daemonless arm falls back to `dashd` (the restart-seed case). `CreditPool` (`credit_pool.hpp`) and `CreditPoolDb` (`credit_pool_db.hpp`) existed but were entirely unwired.

### Mechanism (two independent legs)
- **Per-block advance** — `CoinStateMaintainer::on_block_connected` now folds *this block's own* credit-pool accrual (platform-reward + Σ assetLocks − Σ assetUnlocks) through a running `CreditPool` state machine, so the balance tracks the tip on **every ingested block**, not only at `mnlistdiff` cadence. Wired ahead of the MN-snapshot fence (distinct axis). `CreditPool::apply_block` gains the platform-reward accrual term.
- **Restart persistence** — `CreditPoolDb` is written per accepted `mnlistdiff` with the **same `(blockHash, height)`** the SML persist uses, so its tip tracks `SMLDb`. On warm restart, when the persisted credit-pool tip matches the SML tip (sentinel), both the `NodeCoinState` freshness seed **and** the running accrual are restored — the gate is live at the restart tip and the first ingested block advances contiguously. Reorg/heal wipes the accrual + the persisted store alongside SML/quorum.

### The INDEPENDENT (non-self-referential) verify
The 3 prior soaks were refuted because a self-referential check (`built == seed + reward`) proves the template is internally consistent — **not** that the seed is correct — so a stale seed sailed through.

Here, every ingested block commits, **in its own coinbase CCbTx**, dashd's `creditPoolBalance` for that height. The running accrual is advanced across the block and **required to equal that from-wire committed value at `nHeight == connected height`**. That compares two genuinely independent derivations — our recomputation vs the network's committed value at the same height — never a template against its own seed. The seed height is taken straight off the wire, so it can never be mistaken as current at an unobserved height (the exact stale-seed trap).

**Invariant preserved:** on drift the freshness seed is **not** advanced (fail closed to the reward-safe `dashd` fallback) and the accrual re-anchors to the wire value. The fix makes the advance *correct*, it does **not** relax the gate. The pre-emit GBT-xcheck + fail-closed-to-fallback remain as the maturation safety net.

### Restart KAT (folded into `test_dash_embedded_gbt`)
`E2CreditPoolIndependentAdvancePostRestart` models the sequence:
1. cold snapshot seeds + **persists** the pool @ P;
2. warm restart **without** restore → gate fails closed at the restart tip (the bug reproduced);
3. warm restart **with** restore → serves at the restart tip;
4. first incremental block @ P+1 advances the pool to its from-wire committed value — cross-checked against an **independently re-derived** platform reward (raw dashcore `subsidy*3/4*375/1000`), not self-referentially.

`E2CreditPoolDriftFailsClosedNotSelfReferential` proves the verify is discriminating: an off-by-one **bad-cbtx** block fails closed (seed stays at P), while the correct block verifies and advances to H.

### Build + tests
- `c2pool-dash` builds + links clean.
- `test_dash_embedded_gbt`: **25/25** pass (incl. 2 new E2 KATs).
- `test_dash_coin_state_maintainer`: **11/11** pass (no regression on the existing SML/credit-pool suites).

### Files
- `src/impl/dash/coin/credit_pool.hpp` — `apply_block` platform-reward accrual term.
- `src/impl/dash/coin/coin_state_maintainer.hpp` — per-block advance + independent verify, mnlistdiff re-anchor + aligned persist, restore + persist hooks, reorg wipe.
- `src/c2pool/main_dash.cpp` — `CreditPoolDb` open / warm-restore (sentinel vs SML tip) / persist / clear wiring.
- `test/test_dash_embedded_gbt.cpp` — restart + drift KATs.

**Do NOT merge** — integrator review + soak gate (this area is soak-sensitive; empirical soak is authoritative over review here).

### DIFF
--- a/src/c2pool/main_dash.cpp
+++ b/src/c2pool/main_dash.cpp
@@ -1569,6 +1569,21 @@ int run_node(bool testnet, const std::string& rpc_endpoint,
                 });
         }
 
+        // E4 re-soak fix (constant −66,966,830-duff creditPool bias): the
+        // DIP-0027 platform-share accrual is gated on the NETWORK'S MN_RR
+        // activation height (per-chainparams in dashcore: mainnet 2,128,896,
+        // testnet 1,066,900 — buried, cross-checked live via getblockchaininfo
+        // softforks). With the mainnet constant in force on testnet the
+        // platform reward evaluated to 0 for every current testnet height, so
+        // the embedded template committed creditPoolBalance(N-1) + 0 — exactly
+        // one block's platform reward LOW, at every height, surviving restart
+        // (the persisted seed was correct; the accrual term was the bias).
+        // This threads the network's height into the template build, the
+        // pre-emit value re-check, and the per-block credit-pool advance.
+        node_coin_state.set_mn_rr_height(
+            testnet ? dash::coin::DASH_MN_RR_HEIGHT_TESTNET
+                    : dash::coin::DASH_MN_RR_HEIGHT_MAINNET);
+
         // review PR #780 BLOCKER-1 (CRITICAL): refuse the embedded arm on DKG
         // commitment-window heights. There the block MUST carry mandatory type-6
         // quorum-commitment txs (which the C-3 filter strips) and merkleRootQuorums
--- a/src/c2pool/main_dash.cpp
+++ b/src/c2pool/main_dash.cpp
@@ -63,6 +63,7 @@
 #include <impl/dash/coin/header_chain.hpp>       // dash::coin::HeaderChain — SPV header/tip authority (E2a)
 #include <impl/dash/coin/coin_state_maintainer.hpp>  // dash::coin::CoinStateMaintainer — populate ordering gate (E2a)
 #include <impl/dash/coin/sml_quorum_db.hpp>      // dash::coin::SMLDb / QuorumDb — SML+quorum persistence (incremental restart)
+#include <impl/dash/coin/credit_pool_db.hpp>     // dash::coin::CreditPoolDb — credit-pool tip persistence (E2 restart resume)
 #include <impl/dash/coin/live_feed.hpp>          // E2a live-feed bridge (raw wire events -> derived ingest events)
 #include <impl/dash/coin/mempool_ingest.hpp>     // wire_mempool_ingest (leg 1)
 #include <impl/dash/coin/tip_ingest.hpp>         // wire_tip_ingest (leg 2)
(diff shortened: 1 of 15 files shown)


frstrtr:
**E4 re-soak refutation → root cause found + fixed (new commit): the constant −66,966,830-duff creditPool delta was the MN_RR activation gate, not the seed.**

**Root cause.** `compute_dash_platform_reward_post_v20_mn_rr` gated the DIP-0027 platform share on the MAINNET MN_RR height (2,128,896) unconditionally. On TESTNET, MN_RR is buried at **1,066,900** (confirmed live on the soak oracle .52: `getblockchaininfo` softforks → v20@905100, mn_rr@1066900), so every current testnet height (~1.519M) got platform reward = **0**. The real per-block share at those heights is exactly **66,966,830 duffs** (238,104,286 block reward × 3/4 × 375/1000) — the soak's exact constant delta:

- template build committed `seed(N−1) + 0` → dashd's H−1 value at every height H (verified: emb(1519414) == dashd getblock(1519413) == 34,026,856,819,946 exactly);
- the E2 per-block advance under-stepped by the same term → `[CREDITPOOL] ACCRUAL DRIFT` fired on EVERY block (`computed(h) == from-wire(h−1)`, 22/22 in the soak log) and re-seeded from the wire, which is why the offset never accumulated;
- the persisted seed was actually CORRECT (restored 34,027,327,587,756 == dashd's true balance @1519420) — the bias re-entered per template, so restart "restored the same wrong offset";
- side effect also repaired: on testnet the coinbase payment split now matches dashd's GBT (`[66966830 burn, 111611384 payee]`) — previously no OP_RETURN burn output and MN overpaid by the platform share.

**Fix** (mostly-reuse — dashcore keys this off `Params().GetConsensus().MN_RRHeight`, per-chainparams): `DASH_MN_RR_HEIGHT_TESTNET = 1,066,900` added; the reward function takes the network's `mn_rr_height` (mainnet default, existing callers unchanged); threaded via `NodeCoinState::set_mn_rr_height` into the template build, the pre-emit value re-check, and `advance_credit_pool_on_block`; `main_dash` sets it network-aware beside the superblock-cycle wiring.

**New KAT** — `DashCreditPoolSeedAnchorKat` (in `test_dash_embedded_cbtx_byte_parity.cpp`): asserts the **ABSOLUTE** creditPoolBalance against REAL from-wire testnet cbTx payloads (consecutive coinbase-only blocks 1,519,413/1,519,414 captured off .52). Fails-before verified empirically: with the pre-fix mainnet-only gate, 3/4 assertions fail by exactly 66,966,830 (`TestnetPlatformRewardEqualsOnChainStep`, `BuiltTemplateCommitsDashdAbsoluteValue`, `PerBlockAdvanceLandsOnFromWireAbsolute`); with the fix the suite binary passes 30/30. Every prior credit-pool KAT tested deltas/self-consistent round-trips only — this one anchors the absolute value, closing the 4-refutation hole.

Still **Do-NOT-merge** — awaiting the E4 re-soak on the fixed binary (integrator merges).

PR #803, opened 2026-07-22
dash(embedded): daemonless type-6 quorum + superblock sourcing at DKG/superblock heights [E1 — full independence]

frstrtr:
## What this closes (E1)

The embedded DASH template arm FAILED CLOSED to the dashd fallback at every DKG-commitment-window height — the `(24,[10,18])` window alone is 37.5% of all heights, the full union ~44% — because it could not produce block-N's OWN mandatory type-6 LLMQ quorum commitments (`bad-qc-missing` + wrong `merkleRootQuorums`). This PR serves those heights daemonlessly. Superblock heights (1 in 16616) keep the reward-safe refusal (see "Superblocks" below).

## The daemonless design (reuse-first: every rule is dashcore's own, at the vendored pin cfad414)

`src/impl/dash/coin/dkg_commitments.hpp` (new) ports, without re-derivation:

1. **Which commitments are mandatory** — `llmq/blockprocessor.cpp GetNumCommitmentsRequired/IsMiningPhase`: per enabled llmqType, inside `[cycleStart+windowStart, cycleStart+windowEnd]`, one commitment per quorumIndex (`rotation ? signingActiveQuorumCount : 1`) whose quorum base block (`cycleStart+index`) exists and whose commitment is not yet mined. Params table + per-network enabled sets (AddLLMQ order — also the miner's qc-tx emission order) copied verbatim from `llmq/params.h` + `chainparams.cpp`, with the same re-diff-on-pin-bump maintenance rule as `dkg_window.hpp`.
2. **`HasMinedCommitment` without dashd** — the mnlistdiff-fed `QuorumManager` IS `GetMinedAndActiveCommitmentsUntilBlock(tip)` (that is exactly what `evo/smldiff.cpp BuildQuorumsDiff` serializes, inclusive of commitments mined AT the tip), and embedded viability already requires the SML/quorum state current at the tip we build on. A commitment mined earlier in the current window is necessarily still active, so `has_mined == qmgr.find(type, baseHash)`.
3. **Content** — the consensus-valid **NULL commitment**, i.e. dashcore's own `GetMineableCommitments` "null commitment required" arm (`CFinalCommitment(params, quorumHash)`, version `GetVersion(rotation, basic_bls)`): exactly what dashd itself mines when it holds no verified DKG result. `ProcessBlock` requires `numInBlock == numRequired` and accepts nulls via `VerifyNull`; `CalcCbTxMerkleRootQuorums` **skips** nulls.
4. **`merkleRootQuorums`** — unchanged by an all-null set, so the committed root is the PROVEN from-wire `compute_merkle_root_quorums` (#780). The full `evo/cbtx.cpp` **with-block fold** (non-rotated oldest-eviction at `signingActiveQuorumCount` capacity — ordered by quorum-base height off the header chain, which equals mined order per type; rotated per-`quorumIndex` replacement; null skip) is implemented and KAT'd now, so Phase-L real commitments only flip a verifier.
5. **Placement** — qc txs immediately after the coinbase, before mempool txs (`node/miner.cpp CreateNewBlock` position), zero-fee. Also fills `m_tx_data_hex` on the embedded arm (previously only the dashd-GBT parser filled it — the submit-time block body source).

**Gating**: the plan installs only where the embedded arm is enabled (testnet, or mainnet behind the existing `--embedded-mainnet` opt-in, default OFF — graduation stays soak-gated). Below a per-network **V19 serve floor** (mainnet 1899072 / testnet 850100 — also covers `--regtest` chains riding the testnet flag, where a qc tx would be `bad-qc-premature`) the PHASE-1 refusal is preserved exactly. Any underivable height (header gap at a quorum base, unorderable eviction) => `nullopt` => fail closed to the dashd fallback — never a wrong block.

**Invariant enforcement**: the pre-emit HARD GATE (`NodeCoinState::embedded_template_emit_ok`) re-derives the same deterministic plan and discards any template whose type-6 payload set (count, bytes, order) or committed quorum root drifts — a `bad-qc-missing`/`bad-qc-not-allowed`/wrong-root block cannot leave the node. The GBT-xcheck backstop and the dashd fallback arm are untouched.

## BLS / Phase-L determination (reuse-first, with evidence)

**Question**: can we vendor/reuse Dash Core's own quorum-commitment sourcing + verification and its BLS backend inside the embedded arm, rather than hand-rolling?

**Answer: YES — and E1 needs none of it to be consensus-valid at window heights.** The null-commitment path above is itself reused dashd miner behavior and requires no BLS. For REAL (non-null) commitments — matching dashd's template byte-for-byte when a DKG succeeded — the concrete reuse plan, all seams landed in this PR:

1. **Sourcing (landed here)**: dashd relays every VERIFIED final commitment to all peers as inv `MSG_QUORUM_FINAL_COMMITMENT` (21) + `qfcommit` (`llmq/blockprocessor.cpp ProcessMessage/AddMineableCommitment`) — the exact stream its own miner mines from. This PR pulls the inv, parses the (already-vendored) `CFinalCommitment`, and feeds `MineableCommitmentCache` with full structural validation (`VerifySizes`, thresholds, version, non-null crypto fields — every `CFinalCommitment::Verify` check except the BLS math). `[QC-MINEABLE]` is the live observability signal.
2. **Verification (the exact remaining piece — scoped, not vague)**: vendor, at the same pin and in the same `vendor/` pattern as `llmq_commitment.hpp`:
   - `llmq/commitment.cpp` — `CFinalCommitment::Verify(checkSigs=true)` + `BuildCommitmentHash` (the signed preimage);
   - `llmq/utils.cpp` — `ComputeQuorumMembers`/`CalculateQuorum` (deterministic member selection: score = hash(proTxHash, confirmedHash, modifier); the SML we already sync carries proRegTxHash + confirmedHash + pubKeyOperator — every input, for non-rotated types);
   - `llmq/snapshot.{h,cpp}` — `CQuorumRotationInfo`/`CQuorumSnapshot` + `ComputeQuorumMembersByQuarterRotation` for ROTATED types, fed from the `qrinfo`/`getqrinfo` light-client protocol on the SAME coin-P2P client that already speaks `getmnlistd`/`mnlistdiff` (qrinfo carries the cycle-base mnlistdiffs = the historical-SML input member selection needs);
   - **BLS backend**: link `dashpay/bls-signatures` ("bls-dash", the Chia-BLS fork dashd itself wraps via `src/bls/bls.h`), relic-backed, CMake, Apache-2.0 (engine-license compatible). **Verified absent today**: no BLS lib in `conanfile.txt` (boost/nlohmann/yaml-cpp/gtest/leveldb/zeromq only) or `vendor/`; `vendor/llmq_commitment.hpp` adaptation note 3 explicitly defers point decompression/verification to Phase L.
3. **Seam (landed here)**: `MineableCommitmentCache::set_bls_verify_fn` — Phase L installs the vendored verifier there; nothing else changes (`verified_for` returns nothing until then, so unverified relayed commitments are never mined — a malicious peer cannot hand us a `bad-qc-invalid` block).
4. **Residual after that**: none for the miner role. c2pool never PARTICIPATES in DKG (that requires masternode operator keys and is not a miner concern); it only verifies-and-includes, which the vendored `Verify` + member computation + bls-dash cover.

Consequence of the interim null-only posture: on the rare block where we win a window height AND a real commitment existed, that quorum activates a few blocks later via another miner (dashd accepts our null block as valid; windows are 9 blocks). Phase L removes even that.

## Superblocks (analogous gap, separately scoped)

Still refused (reward-safe fallback). The daemonless piece is NOT BLS: it is governance-trigger sourcing — vendor `governance/classes.cpp CSuperblock::GetPayeeData`/`CSuperblockManager::GetSuperblockPayments` and sync the trigger gobjects + votes over coin-P2P (`govsync`/`govobj`/`govobjvote`, ECDSA-signed votes tallied against the DMN set we already hold). Follow-up scope, unchanged behavior here.

## What's daemonless NOW vs still-fallback

| Height class | Before | After |
|---|---|---|
| Plain heights | embedded | embedded (unchanged) |
| DKG windows, all cycle commitments already mined | REFUSED | embedded (plain template, zero qc txs) |
| DKG windows, commitments pending | REFUSED | embedded with mandatory null qc txs |
| DKG windows, underivable (header gap / below V19 floor / --regtest) | REFUSED | refused (unchanged fail-closed) |
| Superblock heights | REFUSED | refused (scoped follow-up above) |

## Byte-parity KAT (folded into CI-allowlisted `test_dash_embedded_gbt`)

`test/test_dash_dkg_commitments.cpp` (14 tests): the captured testnet `mnlistdiff` @1518412 -> `QuorumManager` -> daemonless plan at **window height 1518420** -> 3 mandatory commitments (types 1/4/6, correct order/hashes) -> `merkleRootQuorums` **byte-equal to the root a real dashd committed** (`1901c172…`); dashcore-parity slot math incl. the 35-slot rotated fan-out at phase 42 (50_60 + 60_75 x32 + 400_85 + 100_67); null-commitment/qc-tx wire-shape KATs (323/342-byte vectors, version|type dword); below-floor refusal proven byte-identical to the PHASE-1 guard over 600 consecutive heights; with-block fold eviction/replacement vs hand-built leaf sets; cache admission + BLS gate. Plus an end-to-end `NodeCoinState` test: the exact height the BLOCKER-1 test proves refused without a plan is now served, the emit gate passes the honest template and discards a tampered one (dropped commitment, wrong root).

## Verification

- `test_dash_embedded_gbt` 37/37 (incl. all prior byte-parity suites), `test_dash_node_coin_state` 16/16 (new end-to-end), plus green: work_source (9), stratum_work_source (44), get_work (4), p2p_messages (8), p2p_node (16), node_reception_wire (21), node_embedded_wire (3), embedded_relay_e2e (4), coin_state_maintainer (11), block_producer (9), quorum_root (11), quorum, p2p_connection (4).
- `c2pool-dash` builds clean; smoke-run OK.
- Default-path safety: no qc plan fn installed => every touched surface is byte-for-byte pre-E1 (KAT'd); the dashd fallback + GBT-xcheck safety net stays throughout maturation.

Do NOT merge — integrator review + soak gate (v37-dev steward) before merge.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,6 +105,50 @@ else()
     set(SECP256K1_INCLUDE_DIRS "")
 endif()
 
+# ── E1 Phase-L: optional dashbls (BLS12-381) backend for type-6 quorum-
+# commitment verification ────────────────────────────────────────────────────
+# dashpay/bls-signatures ("dashbls", relic-backed, Apache-2.0) — the SAME
+# library dashd wraps in src/bls/bls.h. OPT-IN and DASH-ONLY: only the DASH impl
+# (dash_bls_verify) + its KAT consume it, never the shared core the Windows
+# main_ltc launcher builds ([[reference_windows_ci_conan_flake]]: Windows CI
+# builds only the LTC launcher, so DASH-only BLS code cannot break it). When the
+# option is OFF (default) or the library is not found, dash_bls_verify compiles a
+# fail-closed stub and the DKG-window serve path keeps the pre-Phase-L
+# null-commitment posture exactly — no dashbls dependency enters any build.
+#
+# Build the library once for the CI factory with scripts/build_dashbls.sh (pins
+# the exact upstream commit); point cmake at it with -DDASHBLS_ROOT=<prefix>.
+option(C2POOL_DASH_BLS "Link dashbls (BLS12-381) for E1 Phase-L real quorum-commitment verify" OFF)
+set(C2POOL_DASH_BLS_AVAILABLE OFF)
+if(C2POOL_DASH_BLS)
+    # Honour a DASHBLS_ROOT hint (the build-factory install prefix), else search
+    # the system paths. dashbls installs: include/dashbls/*.hpp,
+    # lib/libdashbls.a, lib/librelic_s.a, lib/mimalloc-2.0/libmimalloc-secure.a.
+    find_path(DASHBLS_INCLUDE_DIR dashbls/bls.hpp
+        HINTS ${DASHBLS_ROOT} ENV DASHBLS_ROOT PATH_SUFFIXES include)
+    find_library(DASHBLS_LIB       NAMES dashbls bls-dash
+        HINTS ${DASHBLS_ROOT} ENV DASHBLS_ROOT PATH_SUFFIXES lib)
+    find_library(DASHBLS_RELIC_LIB NAMES relic_s relic
+        HINTS ${DASHBLS_ROOT} ENV DASHBLS_ROOT PATH_SUFFIXES lib)
+    find_library(DASHBLS_MIMALLOC_LIB NAMES mimalloc-secure mimalloc
+        HINTS ${DASHBLS_ROOT} ENV DASHBLS_ROOT PATH_SUFFIXES lib lib/mimalloc-2.0 lib/mimalloc-2.1)
+    find_library(DASHBLS_GMP_LIB NAMES gmp)
+    if(DASHBLS_INCLUDE_DIR AND DASHBLS_LIB AND DASHBLS_RELIC_LIB AND DASHBLS_GMP_LIB)
+        set(C2POOL_DASH_BLS_AVAILABLE ON)
+        set(DASHBLS_LIBRARIES ${DASHBLS_LIB} ${DASHBLS_RELIC_LIB})
+        if(DASHBLS_MIMALLOC_LIB)
+            list(APPEND DASHBLS_LIBRARIES ${DASHBLS_MIMALLOC_LIB})
+        endif()
+        list(APPEND DASHBLS_LIBRARIES ${DASHBLS_GMP_LIB})
+        message(STATUS "dashbls found (E1 Phase-L BLS verify ENABLED): ${DASHBLS_LIB}")
+        message(STATUS "dashbls includes: ${DASHBLS_INCLUDE_DIR}")
+    else()
+        message(WARNING "C2POOL_DASH_BLS=ON but dashbls not found "
+            "(need dashbls + relic + gmp; run scripts/build_dashbls.sh and pass "
+            "-DDASHBLS_ROOT=<prefix>). Falling back to fail-closed null-serve.")
+    endif()
+endif()
+
 # Enable testing only if GTest is found
 if(GTest_FOUND OR GTEST_FOUND)
     enable_testing()
--- a/scripts/build_dashbls.sh
+++ b/scripts/build_dashbls.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: AGPL-3.0-or-later
+#
+# Build + install dashpay/bls-signatures ("dashbls") for the c2pool E1 Phase-L
+# BLS quorum-commitment verifier. This is the SAME library dashd wraps in
+# src/bls/bls.h (libdashbls 2.0, relic-backed, Apache-2.0). It is NOT on
+# ConanCenter, so — like secp256k1 — c2pool consumes it as a prebuilt system
+# library (find_library in the root CMakeLists, gated by -DC2POOL_DASH_BLS=ON).
+# Run this once on the CI build factory / dev host, then configure c2pool with:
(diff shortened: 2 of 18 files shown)


frstrtr:
CI rollup: 25 pass / 1 fail. The single red — Windows x86_64 "Verify binary" (c2pool.exe --help smoke on the LTC/DOGE binary) — is PRE-EXISTING: master's latest build.yml run fails the identical lane (this diff is DASH-only: src/impl/dash + main_dash.cpp + dash tests; the LTC binary path is untouched). All DASH gates green: G1 byte-parity KAT, G3a populated-regtest block production, dash smoke, Linux x86_64 (full ctest allowlist incl. the 14 new DashDkgCommitments KATs + the end-to-end NodeCoinState window-serve/emit-gate test), and the 1h23m ASan+UBSan lane.

frstrtr:
Superseded — this stack was squash-merged into master via #826 (Phase-L daemonless quorum-commitment serving). Content is on master; closing to avoid a duplicate merge. Soak-validated (#816 reward-safety soak: 130 serves / 0 bad-cbtx).

PR #812, opened 2026-07-23
dash(embedded): integrate bls-signatures + E1 Phase-L real quorum-commitment verify

frstrtr:
**DO NOT MERGE** — draft; integrator merges after review + soak. Stacked on #803 (E1); base = `dash/e1-daemonless-dkg-window`.

## What
The BLS backend + the cryptographic verify that E1 (#803) cut its `MineableCommitmentCache::set_bls_verify_fn` seam for. E1 today serves the consensus-valid **null** type-6 commitment at DKG-window heights. On **mainnet** a DKG usually **succeeds**, so the canonical block carries the **real** (non-null) commitment and null-serve would diverge → `bad-qc` reject (the GLM minority-report concern). Phase-L sources the real commitment (E1 already relays `qfcommit` over coin-P2P), **verifies it with BLS reusing Dash Core's own logic**, and makes it includable — true daemonless at DKG windows.

## Part A — BLS library integration
- **dashpay/bls-signatures ("dashbls", relic-backed, Apache-2.0)** — the exact library dashd wraps in `src/bls/bls.h` (libdashbls 2.0 = the dash v23.1.x `src/dashbls` subtree). Not on ConanCenter, so consumed as a **prebuilt system library like secp256k1**: `scripts/build_dashbls.sh` pins the upstream commit; the root `CMakeLists.txt` `find_library`s it behind `-DC2POOL_DASH_BLS=ON` (`-DDASHBLS_ROOT=<prefix>`).
- **DASH-only + Windows-safe:** only `dash_bls_verify` (linked solely by `c2pool-dash`) and its KAT consume it. Option **OFF by default**; when off or the lib isn't found, `dash_bls_verify` compiles a **fail-closed stub** and **no dashbls dependency enters any build** — verified: the default configure reports no dashbls and the stub object carries zero dashbls/relic symbols, so the Windows `main_ltc` launcher (shared core only) is untouched.
- **BLS smoke KAT** proves the lib links + signs + verifies.

## Part B — E1 Phase-L verify (reuse-first; vendored from Dash Core v23.1.x)
- `build_commitment_hash` == `llmq/commitment.cpp` **BuildCommitmentHash** (the signed preimage), **byte-exact** — locked by a from-wire testnet KAT.
- `verify_final_commitment` == **CFinalCommitment::Verify** (checkSigs): `membersSig` `VerifySecureAggregated` over the signers' operator keys **+** `quorumSig` `VerifyInsecure` against `quorumPublicKey`, both over the commitment hash, post-V19 basic scheme.
- `membersSig` is **mandatory** — `quorumSig` alone is a self-signature a peer can forge (it controls `quorumPublicKey`); only the actual members' operator keys prove authenticity.
- Verifier installed via `set_bls_verify_fn` in `main_dash`.
- **Fail-closed throughout** (reward-critical embedded consensus): backend absent, member set uncertain, size mismatch, or any BLS verify failure → `verified_for()` yields `nullopt` → the slot mines the null commitment (or dashd fallback). An unverified/forged relayed commitment is **never** served.

## KATs — `test_dash_bls_verify` (6/6 passing)
Captured **read-only** from the testnet dashd (`192.168.86.52`), block 1519931, llmq_50_60 quorum `0000001b…5c928798` (a **successful** DKG — the mainnet case null-serve diverges on):
- `BuildCommitmentHashRealVector` — preimage of the real commitment hashes to dashd's signed value (runs **with or without** the backend).
- `RealQuorumSigAcceptsAndTamperRejects` — real from-wire `quorumSig` **ACCEPT**; flip a hash/sig byte → **REJECT**.
- `VerifyFinalCommitmentSyntheticRoundTrip` — full `verify_final_commitment` (both legs) on a synthetic valid commitment **ACCEPT**; per-field tamper (`membersSig`/`quorumSig`/`quorumPublicKey`/member key) → **REJECT**.
- `BlsSmokeSignVerify`, `FailClosedNoMembers`, `MakeVerifierNullProviderFailsClosed`.

## Residual (fail-closed until it lands)
The deterministic quorum **member selection** (`llmq/utils.cpp` `ComputeQuorumMembers` over the SML **as of the quorum base block**) needs the historical SML sourced over coin-P2P `qrinfo` — not yet wired. Until then the member-keys provider returns `nullopt` and Phase-L stays at null-serve. The **BLS crypto path is complete + KAT-locked** and needs no change when the member set arrives. The SML `nVersion` (`VER_LEGACY_BLS`/`VER_BASIC_BLS`) disambiguates each member key's scheme (a bare RPC hex is scheme-ambiguous — a mixed quorum has keys valid under both encodings).

## Build / safety
- `c2pool-dash` builds green with `-DC2POOL_DASH_BLS=ON` (verifier symbols linked).
- Default (OFF) configure + `dash_bls_verify` stub build green, **no dashbls symbols** — Windows/`main_ltc` unaffected.
- Running v0.2.4 soaks untouched (all captures read-only on their dashd).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,6 +105,50 @@ else()
     set(SECP256K1_INCLUDE_DIRS "")
 endif()
 
+# ── E1 Phase-L: optional dashbls (BLS12-381) backend for type-6 quorum-
+# commitment verification ────────────────────────────────────────────────────
+# dashpay/bls-signatures ("dashbls", relic-backed, Apache-2.0) — the SAME
+# library dashd wraps in src/bls/bls.h. OPT-IN and DASH-ONLY: only the DASH impl
+# (dash_bls_verify) + its KAT consume it, never the shared core the Windows
+# main_ltc launcher builds ([[reference_windows_ci_conan_flake]]: Windows CI
+# builds only the LTC launcher, so DASH-only BLS code cannot break it). When the
+# option is OFF (default) or the library is not found, dash_bls_verify compiles a
+# fail-closed stub and the DKG-window serve path keeps the pre-Phase-L
+# null-commitment posture exactly — no dashbls dependency enters any build.
+#
+# Build the library once for the CI factory with scripts/build_dashbls.sh (pins
+# the exact upstream commit); point cmake at it with -DDASHBLS_ROOT=<prefix>.
+option(C2POOL_DASH_BLS "Link dashbls (BLS12-381) for E1 Phase-L real quorum-commitment verify" OFF)
+set(C2POOL_DASH_BLS_AVAILABLE OFF)
+if(C2POOL_DASH_BLS)
+    # Honour a DASHBLS_ROOT hint (the build-factory install prefix), else search
+    # the system paths. dashbls installs: include/dashbls/*.hpp,
+    # lib/libdashbls.a, lib/librelic_s.a, lib/mimalloc-2.0/libmimalloc-secure.a.
+    find_path(DASHBLS_INCLUDE_DIR dashbls/bls.hpp
+        HINTS ${DASHBLS_ROOT} ENV DASHBLS_ROOT PATH_SUFFIXES include)
+    find_library(DASHBLS_LIB       NAMES dashbls bls-dash
+        HINTS ${DASHBLS_ROOT} ENV DASHBLS_ROOT PATH_SUFFIXES lib)
+    find_library(DASHBLS_RELIC_LIB NAMES relic_s relic
+        HINTS ${DASHBLS_ROOT} ENV DASHBLS_ROOT PATH_SUFFIXES lib)
+    find_library(DASHBLS_MIMALLOC_LIB NAMES mimalloc-secure mimalloc
+        HINTS ${DASHBLS_ROOT} ENV DASHBLS_ROOT PATH_SUFFIXES lib lib/mimalloc-2.0 lib/mimalloc-2.1)
+    find_library(DASHBLS_GMP_LIB NAMES gmp)
+    if(DASHBLS_INCLUDE_DIR AND DASHBLS_LIB AND DASHBLS_RELIC_LIB AND DASHBLS_GMP_LIB)
+        set(C2POOL_DASH_BLS_AVAILABLE ON)
+        set(DASHBLS_LIBRARIES ${DASHBLS_LIB} ${DASHBLS_RELIC_LIB})
+        if(DASHBLS_MIMALLOC_LIB)
+            list(APPEND DASHBLS_LIBRARIES ${DASHBLS_MIMALLOC_LIB})
+        endif()
+        list(APPEND DASHBLS_LIBRARIES ${DASHBLS_GMP_LIB})
+        message(STATUS "dashbls found (E1 Phase-L BLS verify ENABLED): ${DASHBLS_LIB}")
+        message(STATUS "dashbls includes: ${DASHBLS_INCLUDE_DIR}")
+    else()
+        message(WARNING "C2POOL_DASH_BLS=ON but dashbls not found "
+            "(need dashbls + relic + gmp; run scripts/build_dashbls.sh and pass "
+            "-DDASHBLS_ROOT=<prefix>). Falling back to fail-closed null-serve.")
+    endif()
+endif()
+
 # Enable testing only if GTest is found
 if(GTest_FOUND OR GTEST_FOUND)
     enable_testing()
--- a/scripts/build_dashbls.sh
+++ b/scripts/build_dashbls.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: AGPL-3.0-or-later
+#
+# Build + install dashpay/bls-signatures ("dashbls") for the c2pool E1 Phase-L
+# BLS quorum-commitment verifier. This is the SAME library dashd wraps in
+# src/bls/bls.h (libdashbls 2.0, relic-backed, Apache-2.0). It is NOT on
+# ConanCenter, so — like secp256k1 — c2pool consumes it as a prebuilt system
+# library (find_library in the root CMakeLists, gated by -DC2POOL_DASH_BLS=ON).
+# Run this once on the CI build factory / dev host, then configure c2pool with:
+#
+#     cmake -DC2POOL_DASH_BLS=ON -DDASHBLS_ROOT=<PREFIX> ...
+#
+# Requires: cmake, a C++17 compiler, and libgmp-dev (relic ARITH=gmp backend).
+#
+# Pinned to the exact upstream commit dash v23.1.x vendors as src/dashbls
+# (libdashbls 2.0 / relic 0.5.0). RE-PIN in lockstep with the vendored-dashcore
+# bump that moves vendor/llmq_commitment.hpp (same discipline as dkg_window.hpp).
+set -euo pipefail
+
+DASHBLS_REPO="${DASHBLS_REPO:-https://github.com/dashpay/bls-signatures.git}"
+# libdashbls 2.0 (dash v23.1.x src/dashbls subtree; relic backend, basic+legacy
+# BLS schemes). Pin by commit for a reproducible, offline-cacheable build.
+DASHBLS_COMMIT="${DASHBLS_COMMIT:-dd683653c6eaba7235bc9c600f46bafbb8210291}"
+PREFIX="${1:-${DASHBLS_ROOT:-/opt/dashbls}}"
+BUILD_DIR="${BUILD_DIR:-/tmp/dashbls-build}"
+JOBS="${JOBS:-$(nproc 2>/dev/null || echo 4)}"
+
+echo "[build_dashbls] repo=$DASHBLS_REPO commit=$DASHBLS_COMMIT prefix=$PREFIX"
+
+rm -rf "$BUILD_DIR"
+git clone "$DASHBLS_REPO" "$BUILD_DIR/src"
+git -C "$BUILD_DIR/src" checkout --quiet "$DASHBLS_COMMIT"
+
+cmake -S "$BUILD_DIR/src" -B "$BUILD_DIR/build" \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_INSTALL_PREFIX="$PREFIX" \
+    -DBUILD_BLS_TESTS=0 \
+    -DBUILD_BLS_BENCHMARKS=0 \
+    -DBUILD_BLS_PYTHON_BINDINGS=0 \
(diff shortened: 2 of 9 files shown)


frstrtr:
## Adversarial review — bls-signatures integration + Phase-L real quorum-commitment verify (head afc948a8)

Reviewed for correctness + reward-safety against dashcore v23.1.7 ground truth (`llmq/commitment.cpp`, `llmq/params.h`, `bls/bls.h`, vendored `src/dashbls/src/schemes.cpp`) and against the pinned dashbls `dd683653` sources. Ran both build flavors (`-DC2POOL_DASH_BLS=ON` 6/6, OFF 3/3 stub KATs) and independently re-verified the KAT vector against the live testnet chain.

**Verdict: CHANGES-REQUIRED** — the crypto core is right and the fail-closed posture at this head is airtight, but the port diverges from `CFinalCommitment::Verify` in two structural checks that become block-losing the moment member-set sourcing lands. Both fixes are small; land them in this stack, not in the future sourcing PR.

### Reward-critical findings (ranked)

**1. MUST-FIX: quorumIndex is unauthenticated and unchecked → served malleated commitment = rejected block (rotated types).**
`BuildCommitmentHash` does NOT cover `quorumIndex` (confirmed upstream), so neither BLS leg binds it — it is freely malleable by any relay peer on nVersion-4 (llmq_60_75) commitments. Upstream `Verify` enforces `(base_height % dkgInterval) == quorumIndex`; c2pool checks it nowhere (`ingest` doesn't, `verify_final_commitment` can't, `daemonless_qc_commitments` serves the cached copy as-is). Attack: take the real relayed 60_75 qfcommit, flip `quorumIndex`, relay it first → `ingest` caches it, the honest copy is then refused by the keep-best dedup (`CountSigners()` equal → "already hold equal-or-better"), BLS verifies fine (preimage unchanged) → once member sourcing lands we serve it → every dashd rejects the block (`bad-qc-invalid`). Every 60_75 window block from the node is lost, and the cache stays poisoned for the quorum's lifetime.
Fix: in `daemonless_qc_commitments`, reject/null-serve when `real->quorumIndex != s.quorum_index` (the slot already carries the expected index) — or plumb the expected index into `verified_for`. Consider also allowing cache replacement when the held entry mismatches the expected index, so the honest copy can land.

**2. MUST-FIX: threshold vs minSize — `ingest` is looser than dashcore's Verify.**
Upstream requires `CountValidMembers() >= minSize` AND `CountSigners() >= minSize` (50_60: minSize=40, threshold=30; 60_75: 50/45; 400_60: 300/240; 400_85: 350/340; 100_67: 80/67; 25_67: 22/17). `ingest` checks `>= threshold`. A commitment with e.g. 30–39 signers on 50_60 passes ingest AND passes BLS (threshold-many colluding operators can produce both legs legitimately) but is `bad-qc-invalid` to every dashd → lost block. Exotic (needs ≥threshold colluding members) but a strict consensus divergence with a one-line cost: add `min_size` to `LlmqParamsView` (verbatim from `llmq/params.h`) and enforce it in `ingest` (keep the non-null-crypto checks as-is).

**3. SHOULD-FIX: the `signer_pubkeys.size() == 1` special case is a misport.**
Upstream branches on `llmq_params.is_single_member()` — a **params** property (`size == 1`), which NO enabled mainnet/testnet type has (verified: zero `size = 1` rows) — not on signer count. And dashbls `CoreMPL::VerifySecure` has no n==1 shortcut (the t-coefficient is applied even for one key), so plain `Verify` on a lone signer ≠ what dashd computes on the aggregate path. Unreachable today (ingest floor ≥ 17 signers; ≥ minSize after fix 2), but it is wrong-shaped and one refactor away from being load-bearing. Remove the branch (always `VerifySecure`), or key it on `p.size == 1` if you want the upstream shape.

**4. NOTE (availability, not validity): unverified-first-wins cache = trivial null-serve DoS.**
`ingest` admits structurally-valid-but-BLS-unverified commitments and the dedup refuses equal-`CountSigners` replacements, so any peer can poison a slot with junk sigs (all-ones signers, garbage 96 bytes) before the real one arrives → `verified_for` fails BLS → permanent null-serve for that quorum. Null-serve is consensus-valid (dashd's own no-DKG-result behavior), so no bad block — but it silently defeats Phase-L's purpose. Follow-up (with member sourcing): verify at ingest like dashd's `minableCommitments` (which only stores post-Verify commitments), or evict/replace entries that fail BLS. Also note `ingest` keys by attacker-chosen `quorumHash` with no base-header existence check → unbounded map growth under spam; worth a bound or a known-quorum-base filter.

### The sharp questions, answered

**Q1 — preimage byte-exactness: PROVEN, not just claimed.** Port matches upstream `BuildCommitmentHash` field-for-field (llmqType u8, quorumHash raw-LE 32, DYNBITSET validMembers = CompactSize(nBits) + ceil(n/8) LSB-first — checked against `DynamicBitSetFormatter`, pubKey = the commitment's own 48 wire bytes (post-V19 active scheme = basic = the nVersion-3/4 wire encoding ingest admits, so raw-copy is exact), vvecHash, SHA256d). Provenance of the KAT vector independently confirmed: I fetched testnet block 1519931 (`00000041f922df45…`) from the public insight API — quorumPublicKey, quorumSig, quorumHash and vvecHash all appear verbatim in the canonical block's type-6 tx (signers=50/50, validMembers=50/50, nVersion=3, height field 1519931 ✓). Given genuine chain data, `RealQuorumSigAcceptsAndTamperRejects` passing means the real DKG threshold key signed exactly the hash `build_commitment_hash` computes — cryptographic proof, no circularity. Caveat: the vector is all-ones bitsets; the partial-`validMembers`/partial-`signers` edge (mixed bits in the preimage + the signer-subset skip loop) is exercised by NO test. Add a synthetic case with a proper subset of signers and validMembers — it is the cheapest way to catch index-misalignment before the member-set PR.

**Q2 — membersSig aggregation scheme: CORRECT (highest-risk item clears).** `scheme.VerifySecure(pubkeys, sig, msg)` on `BasicSchemeMPL` is the SAME function dashd's `CBLSSignature::VerifySecureAggregated` dispatches to post-V19 (fLegacy=false). Internals (read from the pinned sources): serializes + **sorts** the pubkeys, derives per-key coefficients t_i = Hash256(concat of all sorted pubkeys) per key, verifies against Σ t_i·pk_i — order-insensitive (so the index-order construction, identical to dashd's skip-non-signers loop, is fine) and rogue-key-safe without proof-of-possession, since an attacker's coefficient depends on every key including the honest ones. The "membersSig mandatory / quorumSig alone forgeable" reasoning is airtight: the commitment hash binds an attacker-CHOSEN quorumPublicKey, so quorumSig is self-referential; only the aggregate over SML-sourced operator keys (which the attacker does not control) authenticates. The only wart is the single-signer misport (finding 3).

**Q3 — quorumSig leg: CORRECT.** Single sig, `BasicSchemeMPL::Verify` against the commitment's quorumPublicKey == upstream `VerifyInsecure` post-V19 (no augmentation). Both legs are strictly ANDed (membersSig failure returns before quorumSig; `true` only after both).

**Q4 — fail-closed audit: PASS at this head.** Traced every path: backend absent → compile-time stub `false`; null provider → constant-`false` verifier; provider nullopt → `false`; members empty / size-mismatch → `false`; signer-key or quorum-key deserialization failure (FromBytes throws on invalid points; both wrapped) → `false`; either sig leg fails → `false`; blanket `catch(...)` → `false`. The `deser_g1` try-declared-scheme-then-other fallback deviates from dashd (no fallback there) but only in the reward-safe direction: a wrong-scheme decode yields a different point whose sigs then cannot verify (false-reject → null-serve), and the widened acceptance still requires real operator signatures. Legacy/basic ambiguity cannot produce a false-accept.

**Q5 — cannot mis-serve before sourcing lands: CONFIRMED.** `main_dash` installs a provider that hard-returns `std::nullopt`, so `verified_for` is nullopt for every slot regardless of backend — at afc948a8 the node cannot serve ANY real commitment, only the consensus-valid null. The SML nVersion→`legacy_scheme` plumb-through is correctly shaped (`MemberOperatorKey.legacy_scheme`, documented as MUST-populate-from-SML-nVersion). One divergence to carry into the sourcing PR: upstream tolerates `members.size() < params.size` (undersized MN pool) with a bits-beyond-members check; c2pool's `signers.size() == members.size()` guard would false-reject that case (fail-closed, mainnet-irrelevant, but port the upstream loop when the provider lands).

**Q6 — lib integration: sound, one pin nit + CI gap.** Pin `dd683653` is NOT the byte-identical state dash v23.1.7 vendors as `src/dashbls` (the subtree is a slightly newer refactor: `SerializeToArray`, templated `HashPubKeys`, and `bn_free(order)` leak fixes). I diffed `schemes.cpp` end-to-end: the Verify/VerifySecure/AggregateSecure math and hashing are semantically identical, so this is LOW severity — but `dd683653`'s `HashPubKeys` leaks one relic bn per VerifySecure call (freed upstream); negligible at DKG-window frequency, still worth re-pinning to the actual subtree state at the next vendored-dashcore bump (the script comment claiming "the exact upstream commit dash v23.1.x vendors" is currently inaccurate). Build pattern (prebuilt system lib, default-OFF, DASH-impl-only linkage, unconditional fail-closed stub TU) is clean and keeps the Windows LTC launcher untouched. CI gap: no workflow sets `C2POOL_DASH_BLS=ON`, so the BLS crypto KATs run only on dev hosts — the hash-preimage + fail-closed KATs do run everywhere (verified the stub build passes 3/3). Needs a local-CI build-factory job with the flag ON (+ cached `build_dashbls.sh` output) wired BEFORE the enable gate, as its own follow-up.

### Required soak scope (the actual gate, once member sourcing lands)

Analytical review is not the gate here — this area has refuted review-approved fixes repeatedly. Before `verified_for` can return non-null in production:

1. **Mainnet-representative window crossing with oracle parity**: run against a live dashd oracle across ≥1 full successful-DKG mining window per non-rotated type (50_60/100_67 hourly; 400_60 12h; 400_85 24h) AND ≥1 rotated 60_75 window covering all quorumIndex slots. At every window height: served-commitment bytes == the commitment in dashd's template/canonical block, and merkleRootQuorums parity.
2. **Negative windows**: a failed-DKG window (both sides null), an already-mined-earlier-in-window skip, and a mid-window RESTART (cache is memory-only — verify clean re-arm, no stale serve).
3. **Adversarial injection**: replay real commitments with malleated quorumIndex, junk-sig spam at cache capacity, and sub-minSize crafted commitments — assert every one either fails verify or is refused at ingest, and never reaches a template.
4. **The ultimate gate**: a real block found at a DKG-window height carrying a verified REAL commitment, accepted by the network (dashd-validated), per the established embedded-gate discipline.

Summary: crypto core correct and KAT-anchored to real chain data; fail-closed complete at this head; fix quorumIndex + minSize (and drop the single-signer branch) in this stack before the member-sourcing PR can make them reachable.

PR #814, opened 2026-07-23
dash(embedded): quorum member-set sourcing (qrinfo) — enable E1 Phase-L real-commitment serving

frstrtr:
**DO NOT MERGE** — integrator merges after review + soak. Stacks on #812 (base = its branch) and folds in #812's CHANGES-REQUIRED fixups (cleaner than a concurrent editor on the same files). Fail-closed throughout.

## What this is
The last piece so the embedded arm serves **REAL** verified type-6 quorum commitments instead of null. #812 landed the BLS crypto verify (`CFinalCommitment::Verify`) but its member-key provider returned `nullopt`, so verify stayed fail-closed at null-serve. This provides the ordered member operator-key set it needs.

## Member computation — `vendor/quorum_members.hpp`
Reuses dashcore `llmq/utils.cpp` `ComputeQuorumMembers` + `CDeterministicMNList::CalculateQuorum`, over the SML **as of the quorum base block**:
- **modifier** = `utils::GetHashModifier`. Post-V20 non-rotated form is `SerializeHash(make_tuple(llmqType, workHeight, bestCLSignature))`, `workHeight = quorumBaseHeight-8`, `bestCLSignature` = the coinbase ChainLock of the work block (`GetNonNullCoinbaseChainlock`). **Verified byte-exact against testnet dashd** — reproduces the exact 50-member order dashd reports for `llmq_50_60 @ base 1519920`. (The pre-V20 `hash(llmqType,quorumHash)` form does **not** — V20 is active on testnet since 905100 and the KAT height is well past it. This was the initial gap; the empirical anchor caught it.)
- **score** = `CalculateScores`: `sha256(sha256(proReg||confirmed)||modifier)`, confirmed+valid MNs only, top `size` by descending score.
- **keys** = each selected MN's SML `pubKeyOperator`, `legacy_scheme = (nVersion == VER_LEGACY_BLS)` — the #812 finding; only the per-MN SML nVersion disambiguates a mixed-scheme quorum's key encoding.

Rotated (DIP-24) returns `nullopt` -> null-serve (documented follow-up).

## Historical-SML sourcing — `quorum_member_source.hpp`
The E3 SML tracks the **tip**, not arbitrary heights. `QuorumMemberSource` sources both inputs over the **same coin-P2P client** the E3 sync uses:
- SML @ quorum base block: `getmnlistd(ZERO, quorumHash)` full snapshot;
- ChainLock @ work block: `getmnlistd(ZERO, workBlock)`, `bestCLSignature` off the diff's cbTx (bounded walk-back on a null CL).

Async by necessity — the provider is a pure cache lookup (never blocks the template path); population is kicked off relayed qfcommits. A **demux** in the coin-P2P client routes these historical `base=ZERO` replies **away** from the tip-SML maintainer (which treats any ZERO-base diff as a full snapshot and would overwrite the tip SML to the old block).

Wired into `main_dash`: the `nullopt` provider stub is replaced by the source lookup; qfcommit ingest kicks `request()`; the demux filter is installed. Non-rotated only; anything uncertain -> `nullopt` -> fail-closed.

## KATs (`test_dash_quorum_members`, real from-wire testnet vectors)
- `compute_quorum_members` reproduces dashd's **exact** ordered 50-member set + operator keys + legacy/basic scheme flags;
- the computed set makes #812's `verify_final_commitment` **ACCEPT the real mined commitment** (block 1519931, llmq_50_60) — the end-to-end proof Phase-L now serves a real commitment;
- partial signers/validMembers verify (subset aggregate + bitset-in-hash); member-key corruption + fail-closed (rotated / SML gap / too-small SML) reject.

All KATs pass with `-DC2POOL_DASH_BLS=ON`. `c2pool-dash` builds clean.

## Folded-in #812 fixups (block-losing exactly when this sourcing serves real commitments)
- **MUST-FIX quorumIndex malleation**: `BuildCommitmentHash` doesn't cover `quorumIndex`, so neither BLS leg binds it. `daemonless_qc_commitments` now refuses a cached commitment whose `quorumIndex` != the slot's (dashd's `base_height % dkgInterval == quorumIndex`) -> mines null instead (reward-safe).
- **MUST-FIX minSize**: ingest enforced `>= threshold`; Verify's floor is `minSize` (50_60: 40 vs 30). Added `min_size` to `LlmqParamsView` (exact `llmq/params.h` values) and enforce it.
- **SHOULD-FIX** single-signer branch in `bls_verify.cpp`: upstream keys on `is_single_member()` (params size==1, no enabled type), not signer count — removed; always `VerifySecure`.
- **BUILD re-pin**: `build_dashbls.sh` now builds dash's own `src/dashbls` subtree at **v23.1.7** (byte-identical to dashd), not standalone bls-signatures 1.3.6 (`dd683653`), which lacks fixes dash applied on top (e.g. a relic `bn_free` leak).
- **KAT gap**: added partial signers/validMembers verify + a `>=threshold-<minSize` ingest-reject case.

## Follow-up (noted, not here)
- qrinfo-based **rotated** (DIP-24) member sourcing (this PR sources the historical SML via `getmnlistd` for the non-rotated path — the primitive rotated needs qrinfo for the quarter-rotation snapshots);
- cache-poisoning DoS (unverified-first-wins admission / keep-best dedup can force permanent null-serve — availability-only, consensus-safe) + unbounded quorumHash map;
- CI must set `C2POOL_DASH_BLS=ON` before the enable gate.


frstrtr:
## Adversarial re-review — head b0f485e1 (member-set sourcing, stacked on #812)

**Verdict: CHANGES-REQUIRED.** The scoring/selection/modifier pipeline is byte-verified against dashcore v23.1.7 and the KAT is genuinely strong (it exercises the real CL-tuple modifier path and proves verify-ACCEPT end-to-end at one height). But the sourcing layer has two reward-critical defects (R1, R3) and the keystone sources the SML at the **wrong block** (R2). All folded #812 fixes check out except the re-pin execution gap. Findings ranked by reward impact.

---

### R1 — MUST-FIX (reward-critical): duplicate getmnlistd replies leak past the demux and REPLACE the tip SML

`request()` pushes onto `m_await_sml[hash]` / `m_await_cl[hash]` and calls `m_send()` **unconditionally**. On mainnet, non-rotated types share quorum base blocks every cycle: 50_60 (interval 24) and 100_67 (24) share **every** cycle start; 400_60/400_85 align at 288/576 boundaries. Two `request()` calls for the same `quorum_hash` (different llmqType) issue **two identical `getmnlistd(ZERO, hash)` requests**, and the peer replies twice. The first reply consumes and **erases** the await entry; the second reply no longer matches, falls through the filter in `p2p_client.hpp`, and reaches `CoinStateMaintainer::on_mnlistdiff` — which by design accepts **any ZERO-base diff as a full snapshot** ("always safe — it replaces"). Result: the tip SML/quorum/credit-pool state is replaced by a ~10–30-block-old historical snapshot, recurring **every shared cycle**. Until the next tip diff re-converges, templates carry a wrong merkleRootMNList/merkleRootQuorums (lost block if won) or, at best, the credit-pool freshness gate fail-closes the arm (recurring fallback flap). The CL walk-back loop amplifies this: `consume_cl` calls `m_send` **per pending key** for the same `prev_hash`.

Fixes (all three):
1. Send once per block hash: only `m_send` when the await vector for that hash was empty (both legs, including the walk-back loop).
2. `consume_sml`/`consume_cl` must require `diff.baseBlockHash.IsNull()` — they treat the diff as a full snapshot, so a non-ZERO-base diff must never match.
3. Hardening (strongly recommended, pre-existing exposure this PR makes reachable via honest peers): the tip maintainer should reject an unsolicited ZERO-base snapshot whose height is below the current SML height unless a cold resync is actually in flight (`sml_base == ZERO`). Today a malicious peer can push one unsolicited at any time.

### R2 — MUST-FIX (keystone correctness): SML sourced at the quorum BASE block; dashcore v23.1.7 uses the WORK block (base − 8)

v23.1.7 `GetAllQuorumMembers`, non-rotated post-V20 branch (llmq/utils.cpp ~640):

```cpp
const CBlockIndex* pWorkBlockIndex = DeploymentActiveAfter(...V20)
    ? base->GetAncestor(base->nHeight - WORK_DIFF_DEPTH)   // = 8 (llmq/snapshot.h:39)
    : base;
CDeterministicMNList mn_list = dmnman.GetListForBlock(pWorkBlockIndex);
quorumMembers = ComputeQuorumMembers(llmqType, params, mn_list, base);
```

Post-V20 — the **only** era this module serves, per its own kV20Floor — the member list is the DMN list **as of the work block**, not the base block. This PR fetches `getmnlistd(ZERO, quorumHash)` (base-block list). The member set diverges whenever the DIP-3 list mutates inside `(base−8, base]` — registrations, PoSe isValid flips, confirmedHash confirmations, operator-key rotations — routine on mainnet. Direction is safe (wrong set → verify fails → null-serve) but it silently defeats the feature intermittently, and the single-height KAT cannot catch it (its 8-block window evidently had no churn). Fix: source the SML at the **work block** — which collapses both legs into ONE `getmnlistd(ZERO, workHash)` carrying the list AND the cbTx CL, simplifying the module and shrinking R1's surface. The header comment "over the SML AS OF the quorum base block" needs the same correction.

### R3 — MUST-FIX (reward-critical, the one pass-wrong path): the historical SML is accepted unauthenticated

`consume_sml` applies the peer's diff with **no verification**: `cbTxMerkleTree` is never checked and the computed SML root is never compared to `cbTx.merkleRootMNList`. Under the daemonless trust model (dialing untrusted mainnet peers), a single malicious peer can (a) serve a fabricated full snapshot whose entries carry attacker operator keys, and (b) relay a fabricated qfcommit whose membersSig/quorumSig are valid **against that fake member set**. It passes structural ingest, passes BLS verify (the member set is the root of trust, and it's attacker-supplied), and gets served → the network rejects the block (bad-qc) → lost block. This is the only path in the PR where verify can PASS wrongly, and it's exactly what the member set was supposed to prove. Fix (DIP-4 client verification for member-sourcing snapshots): verify the diff's cbTx against the block header's hashMerkleRoot via `cbTxMerkleTree` (the header chain is already verified), then require computed-SML-root == that cbTx's `merkleRootMNList`; fail closed otherwise. This also authenticates the CL leg (bestCLSignature comes from the now-proven cbTx). The tip path shares the unverified posture (pre-existing, separate issue), but real-commitment serving must not ship on an unauthenticated member set.

### R4 — MUST-FIX or explicit carve-out: platform quorums are Evo-only; the filter is missing

`ComputeQuorumMembers` (utils.cpp:290): `EvoOnly = (llmqTypePlatform == llmqType) && V19` → `CalculateScoresForQuorum` skips `dmn->nType != MnType::Evo`. Mainnet `llmqTypePlatform = LLMQ_100_67` (chainparams.cpp:276) — which **is** in this PR's enabled mainnet set; testnet is LLMQ_25_67. `compute_quorum_members` scores all valid+confirmed MNs, so every 100_67 member set is wrong → verify always fails → permanent silent null-serve for type 4, plus a wasted ~100-key VerifySecure per template per window. The SML entry carries `nType` (nVersion ≥ 2), so the filter is implementable; minimum acceptable is an explicit `nullopt` for the network's platform type with a comment (clean fail-closed, no per-template verify burn).

### R5 — should-fix: the null-CL branch diverges from upstream (and the walk-back is not what dashcore does)

v23.1.7 `GetNonNullCoinbaseChainlock` (evo/cbtx.cpp:219–247) does **not** walk back: it reads only the work block's own cbTx and returns nullopt if `bestCLSignature.IsValid()` fails, whereupon `GetHashModifier` uses the fallback `SerializeHash(make_pair(type, workBlockHash))`. (The "nearest non-null" semantics live in the cbTx **field** itself — consensus makes bestCLSignature effectively monotone — not in a block-index walk.) This PR's ≤32-block walk-back would, on a genuinely null-CL work block, build a modifier no dashd computes (an older block's CL with the work height), and the fallback branch of `compute_quorum_modifier` is unreachable dead code (`try_finalize` requires `clsig`). Fail-safe direction, but it is precisely the "right at one height, wrong on the CL-fallback branch" divergence: drop the walk-back and, on a null/absent/unparseable CL at the work block, finalize with the fallback modifier — byte-matching upstream in all branches. The quorum_members.hpp header comment describing GetNonNullCoinbaseChainlock as walking back should be corrected too.

### Verified correct (byte-checked against v23.1.7 sources)

- **Modifier (Q1)**: `SerializeHash(make_tuple(LLMQType:uint8, workHeight:int32 LE, CBLSSignature:96B))`, workHeight = base − WORK_DIFF_DEPTH(8), SHA256d — matches utils.cpp:88–111 exactly; fallback preimage `(type, workBlockHash)` correct in form (see R5 for when it must apply). The KAT **does** exercise the CL-tuple path (real 96-byte testnet CL); the fallback branch has no dashd vector (only a "differs" test) and is currently unreachable — acceptable once R5 makes it upstream-exact.
- **Scores/selection (Q2)**: single-SHA256(single-SHA256(proTxHash‖confirmedHash)‖modifier) with proTx first (dmnstate.h UpdateConfirmedHash); skip PoSe-banned (isValid) + null-confirmedHash; descending arith_uint256 (LE-numeric) sort; top `size` — all match. Tie-break nit: upstream ties resolve by collateralOutpoint, not proRegTxHash — unreachable (needs a SHA256 collision), but the comment claims parity; fix comment or match.
- **Scheme (Q4)**: per-entry `legacy = (nVersion == VER_LEGACY_BLS)` is right; the KAT mixes nVersion 1/2 entries and asserts per-index flags; a wrong flag mis-deserializes and can only fail (no pass-wrong).
- **Folded #812 fixes (Q5)**: (a) quorumIndex — upstream commitment.cpp:120 requires `base_height % dkgInterval == quorumIndex` unconditionally; the serve-time guard is the right formula in the right place, refusing to null (and for non-rotated wire nVersion 3 the field isn't even serialized → always 0 → equivalent). (b) minSize 40/50/300/350/80/22 byte-match llmq/params.h; ingest enforces both CountValidMembers and CountSigners ≥ minSize. (c) dead single-signer branch removed; always-VerifySecure matches upstream for every enabled type. (d) re-pin target (dash v23.1.7 `src/dashbls` subtree, tag 733786d5) is the correct source and the script is correct in form — but it has NOT been executed, so the passing KATs attest the **old** standalone-pin lib. Rebuild from the new pin + full KAT rerun is a merge/tag gate. Nit: the script comment claims a sparse-checkout fallback that doesn't exist.
- **Fail-closed audit (Q6)**: rotated → nullopt at both request() and compute (double-guarded, cannot serve rotated); pre-V20, header gaps, short SML, zero operator key, ambiguous compute, walk-back exhaustion, absent verifier — all fail closed. The audit fails at R1 (leak corrupts a *different* subsystem), R3 (fail-open to a lying peer), R4 (wasted verify instead of clean refusal). Also: `m_pending`/awaits never expire (dead requests accumulate; add a height- or time-based reap), `request()` lacks upstream's `base % dkgInterval == 0` refusal (utils.cpp:294) which would also bound peer-driven request amplification, and `awaiting()` is now unused.
- **End-to-end (Q7)**: verified_for → quorumIndex guard → build_qc_tx → compute_merkle_root_quorums_with_block folds non-null commitments with replacement/eviction. The KAT proves verify-ACCEPT; it does **not** prove merkleRootQuorums parity at a DKG-window height with a real commitment folded — that is exactly the soak gate below.

### Required mainnet-representative soak scope (gate for enabling real-serve)

1. Multi-cycle soak (testnet first, then mainnet read-only) spanning windows of types 1/2/3/4: at every window height where a real commitment would be served, assert (a) commitment byte-equal to the one actually mined (getblock), (b) would-be merkleRootQuorums == dashd's for the same height+set, (c) member-set READY latency < window start (else the feature is silently inert).
2. Churn coverage: cycles where the DIP-3 list mutates inside (base−8, base] — member sets must match dashd's `quorum info` 100% post-R2; pre-R2 this is where mismatches will appear.
3. A 576-boundary cycle (all four non-rotated types committing on shared bases): zero tip-SML regressions (watch base-continuity REJECT logs, sml_current_hash monotone) and zero unexplained fallback flaps (R1).
4. Adversarial peer leg (R3): a peer serving fabricated historical snapshots must never yield a READY set post-fix.
5. Restart mid-sourcing + peer disconnect: no stuck pendings, no leak-through.
6. dashbls rebuilt from the v23.1.7 subtree pin, full KAT suite green on the new lib, before tag.
7. Thread audit (TSAN pass or strand proof) that lookup() on the template path and consume on the coin ioc thread are actually the same strand.

Direction summary: every defect except R1 and R3 fails safe (null-serve). R1 corrupts the tip SML off honest traffic on a recurring schedule; R3 is the single pass-wrong path. Both must land before this branch can serve a real commitment anywhere near mainnet.

frstrtr:
Superseded — this stack was squash-merged into master via #826 (Phase-L daemonless quorum-commitment serving). Content is on master; closing to avoid a duplicate merge. Soak-validated (#816 reward-safety soak: 130 serves / 0 bad-cbtx).

PR #818, opened 2026-07-23
dash(embedded): superblock governance-vote BLS verify — enable daemonless superblock serving (R3)

frstrtr:
## R3 — governance-vote BLS operator-key verification (enables daemonless superblock serving)

**DO-NOT-MERGE draft.** Base: `dash/embedded-superblock-govsync` (#810). Depends on #810 (govsync superblock sourcing) + the dashbls BLS lib from the #814/#812 stack (brought in here).

### Problem
#810's daemonless superblock arm is **fail-closed-inert**: its vote-verifier seam (`CoinStateMaintainer::set_vote_verifier`) is UNSET, so no funding vote is counted → no trigger reaches threshold → every superblock height falls back to dashd. R3 makes that verifier real.

### What R3 adds
- **`verify_govvote_operator_sig`** (`vendor/bls_verify.cpp`): single-signature verify of a governance vote's `vch_sig` against the voting MN's **operator** BLS key over `govvote_signature_hash()` — dashcore `CGovernanceVote::CheckSignature(const CBLSPublicKey&)`. TRIGGER funding votes are operator-key-signed (NOT the ECDSA voting key — that's proposal-funding-only, the #810 R3 finding). Sits alongside the aggregate quorum-commitment verify already in this TU. The key's wire-encoding (`LEGACY_BLS` registration) and the signing scheme (post-V19 `BASIC`) are **independent** and varied separately — pinned against a real from-wire vote.
- **Wiring** (`main_dash.cpp`, `--embedded-superblock`): `set_vote_verifier` resolves the voting MN by **collateral outpoint** in the live DMN view (`mnstates().find_by_collateral` → `pubKeyOperator` + `nVersion` scheme + `isValid`), enforces the `IsValid` time bound (`nTime <= now+1h`), then BLS-verifies. Companion `set_vote_weight_fn` (EvoNode 4x, Regular 1x, 0 when not in the valid set) wired from the same DMN view.
- **BLS lib**: root `CMakeLists` `C2POOL_DASH_BLS` option + `scripts/build_dashbls.sh` + `dash_bls_verify` static lib, from the #814 stack (Dash Core v23.1.7 `src/dashbls` subtree — the same lib dashd wraps). Default **OFF** → fail-closed stub, no dashbls dependency in any build (Windows `main_ltc` clean).

### Fail-closed chain HOLDS
Unverified votes never tally; a trigger wins only past the R4 weighted threshold (`max(minQuorum, weighted/10)`); **R5 completeness (`set_superblock_sync_complete_fn`) + the R6 desync latch remain UNWIRED**, so R3 alone can never open the serve path. Landing R3 keeps every superblock height fail-closed to dashd until R5 + a funded soak.

### KATs (`test_dash_govvote_bls`, links `dash_bls_verify`)
- **REAL from-wire** (testnet dashd @192.168.86.52, read-only): a DELETE-signal operator-key vote (96-byte BLS `vchSig`, object `4fe428f7…`, MN collateral `4ee3ff50…-95`, proTxHash `bc77a5a2…`, operator key `174de566…`) **verifies** against the real operator key. This simultaneously pins the `govvote_signature_hash` preimage, the BLS verify, and the scheme handling.
- **Tamper / wrong-key / wrong-digest / wrong-size REJECT** — incl. a 65-byte ECDSA proposal-FUNDING vote (real from-wire contrast: proposal-funding = ECDSA voting-key, delete/trigger = BLS operator-key).
- **Synthetic end-to-end**: no funded superblock trigger exists on testnet, so the threshold → winner → payee ladder + the maintainer's verified-only tally gate are pinned synthetically on the real `GovernanceStore`/maintainer code. Real-vector vs synthetic documented in the test header.

### Build / verify
`cmake … -DC2POOL_DASH_BLS=ON -DDASHBLS_ROOT=/home/ubuntu/dashbls`; `c2pool-dash` links dashbls; `test_dash_govvote_bls` 7/7 PASS, `test_dash_superblock` 24/24 PASS (no regression).

---

### Review must-fixes landed (015c9e72) — governance-tally dashcore-v23.1.7 parity
All three CHANGES-REQUIRED items were one hazard class: c2pool tallying votes dashd does NOT (or dropping ones dashd counts) → tally divergence → wrong trigger wins → wrong superblock payees → lost block. The crypto core is unchanged.

1. **Signature axis hard-pinned BASIC** (`vendor/bls_verify.cpp`): dashcore `CheckSignature` does `SetBytes(vchSig, false)` + `VerifyInsecure(…, false)` — no legacy-sig retry (that fallback exists only in stream `Unserialize`). The legacy-sig verify fallback (a superset of dashcore — tally-inflation vector via a hostile MN legacy-signing a YES that dashd rejects and never relays) is removed; the **pubkey-ENCODING** fallback is kept (mirrors `CBLSLazyPublicKey` ingest of `LEGACY_BLS`-registered keys). Pre-V19 note documented in-code: this tally only runs post-V19; if that ever changes, gate on the V19 fork height.
2. **PoSe-ban (`isValid`) filters removed** from BOTH the verify and the weight closures: dashcore resolves the voter with the **unfiltered** `GetMNByCollateral` in both `CGovernanceVote::IsValid` and `CGovernanceObject::CountMatchingVotes` — a banned MN's vote verifies and counts at full weight (dropping its NO inflates yes−no under natural ban churn). **Preserved asymmetry** (dashcore's own): the threshold denominator (`reseed_funding_threshold` = `max(minQuorum, valid_weighted/10)`, dashcore `m_valid_weighted`) stays VALID-only; only the per-vote tally is unfiltered. Both closures now share `gov_mn_by_collateral` / `gov_vote_weight_for_key` (`coin_state_maintainer.hpp`).
3. **Replacement semantics matched exactly** (`governance_store.hpp` + `on_govvote`): outcome **NONE is stored** (a newer NONE replaces a stored YES → yes-count drops); replacement per `ProcessVote`: reject `nTime <` stored ("Obsolete vote"); on EQUAL `nTime` reject ONLY new outcome `<` stored (upstream's explicit tie-break); otherwise replace.

**KATs** (fail-before/pass-after verified against the pre-fix tree: 3 FAILED before → **11/11 PASS** after; the 5 real-from-wire crypto KATs pass unchanged on both): legacy-scheme-signed vote rejects under every encoding combination while the same key basic-signed accepts under both pubkey encodings (fallback preserved); banned-MN vote resolves + counts at full weight while the threshold reseeds valid-only (asymmetry asserted both directions); NONE-replaces-YES + obsolete-reject + exact equal-nTime tie-break, at both the store and the `on_govvote` ingest legs. `test_dash_superblock` 24/24 PASS, `c2pool-dash` builds.

**Review nits**: the `IsValid` time bound uses the **local clock** where dashcore uses `GetAdjustedTime()` — documented at the bound (no peer time-offset plumbing here; NTP-disciplined hosts make the divergence negligible; revisit if soak shows drift). **No orphan pool for early votes**: a vote arriving before its trigger object is dropped, where dashd would post-process it once the object lands — an AVAILABILITY-only risk (we under-count vs dashd → tally short of threshold → fail closed to dashd; never the unsafe direction). To be confirmed during the funded-superblock soak — if govsync's object-before-votes ordering doesn't hold in practice, an orphan buffer becomes a follow-up.

Fail-closed chain unchanged: **R5 (`set_superblock_sync_complete_fn`) remains unwired** — no production caller sets it, `resolve_superblock` still refuses the serve path — so R3 + these fixes still cannot open embedded superblock serving by themselves.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -194,6 +194,58 @@ if(GTest_FOUND OR GTEST_FOUND)
 else()
     message(STATUS "GTest not found - testing disabled")
 endif()
+# ── E1 Phase-L: optional dashbls (BLS12-381) backend for type-6 quorum-
+# commitment verification ────────────────────────────────────────────────────
+# dashpay/bls-signatures ("dashbls", relic-backed, Apache-2.0) — the SAME
+# library dashd wraps in src/bls/bls.h. OPT-IN and DASH-ONLY: only the DASH impl
+# (dash_bls_verify) + its KAT consume it, never the shared core the Windows
+# main_ltc launcher builds ([[reference_windows_ci_conan_flake]]: Windows CI
+# builds only the LTC launcher, so DASH-only BLS code cannot break it). When the
+# option is OFF (default) or the library is not found, dash_bls_verify compiles a
+# fail-closed stub and the DKG-window serve path keeps the pre-Phase-L
+# null-commitment posture exactly — no dashbls dependency enters any build.
+#
+# Build the library once for the CI factory with scripts/build_dashbls.sh (builds
+# dash's OWN src/dashbls subtree at the v23.1.7 tag — byte-identical to dashd);
+# point cmake at it with -DDASHBLS_ROOT=<prefix>.
+option(C2POOL_DASH_BLS "Link dashbls (BLS12-381) for E1 Phase-L real quorum-commitment verify" OFF)
+set(C2POOL_DASH_BLS_AVAILABLE OFF)
+if(C2POOL_DASH_BLS)
+    # Honour a DASHBLS_ROOT hint (the build-factory install prefix), else search
+    # the system paths. dashbls installs: include/dashbls/*.hpp,
+    # lib/libdashbls.a, lib/librelic_s.a, lib/mimalloc-2.0/libmimalloc-secure.a.
+    find_path(DASHBLS_INCLUDE_DIR dashbls/bls.hpp
+        HINTS ${DASHBLS_ROOT} ENV DASHBLS_ROOT PATH_SUFFIXES include)
+    find_library(DASHBLS_LIB       NAMES dashbls bls-dash
+        HINTS ${DASHBLS_ROOT} ENV DASHBLS_ROOT PATH_SUFFIXES lib)
+    find_library(DASHBLS_RELIC_LIB NAMES relic_s relic
+        HINTS ${DASHBLS_ROOT} ENV DASHBLS_ROOT PATH_SUFFIXES lib)
+    find_library(DASHBLS_MIMALLOC_LIB NAMES mimalloc-secure mimalloc
+        HINTS ${DASHBLS_ROOT} ENV DASHBLS_ROOT PATH_SUFFIXES lib lib/mimalloc-2.0 lib/mimalloc-2.1)
+    find_library(DASHBLS_GMP_LIB NAMES gmp)
+    # NOTE: the autotools install (scripts/build_dashbls.sh, the only build
+    # path dash's v23.1.7 src/dashbls subtree supports) folds relic + mimalloc
+    # INTO libdashbls.a — so the separate relic/mimalloc libs are OPTIONAL and
+    # only linked when present (the legacy standalone-CMake layout).
+    if(DASHBLS_INCLUDE_DIR AND DASHBLS_LIB AND DASHBLS_GMP_LIB)
+        set(C2POOL_DASH_BLS_AVAILABLE ON)
+        set(DASHBLS_LIBRARIES ${DASHBLS_LIB})
+        if(DASHBLS_RELIC_LIB)
+            list(APPEND DASHBLS_LIBRARIES ${DASHBLS_RELIC_LIB})
+        endif()
+        if(DASHBLS_MIMALLOC_LIB)
+            list(APPEND DASHBLS_LIBRARIES ${DASHBLS_MIMALLOC_LIB})
+        endif()
+        list(APPEND DASHBLS_LIBRARIES ${DASHBLS_GMP_LIB})
+        message(STATUS "dashbls found (E1 Phase-L BLS verify ENABLED): ${DASHBLS_LIB}")
+        message(STATUS "dashbls includes: ${DASHBLS_INCLUDE_DIR}")
+    else()
+        message(WARNING "C2POOL_DASH_BLS=ON but dashbls not found "
+            "(need dashbls + relic + gmp; run scripts/build_dashbls.sh and pass "
+            "-DDASHBLS_ROOT=<prefix>). Falling back to fail-closed null-serve.")
+    endif()
+endif()
+
 include_directories(include)
 include_directories(src)
 
--- a/src/c2pool/CMakeLists.txt
+++ b/src/c2pool/CMakeLists.txt
@@ -327,6 +327,8 @@ target_link_libraries(c2pool-dash
                        # CBlockHeaderAndShortTxIDs::FillShortTxIDSelector()
     dash_bls_verify    # E1 Phase-L: real type-6 quorum-commitment BLS verify
                        # (make_commitment_bls_verifier -> set_bls_verify_fn seam).
+                       # R3: also verify_govvote_operator_sig (governance-vote BLS
+                       # operator-key verify, daemonless superblock serving).
                        # Carries the C2POOL_DASH_BLS define + dashbls libs PUBLIC
                        # when the backend was found; fail-closed stub otherwise.
     btclibs
--- a/src/c2pool/main_dash.cpp
+++ b/src/c2pool/main_dash.cpp
@@ -80,6 +80,7 @@
 #include <impl/dash/coin/mn_list_ingest.hpp>     // wire_mn_list_ingest (leg 4)
 #include <impl/dash/coin/govsync_ingest.hpp>     // wire_govobject_ingest / wire_govvote_ingest (E-SUPERBLOCK)
 #include <impl/dash/coin/superblock.hpp>         // get_superblock_payments / superblock_budget (E-SUPERBLOCK)
+#include <impl/dash/coin/vendor/bls_verify.hpp>  // R3: verify_govvote_operator_sig — governance-vote BLS operator-key verify
 #include <impl/dash/coin/mn_seed.hpp>            // E2c: RPC protx-list MN-set seed (parse_protx_list_seed)
 #include <impl/dash/coin/mn_checkpoint.hpp>      // E2d: pinned MN-set checkpoint format + fail-closed parser
 #include <impl/dash/coin/mn_checkpoint_lane.hpp> // E2d: checkpoint -> forward-replay bridge -> leg-4 publish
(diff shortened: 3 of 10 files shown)


frstrtr:
## Adversarial review — R3 governance-vote BLS verify (correctness + reward-safety)

Reviewed head `0d55ea15` vs base `dash/embedded-superblock-govsync`, checked line-by-line against dashpay/dash **v23.1.7** (+ master) `governance/vote.{h,cpp}`, `governance/object.cpp`, `evo/deterministicmns.cpp`, `bls/bls.{h,cpp}`.

**Verdict: CHANGES-REQUIRED** — the two highest-risk items (digest preimage, encoding/scheme decoupling) check out, and the fail-closed chain holds (R3 cannot open the serve path). But three dashcore-parity deviations all err in the same DANGEROUS direction — c2pool tallying votes (or vote-effects) the network does not — and each is a wrong-trigger/wrong-payee → rejected-block hazard once R5 lands. All are small, surgical fixes. None is reachable while R5 is unwired, so they gate serve-enable, not this draft's structure.

### ★ Digest preimage — VERIFIED byte-exact, and the KAT is genuine proof
`govvote_signature_hash` (outpoint hash 32 raw + index u32LE, parentHash 32, **outcome i32LE before signal i32LE**, time i64LE) matches `CGovernanceVote::GetSignatureHash() == SerializeHash(*this)` exactly: the `SERIALIZE_METHODS` guard `if (!(s.GetType() & SER_GETHASH)) READWRITE(obj.vchSig);` excludes the sig, and field order/types match (vote.h, v23.1.7). The identity hash (signal-before-outcome + legacy dummies) is correctly kept separate.

The real-vote KAT is **not** tautological: `vchSig` and `pubKeyOperator` are external from-wire/RPC constants; only the digest is computed by the code under test. A BLS verify passing over a wrong digest is equivalent to forging a BLS signature — so the pass simultaneously pins preimage layout, key resolution, and scheme handling. Caveat (documented in the test, fine): the vector is a DELETE-signal proposal vote, not a TRIGGER FUNDING vote — same signing path, since `onlyVotingKeyAllowed = (type==PROPOSAL && signal==FUNDING)` (object.cpp) is the ONLY ECDSA carve-out; everything else is operator-key BLS.

### ★ Decoupled encoding/scheme — model CONFIRMED, but the sig axis is a superset of dashcore (MUST-FIX 1)
The decoupling claim is exactly how dashcore works: `CheckSignature(const CBLSPublicKey&)` receives `dmn->pdmnState->pubKeyOperator.Get()` — deserialized at DMN-ingest time under the key's **registration** scheme (nVersion-driven, `CBLSLazyPublicKey`) — and then does `sig.SetBytes(vchSig, false); sig.VerifyInsecure(pubKey, GetSignatureHash(), false)` (vote.cpp v23.1.7): sig-encoding **and** DST hard-pinned **BASIC**, with literals, no `bls_legacy_scheme.load()`, no fallback (`SetBytes` has no CheckMalleable retry — that retry exists only in stream `Unserialize`).

Deviation: the `for (bool net_legacy : {false, true})` loop in `verify_govvote_operator_sig` additionally accepts a **legacy-scheme-signed** vote that dashd unconditionally rejects post-V19. That is not a forgery hole (only the operator key-holder can produce it under either DST) — but it IS a **network-view divergence in the tally-inflating direction**: a hostile *registered* MN can sign a legacy-scheme YES vote on a near-threshold trigger and feed it to c2pool nodes (dashd peers reject → won't relay → the honest network never counts it). c2pool then deems the trigger funded when the network does not → serves superblock payees on a height validators treat as untriggered → block rejected → lost block. Fix: drop the legacy iteration, pin the sig axis to basic exactly as dashcore does. Pre-V19 chains don't need the legacy arm (V19 is buried on both networks this targets).

The **pubkey**-axis fallback (declared encoding first, other on failure) is fine to keep — dashcore's own `CBLSWrapper::Unserialize` does an opposite-scheme retry at ingest, and the pubkey bytes come from our own DMN view, not the attacker.

`VerifyInsecure` is the right primitive: single sig, single fixed key from the DMN registration — no aggregation anywhere in this path, so rogue-key is moot. Confirmed it's a plain core-scheme verify (`Scheme(fLegacy)->Verify(pk, hash, sig)`), no message augmentation — matching dashcore.

### MN resolution — collateral outpoint is CORRECT; the `isValid` filter is NOT dashcore (MUST-FIX 2)
Modern `CGovernanceVote` carries `COutPoint masternodeOutpoint` (vote.h) — no proTxHash — and dashcore resolves it via `tip_mn_list.GetMNByCollateral(...)`. Resolving by collateral in the live DMN view is exactly right, and unknown-MN / unresolvable-key / wrong-size all fail closed. ✔

But dashcore uses the **unfiltered** lookup in BOTH `CGovernanceVote::IsValid` and `CountMatchingVotes` — `GetMNByCollateral` has no `IsBanned()` filter (deterministicmns.cpp; contrast `GetValidMNByCollateral`, which governance does NOT use). **A PoSe-banned MN's governance votes verify and count at full weight in dashd.** The wiring here rejects banned MNs at verify AND weighs them 0 at tally. Dropping a banned MN's YES only under-counts (safe fallback); dropping a banned MN's **NO** vote *inflates* `yes − no` relative to the network — no attacker needed, natural PoSe-ban churn near a threshold produces the divergence. Fix: remove the `isValid` check from both closures in main_dash.cpp (verifier + weight fn). Keep the threshold reseed exactly as is — dashcore's `nAbsVoteReq` genuinely uses `m_valid_weighted` (valid-only), so the tally-counts-banned / threshold-excludes-banned asymmetry is dashcore's own and must be reproduced, not "fixed".

### Vote replacement / outcome-range parity (MUST-FIX 3, same family, smaller)
- `on_govvote` drops outcome NONE; dashcore's `ProcessVote` stores it (IsValid passes `outcome >= VOTE_OUTCOME_NONE`), so an MN's newer NONE vote *replaces* its earlier YES in dashd but leaves the YES standing in c2pool → inflation, same direction as above.
- Equal-timestamp tie-break: dashcore rejects the new vote only when `new outcome < stored outcome`; `add_verified_funding_vote` keeps the stored vote on `timestamp ==` unconditionally.
Both are MN-self-signed edge cases, but they're two-line alignments — take them with fixes 1–2.

### Weights, tally, threshold, winner — parity confirmed
EvoNode 4× / Regular 1× matches `GetMnType(nType).voting_weight`; membership-at-tally via the live view matches `CountMatchingVotes` (modulo the banned-MN point); `yes − no` vs `max(minQuorum, weighted/10)` matches `GetAbsoluteYesCount` + `UpdateSentinelVariables`; highest-yes with lowest-hash tie-break matches `GetBestSuperblock`'s ascending-map iteration with strict `>`. The `"<txid>-<index>"` key round-trips (`to_key` GetHex ↔ `SetHex`). ✔

### nTime bound — correct
`nTime <= now + 3600` is exactly `IsValid`'s only time check; dashcore has **no lower bound** there (stale votes die via newest-wins replacement), so none is missing. Nit: dashcore uses `GetAdjustedTime()` (network-adjusted) vs `std::time` here — acceptable, worth a comment. Second nit: no `GOVERNANCE_UPDATE_MIN` (60-min) replacement rate-limit — low exposure since rate-limited votes aren't relayed by dashd peers, and rate checks are off during sync; document rather than implement.

### Fail-closed audit — HOLDS, R3 cannot open the serve path
Verified in `NodeCoinState::resolve_superblock`: a superblock height serves only if (flag ON) ∧ (provider wired) ∧ (**R5 predicate PRESENT and true — not wired anywhere in this PR**) ∧ (provider returns a value), else refuse → dashd fallback; the R6 latch is folded inside `superblock_schedule` upstream of that. Independently: BLS-absent build → stub `false` → zero votes tally; verifier unset → zero; threshold 0 → no winner; unknown-MN weight 0. Every layer confirmed in code, and tests 6–7 pin the tally gates. Enabling the verifier cannot accidentally serve.

### Other notes
- Vote-before-object ordering: a govvote arriving before its trigger is dropped with no orphan pool (dashd keeps orphan votes). Fail-safe direction (under-count → fallback), but it's a **permanent-fallback availability risk** if govsync ordering doesn't guarantee objects-then-votes — worth confirming in the soak.
- The verifier binds the digest correctly: `vote_hash` is recomputed in p2p_client from the wire fields, never attacker-supplied independently of them. ✔

### Required funded-superblock soak scope (gates serve-enable, with R5)
No funded trigger exists on testnet, so before `--embedded-superblock` can serve:
1. **Mainnet read-only shadow, ≥2 superblock cycles**: continuous diff of c2pool `absolute_yes_count`/winner vs dashd `gobject list`/`getcurrentvotes` for every live trigger (tally parity is the direct probe for all three must-fixes — banned-MN votes and scheme handling included), plus payee-vector byte-parity of `superblock_schedule(H)` vs the actual superblock coinbase at H.
2. **Threshold-crossing timing**: confirm c2pool flips triggered no earlier than dashd's `fCachedFunding` (early-flip = the dangerous direction).
3. **Cold-restart mid-cycle**: re-sync rebuilds the tally to dashd's counts (probes the vote-before-object ordering risk).
4. **R6 cross-check fires zero times** across both cycles.
5. If a controlled net is preferred: devnet with owned MNs (incl. one PoSe-banned voter and one EvoNode for the 4× weight) producing a genuinely funded trigger end-to-end.

With fixes 1–3 applied this is APPROVE for landing as the fail-closed draft it is; the soak above remains the gate for ever wiring R5.

frstrtr:
Un-drafted. Was left draft by habit — no do-not-merge intent. Reds are all infra: Analyze (c-cpp/python) is the CodeQL uplink timeout; the failing standalone `dash smoke` leg hung ~6h and timed out on a runner while its canonical `coin-matrix / dash smoke` twin passed in 5m53s the same run. Every Dash gate that actually executed (G1 byte-parity, G3a, coin-matrix dash smoke) is green.

PR #825, opened 2026-07-23
dash(embedded): rotated (DIP-24) member-sourcing seam + real 60_75 ground truth

frstrtr:
## ITEM 2 (partial): rotated (DIP-24) quorum member sourcing

**Draft — do not merge.** Stacks on `dash/e1-phasel-height-completeness` (#816). This lands the durable, testable, **fail-closed groundwork**; the full rotated compute remains a documented follow-up (see below).

### Why partial
The full rotated path — `getqrinfo` P2P wire + `CQuorumRotationInfo`/`CQuorumSnapshot` decode + `ComputeQuorumMembersByQuarterRotation` port — is a reward-critical, multi-part port. A **wrong** rotated member order serves a bad commitment (bad-cb-payee). No dashcore `llmq/utils.cpp` source was available locally to port from faithfully, and the compute cannot be validated end-to-end without the qrinfo decode. Shipping an unvalidated consensus algorithm would be worse than a clean fail-closed seam + captured ground truth. So this slice makes the follow-up **pinnable against real data** and keeps the reward-safe fallback intact.

### What this lands
1. **Real ground truth** (`test/data/dash_rotated_quorum_members_kat.hpp`): dashd's authoritative `llmq_60_75` member set at testnet height **1520064** — 60 ordered members + operator BLS keys + quorumHash + quorumPublicKey — captured read-only from `quorum info 5 <hash>` on testnet dashd @192.168.86.52. This is the exact order the quarter-rotation port must reproduce (rotated analogue of `dash_quorum_members_kat.hpp`).
2. **Structured fail-closed seam** in `QuorumMemberSource::request()` replacing the terse rotated no-op, naming the exact remaining work:
   - `getqrinfo` request (`CGetQuorumRotationInfo`)
   - `CQuorumRotationInfo` decode (3 `CQuorumSnapshot` skip-lists + cycle-base mnlistdiffs, each DIP-4 authenticated like `authenticate_snapshot`)
   - `compute_quorum_members_by_quarter_rotation` port
   - feed the result to the same `m_ready` map so #812 `verify_final_commitment` serves rotated real
   Until then: no ready set → the #816 completeness gate leaves the mixed-quorum window unserveable → reward-safe dashd fallback (unchanged).

### KAT (`test_dash_quorum_member_source`, built `-DC2POOL_DASH_BLS=ON`)
- `RotatedRealVectorGroundTruthAndFailClosed`: pins the captured 60_75 vector shape and proves the rotated path fails closed at **every** layer — vendor `compute_quorum_members` returns nullopt for `use_rotation`; the sourcing plumbing emits no wire traffic and never yields a ready set.
- `RequestGuardsAndPendingReap` strengthened to assert `lookup()` stays nullopt for the rotated quorum.
All 7 `DashQuorumMemberSource` tests pass.

### Remaining work to make rotated SERVEABLE
The four numbered items above, each gated on reproducing the captured 60_75 order (member order + operator keys) before enabling — same discipline as the non-rotated authenticate-snapshot path.


frstrtr:
Un-drafted + `[DRAFT/do-not-merge]` stripped. All reds are known infra, not code: the Analyze (c-cpp/python) pair is the CodeQL uplink-timeout reddening the whole board; the DGB Phase-B smoke died downloading actions/cache (100s HttpClient timeout, 3 attempts) and never checked out or linked any code — it never reached this Dash-only change. Base is master, MERGEABLE. Every Dash gate + dash smoke is green.

PR #833, opened 2026-07-24
core(web): forward stale/race won-blocks to the daemon instead of dropping them

frstrtr:
## Problem

The shared-core HTTP `submitblock` handler (`MiningInterface::submitblock`, `src/core/web_server.cpp`) compared the submitted header's prev-hash against the cached template and, on a mismatch, logged `stale block`, fired the 253 (orphan) telemetry, and **early-returned a synthetic `stale block: previous block hash mismatch` error without ever calling `submit_block_hex()`**. The block never reached the daemon and a winnable race block was forfeited.

This is the **primary won-block submit path for LTC and NMC**, and is shared by DASH / BTC / DGB / BCH (it is `src/core`, so one change fixes all coins).

## Why it's a bug

A prev-hash mismatch is a **RACE, not a guaranteed loss**:
- chain-extended = our block is a valid same-height competitor (daemon accepts, we can win the race);
- 1-block reorg = our block leads by one (network reorgs to us → we win);
- only when buried 2+ deep is it unwinnable — and forwarding even that is free (the daemon stores a dead orphan).

Guiding invariant: **never refuse a block the daemon might accept.** A false local refusal is a guaranteed irreversible loss; a truly-dead block is rejected by the daemon for free.

## Fix

- Make `stale` **advisory**: keep the log + the 253 race-telemetry (fired once, up front), then **fall through** to the existing `submit_block_hex(hex, false)` path and let the daemon be the authority on validity at the block's real height.
- Return the **daemon's actual verdict** to the caller (accept ⇒ success, reject ⇒ the reject) instead of a synthetic stale error.
- Guard the accepted / rejected / embedded found-block callbacks so a stale block is **never double-recorded**. The reject branch does **not** fire a found-block callback — a `false` return from `submit_block_hex` can also mean daemon-**duplicate** (block already accepted, e.g. via P2P), which must not be mis-counted as DOA.
- The **normal matching-prev-hash accepted path is unchanged** (byte-identical): valid block → `accepted == true` → fire(0) + relay → success.

No consensus / coinbase / serialization change — purely the submit/no-submit **decision** in the RPC handler.

## Reward safety

Forwarding a stale/race block is safe: the daemon validates it. An invalid block is rejected locally by the daemon and never propagates beyond it. The only case that changes behavior for a *valid* block is winning a race that was previously dropped.

## Cross-coin

Because the change is in `src/core`, it fixes DASH, BTC, DGB, BCH, LTC and NMC at once. No per-coin copies of this HTTP-handler stale-drop exist (the `stale block: previous block hash mismatch` string was unique to `web_server.cpp`; the per-coin `template_builder`/`rpc` files reference `previousblockhash` only for template *building*, not submit-drops).

## Tests

New `src/core/test/web_server_submitblock_test.cpp` KATs (mock `ICoinNode`):
- prev-mismatch (race) block **is forwarded** to the daemon and the daemon's verdict is returned — accept ⇒ success, reject ⇒ error (and the error is **not** the old synthetic stale drop);
- normal matching-prev block still forwards and succeeds;
- malformed short payload is still rejected locally (never forwarded).

Build + test: `core_test` green (64/64, incl. the 4 new KATs). `c2pool-ltc` and `c2pool-dash` both compile + link (cross-coin core-lib compile confirmed).

### DIFF
--- a/src/core/coin/submitblock_result.hpp
+++ b/src/core/coin/submitblock_result.hpp
@@ -1,0 +1,43 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#pragma once
+
+// ---------------------------------------------------------------------------
+// core::coin::submitblock_result_accepted -- shared classifier for a coin
+// daemon's `submitblock` RPC result under the dual-path won-block contract.
+//
+// submitblock returns JSON null on ACCEPT; a non-null string is the reject
+// reason. A "duplicate" / "inconclusive" / "already-have" result means the
+// block is ALREADY on the network — the OTHER broadcast arm (our own P2P relay,
+// or a peer that won the very race we are submitting into) landed it first.
+// That is SUCCESS (the block reached the network), NOT a failure. The lone
+// exception is a "duplicate-invalid" reason: the block was rejected as invalid,
+// a genuine failure.
+//
+// Historically LTC/BTC/DGB NodeRPC::submit_block_hex classified with a bare
+// `result.is_null()`, so a daemon "duplicate" (the common dual-path / race
+// outcome) surfaced as FALSE and a WON block was reported as a reject through
+// the shared web_server submit seam. This predicate lifts them onto the same
+// contract BCH and DASH already implement. Pure over the JSON so a KAT can pin
+// it without a live RPC client.
+// ---------------------------------------------------------------------------
+
+#include <string>
+
+#include <nlohmann/json.hpp>
+
+namespace core::coin {
+
+inline bool submitblock_result_accepted(const nlohmann::json& result)
+{
+    if (result.is_null()) return true;      // canonical accept
+    if (!result.is_string()) return false;  // unexpected shape — treat as reject
+    std::string code = result.get<std::string>();
+    for (char& c : code)
+        if (c >= 'A' && c <= 'Z') c = static_cast<char>(c - 'A' + 'a');
+    const bool already_have = code == "duplicate"
+                           || code.find("inconclusive") != std::string::npos
+                           || code.find("already") != std::string::npos;
+    return already_have && code.find("invalid") == std::string::npos;
+}
+
+} // namespace core::coin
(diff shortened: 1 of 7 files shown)


frstrtr:
Addressed the CHANGES-REQUIRED review in commit `243d376d`.

**B1 (BLOCKER — won race recorded as ORPHAN/pending):** removed the eager pre-verdict 253 fire. The RPC-armed path now fires **exactly one** found-block callback **after** the daemon verdict: accepted ⇒ `0` (even when stale, so a won race is credited → `broadcast_bestblock` + `schedule_block_verification`), rejected ⇒ `253` for stale / **nothing** for non-stale (a `false` may be daemon-duplicate — no DOA mis-count), RPC throw ⇒ `253` stale / `254` else. The eager fire is kept only where there is no synchronous verdict: the embedded arm (`253`/`0`) and the no-node discard arm (`253`). `record_found_block` dedups by block hash, so verdict-time firing is double-record-safe.

**M2 (daemon duplicate misreported as reject on LTC/BTC/DGB):** added `core::coin::submitblock_result_accepted` (`src/core/coin/submitblock_result.hpp`) and wired it into `NodeRPC::submit_block_hex` for **LTC, BTC, DGB** — null / duplicate / inconclusive / already-have = ACK; `duplicate-invalid` and real reject reasons = reject. Lifts them onto the same contract DASH and BCH already implement. DASH/BCH untouched.

**M3:** `web_server_submitblock_test.cpp` now installs `set_on_block_submitted` and asserts exactly one fire with the correct code per path (stale+accept⇒`0`, stale+reject⇒`253`, normal+accept⇒`0`, normal+reject⇒error+no-fire, RPC-throw stale⇒`253` / non-stale⇒`254`), plus a pure KAT for `submitblock_result_accepted`.

`core_test` green (68/68). `c2pool-ltc`, `c2pool-dash`, `c2pool-btc`, `c2pool-dgb` all compile + link.

PR #872, opened 2026-07-26
fix(p2p): addrme relays when we HAVE peers, not when we have none (#869)

### LINKED ISSUE #869: dash(p2p): addrme handler guard is inverted — random_choice on empty m_peers (UB) and addrme never relayed
## Summary

`Legacy::HANDLER(addrme)` in `src/impl/dash/protocol_legacy.cpp` (~lines 53-72) has an inverted peer guard on **both** branches:

```cpp
if (m_peers.empty() && (core::random::random_float(0, 1) < 0.8))
{
    auto random_peer = core::random::random_choice(m_peers);   // m_peers is EMPTY here
    ...
}
```

Canonical p2pool relays **if we have peers** (`p2p.py`). Ours fires only when `m_peers` is **empty**, then calls `random_choice` on the empty container.

## Two defects

1. **Undefined behaviour.** `random_choice` on an empty container has no valid element to return. Reached whenever an inbound `addrme` arrives while we hold zero peers — uncommon but reachable (startup, post-restart, transient full disconnect).
2. **`addrme` is never relayed in normal operation**, because in normal operation `m_peers` is non-empty and the branch is skipped. Combined with the absence of any origination site, **peers never learn our listen port**, which degrades inbound peer discovery.

`HANDLER(addrs)` in the same file should be checked for the same inversion.

## Fix

Invert to `!m_peers.empty()`, matching canonical, and guard `random_choice` against an empty container regardless. Add an origination path so we advertise our own listen port after handshake.

## Note

Verified during the 2026-07-26 PPLNS session audit: every other p2p message type (version, getaddrs, addrs, shares, sharereq, sharereply, bestblock, have_tx, losing_tx, remember_tx, forget_tx, ping) has a definition, an inbound handler, and a reachable outbound send in the `--run` path. `addrme` is the only genuinely broken one. None of these messages emit any log line at any verbosity, so log-grep counts of zero are not evidence of absence — that misled an earlier investigation.

frstrtr:
Fixes #869.

## The bug

`HANDLER(addrme)` in each coin's `protocol_legacy.cpp` guards **both** relay branches on `m_peers.empty()`. Canonical p2pool ([p2p.py `handle_addrme`](https://github.com/p2pool/p2pool/blob/master/p2pool/p2p.py)) relays when the node **has** peers:

```python
if random.random() < .8 and self.node.peers:
    random.choice(self.node.peers.values()).send_addrme(port=port)
```

Ours is inverted, so:

1. **Functional bug (total):** addrme is relayed on *no* coin, ever.
2. **Latent UB:** when the guard does pass, `m_peers` is by construction empty, and `core::random::random_choice(m_peers)` — `std::map<uint64_t, peer_ptr>` — does `random_int(0,0) == 0`, `std::advance(begin, 0) == end()`, then `item->second`. Past-the-end dereference.

`protocol_actual.cpp` already carried the correct `!m_peers.empty()` in all five coins, so this aligns Legacy with its Actual sibling.

## Verified sites

| coin | legacy (fixed) | actual (already correct, reference) |
|---|---|---|
| dash | `protocol_legacy.cpp:57,67` | `protocol_actual.cpp:59,69` |
| ltc  | `protocol_legacy.cpp:57,67` | `protocol_actual.cpp:59,69` |
| btc  | `protocol_legacy.cpp:57,67` | `protocol_actual.cpp:59,69` |
| dgb  | `protocol_legacy.cpp:57,67` | `protocol_actual.cpp:59,69` |
| bch  | `protocol_legacy.cpp:86,96` | `protocol_actual.cpp:94,104` |

## Changes

- **Invert the guard** to `!m_peers.empty()` on both branches in all five `protocol_legacy.cpp`.
- **Defence in depth:** `core::random::random_choice()` now throws `std::out_of_range` on an empty container instead of reading past the end. All 30 call sites already guard on `!empty()`, and the protocol dispatchers wrap `handle()` in `catch (const std::exception&)`, so a future missed guard degrades into a logged handler warning rather than a freed-memory read.
- **`bch/protocol_actual.cpp` header comment** claimed the inverted Legacy guard was a deliberate wire-generation difference. It was a defect; the comment is corrected. (The 127.0.0.0/127.0.0.1 half of that comment is left standing — see "Not addressed" below.)

## `HANDLER(addrs)` audit — confirmed already correct

Audited in all ten files. Every one guards on `(random_float(0,1) < 0.8) && (!m_connections.empty())` and relays through `m_connections`, not `m_peers`. No change needed; now pinned by test so it cannot drift into the same inversion.

## Blast radius

`handle_version` returns `PeerConnectionType::legacy` unconditionally on ltc (`node.cpp:365`), btc (`:366`), dgb (`:373`) and bch (`:366`), so Legacy is the only live dispatch on those four. DASH is the exception: its dispatch lives in `node.hpp:713-714` (not `node.cpp`) and *can* select Actual once the operator ratchets the min-protocol floor above the oracle baseline. Either way the Legacy path is live on every coin, and addrme relay is currently dead on all five.

No consensus surface: addrme/addrs are peer-book gossip only.

## Test

Folded into the **existing allowlisted `core_test`** target (`src/core/test/random_choice_guard_test.cpp` added to the existing `add_executable`) — never a new standalone target, which CI would not build and CTest would report "Not Run".

- `RandomChoiceGuard.*` — all four `random_choice` overloads throw on empty, and still return a real member for non-empty containers.
- `AddrmeRelayGuard.GuardsOnHavingPeersNotOnHavingNone` — source KAT over all ten `protocol_{legacy,actual}.cpp` addrme handlers: exactly two `!m_peers.empty()` guards each, zero bare `m_peers.empty()`.
- `AddrsRelayGuard.GuardsOnHavingConnections` — same pin for the addrs handler.

```
[==========] 96 tests from 16 test suites ran.
[  PASSED  ] 96 tests.
```

**Negative control:** re-inverting a single guard in `ltc/protocol_legacy.cpp` makes the KAT fail with
`ltc::legacy addrme still carries the inverted m_peers.empty() guard (#869)`.

Full `c2pool` target also builds clean with the `random.hpp` change.

## Not addressed here (separate finding)

All five `protocol_legacy.cpp` compare the addrme self-probe against **`"127.0.0.0"`**; `protocol_actual.cpp` and canonical p2pool both use **`"127.0.0.1"`**. The Legacy string matches no real loopback peer, so a genuine 127.0.0.1 addrme falls into the else branch, gets `got_addr()`d into the peer book and gossiped out as a routable address. That is a real parity defect but it is a behavioural change to p2p address handling rather than the guard inversion #869 describes, so it is deliberately left out of this PR. Happy to fold it in or open a follow-up — say which.

## One correction to the issue text

> Combined with the absence of any origination site, peers never learn our listen port

An origination site **does** exist on four of the five coins — `handle_version` writes a `message_addrme` carrying `core::Server::listen_port()` immediately after the handshake:
`ltc/node.cpp:343`, `btc/node.cpp:344`, `bch/node.cpp:344`, `dgb/node.cpp:351`.

**DASH is the exception** — `dash::message_addrme` is registered as a handler (`node.hpp:1125,1144`) but nothing ever writes one, so DASH genuinely has no origination path. That is a second, independent gap and is *not* fixed here (it means touching the DASH handshake write ordering, which is load-bearing). Flagging it for a follow-up.

### DIFF
--- a/src/core/random.hpp
+++ b/src/core/random.hpp
@@ -2,6 +2,7 @@
 #include <vector>
 #include <map>
 #include <list>
+#include <stdexcept>
 #include <core/uint256.hpp>
 
 namespace core::random
@@ -16,9 +17,23 @@ namespace core::random
 
 	uint256 random_uint256();
 
+    // random_choice on an EMPTY container used to be undefined behaviour:
+    // random_int(0, 0) returns 0, and the vector overload then indexed [0]
+    // past the end while the map/list overloads dereferenced end(). Every
+    // call site is expected to guard on !container.empty() first; this throw
+    // is the defence-in-depth backstop so a missed guard degrades into a
+    // logged handler error (the protocol dispatchers wrap handle() in a
+    // catch(const std::exception&)) instead of a freed-memory read.
+    [[noreturn]] inline void throw_empty_choice()
+    {
+        throw std::out_of_range("core::random::random_choice on empty container");
+    }
+
     template <typename T>
     T random_choice(std::vector<T> &list)
     {
+        if (list.empty())
+            throw_empty_choice();
         int pos = core::random::random_int(0, list.size());
         return list[pos];
     }
@@ -26,6 +41,8 @@ namespace core::random
     template <typename T>
     T random_choice(const std::vector<T> &list)
     {
+        if (list.empty())
+            throw_empty_choice();
         int pos = core::random::random_int(0, list.size());
         return list[pos];
     }
@@ -34,6 +51,8 @@ namespace core::random
     template <typename Key, typename Value>
     Value random_choice(std::map<Key, Value> _map)
     {
+        if (_map.empty())
+            throw_empty_choice();
         int pos = core::random::random_int(0, _map.size());
         auto item = _map.begin();
         std::advance(item, pos);
@@ -43,6 +62,8 @@ namespace core::random
     template <typename Value>
     Value random_choice(std::list<Value> _list)
     {
+        if (_list.empty())
+            throw_empty_choice();
         int pos = core::random::random_int(0, _list.size());
         auto item = _list.begin();
         std::advance(item, pos);
--- a/src/core/test/CMakeLists.txt
+++ b/src/core/test/CMakeLists.txt
@@ -30,7 +30,14 @@ if (BUILD_TESTING AND (GTest_FOUND OR GTEST_FOUND))
     # ONE definition instead of each growing a copy. Same reasoning again:
     # header-only over core, so it is FOLDED into the EXISTING allowlisted
     # core_test target, never a new add_executable (the #769 "Not Run" trap).
-    add_executable(core_test pack_test.cpp events_test.cpp chain_test.cpp packet_test.cpp block_broadcast_test.cpp broadcast_convergence_matrix_test.cpp core_merkle_branches_test.cpp version_gate_test.cpp filesystem_test.cpp web_server_submitblock_test.cpp known_txs_eviction_test.cpp tx_advertiser_test.cpp socket_write_queue_test.cpp known_txs_backing_test.cpp)
+    # random_choice_guard_test.cpp: the #869 addrme peer-relay guard. Covers the
+    # core/random.hpp empty-container backstop plus a source KAT pinning the
+    # !m_peers.empty() / !m_connections.empty() relay predicates across all five
+    # per-coin protocol_legacy.cpp / protocol_actual.cpp addrme+addrs handlers.
+    # Same reasoning as above: FOLDED into the EXISTING allowlisted core_test
+    # target, never a new add_executable.
+    add_executable(core_test pack_test.cpp events_test.cpp chain_test.cpp packet_test.cpp block_broadcast_test.cpp broadcast_convergence_matrix_test.cpp core_merkle_branches_test.cpp version_gate_test.cpp filesystem_test.cpp web_server_submitblock_test.cpp known_txs_eviction_test.cpp tx_advertiser_test.cpp socket_write_queue_test.cpp known_txs_backing_test.cpp random_choice_guard_test.cpp)
+    target_compile_definitions(core_test PRIVATE C2POOL_SRC_ROOT="${CMAKE_SOURCE_DIR}/src")
     target_link_libraries(core_test PRIVATE GTest::gtest_main core c2pool_merged_mining GTest::gtest)
     target_link_libraries(core_test PRIVATE c2pool_payout c2pool_hashrate ltc_coin)  # OBJECT-lib SCC direct-naming (#22/#39)
 
--- a/src/core/test/random_choice_guard_test.cpp
+++ b/src/core/test/random_choice_guard_test.cpp
@@ -1,0 +1,206 @@
+// Regression cover for issue #869 — the inverted addrme peer-relay guard.
+//
+// Two independent layers:
+//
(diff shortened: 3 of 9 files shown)


frstrtr:
CI note: `Linux x86_64 (AsAN+UBSan)` red on `VerifyShareThreading.Phase2HoldsLockAgainstComputePrune` — an unrelated timing flake, not this change.

```
test/test_threading.cpp:589: Failure
Expected: (io_ops.load()) > (0u), actual: 0 vs 0
IO body never acquired the lock — invariant not exercised
```

That assertion is the test's own self-check that the race window was entered: both IO threads take the chain mutex with `std::try_to_lock` and `continue` on failure, so under sanitizer slowdown on a loaded runner the compute thread can hold the write lock for every iteration and `io_ops` stays 0. The test asserts nothing about correctness in that state — it reports the window was never exercised.

`test/test_threading.cpp` includes only `impl/ltc/share*.hpp`, `sharechain/sharechain.hpp`, `core/pack.hpp` and the btclibs crypto headers — it does not include `core/random.hpp` and does not link any `protocol_legacy.cpp` path, so nothing in this PR's surface can reach it. Everything else in the same job is green (2367/2368), and the non-sanitizer lanes are green.

Re-running the job once the rest of the run drains.

frstrtr:
Follow-up on the AsAN flake: the sibling PR #877 — which touches a **larger** surface (`core/web_server.cpp` plus both `main_ltc.cpp` and `main_dash.cpp`) — just went **green on the same `Linux x86_64 (AsAN+UBSan)` job**. Same runner class, same sanitizer config, `VerifyShareThreading.Phase2HoldsLockAgainstComputePrune` passed there. That confirms the red here was runner-load scheduling, not code.

Locally, on this branch, the full suite is green including that test:

```
431/2365 Test #1054: VerifyShareThreading.Phase2HoldsLockAgainstComputePrune ... Passed 0.01 sec
...
2364/2365 Test #96: AddrmeRelayGuard.GuardsOnHavingPeersNotOnHavingNone ....... Passed 0.01 sec
2365/2365 Test #97: AddrsRelayGuard.GuardsOnHavingConnections ................. Passed 0.01 sec

100% tests passed, 0 tests failed out of 2365
```

Will re-run the job once the rest of this workflow drains (it is still queued, so `gh run rerun --failed` is refused).

frstrtr:
**INTEGRATOR REVIEW — APPROVE, merging.** 27/27 green, and the #883 flake that held this out twice is now fixed on master (#892).

The fix inverts `if (m_peers.empty() && ...)` to `!m_peers.empty()` on BOTH branches across all five coins' `protocol_legacy.cpp`, matching `protocol_actual.cpp` which was already correct — and adds a defensive bounds guard to `core::random::random_choice` so an empty container throws rather than dereferencing `end()`. `HANDLER(addrs)` audited and confirmed already correct in all ten files.

The liveness point is what makes this worth having: `handle_version` returns `PeerConnectionType::legacy` unconditionally on ltc/btc/dgb/bch, so Legacy is the ONLY live dispatch there and `addrme` has never been relayed on any coin. Ten characters of inversion, and it restores a peer-discovery path that has been silently dead.

Two follow-ups deliberately left out of scope and tracked as #882 — the Legacy self-probe compares against `127.0.0.0` (the loopback NETWORK address, which no peer can present) where canonical uses `127.0.0.1`, and DASH never originates an `addrme` at all. Correct call to keep those separate so each behavioural change is independently bisectable.

Thanks also for correcting my issue text — I wrote that we never originate addrme, which was wrong for four of five coins.

PR #828, opened 2026-07-24
dash(node): close the join-burst tracker read races in the parallel download/think path

frstrtr:
## Summary

Fixes the SIGSEGV/general-protection fault a joining c2pool-dash node hits during the initial concurrent sharechain download burst (GP fault in libc.so.6 at ~4337 shares into an ~8848-share join, ASYNC-THINK/ASYNC-DEFER active; kernel: `traps: c2pool-dash[...] general protection fault ... in libc.so.6`). This is the exact path a production node exercises when it repoints to the live p2pool-dash sharechain, so it gates that repoint.

## Root cause — unlocked tracker reads racing the compute thread

The DASH node's documented concurrency contract (`src/impl/dash/node.hpp`, `m_tracker_mutex` block) is:

- compute thread (`m_think_pool`, 1 thread): `unique_lock(m_tracker_mutex)` — exclusive, blocking
- IO-thread reads: `shared_lock(try_to_lock)` — non-blocking, skip if busy
- IO-thread writes: `unique_lock(try_to_lock)` — non-blocking, queue if busy

During the join burst, `think()` runs almost continuously on the compute thread holding the exclusive lock and **structurally mutating the chain containers** — `verified.add(...)` during the bootstrap verify pass, `chain.remove(...)` of unverifiable shares — which rehashes/reshapes the underlying `unordered_map`s.

Five hot call sites read those same containers with **no lock at all**, from the IO thread and the WebServer HTTP thread:

1. `NodeImpl::best_share_hash()` (node.cpp) — `dash::mint::elect_best_share()` iterates `verified.get_heads()` and calls `verified.get_work()` on every call. Called from the stratum work-refresh on the IO thread and from the WebServer HTTP thread (pool-hashrate fn, `set_best_share_hash_fn`).
2. `NodeImpl::advertised_best_share()` (node.hpp) — iterates `chain.get_heads()` + `chain.get_height()` on every `send_version` handshake.
3. `handle_version()` (node.hpp) — `m_chain->contains(msg->m_best_share)`.
4. The share-download backfill continuation (node.cpp, `download_shares` reply lambda) — `m_chain->contains(next)` on **every sharereply** — the hottest site during the burst.
5. `drain_peer_best_adverts()` (node.cpp) — `m_chain->contains(best)` on the 2s advert timer and every think() IO-phase.

[...]

PR #873, opened 2026-07-26
ltc/doge: gate share broadcast on tx completeness, mark only what was sent

frstrtr:
## Summary

Two reward-path share-loss fixes on the **shipping LTC/DOGE lane**, ported from the DASH lane. Neither changes a share byte, a consensus rule, or a minting decision — they change only **whether and when** a share is put on the wire.

DOGE rides this lane as an AuxPoW child on the LTC sharechain, so both fixes cover it.

---

## FIX A — F3 tx-completeness broadcast gate

`send_shares()` collected the txs a peer still needs, looked each one up in `m_known_txs`, and **silently dropped** the ones it could not find — writing the share anyway (`src/impl/ltc/node.cpp`, pre-fix L745-749):

```cpp
auto it = m_known_txs.find(th);
if (it != m_known_txs.end())
    full_txs.emplace_back(it->second);
// no else, no missing counter, share written regardless
```

Canonical p2pool drops the connection over a share referencing a transaction it does not hold (`p2p.py:404`, "referenced unknown transaction") → sharechain isolation → orphaned shares → direct PPLNS loss. #834's bounded eviction (`core/known_txs_eviction.hpp`) made this rarer, not unreachable.

Now: a share is withheld unless we hold the bytes of **every** new tx it references, and an unfindable tx logs loudly (`++missing` + `LOG_WARNING`) instead of vanishing.

* Withholding costs no propagation — `m_known_txs` in this lane is fed **solely** by inbound peer `remember_tx`, so an unbackable share is by construction one we downloaded from a network that already has it.
* The batch is filtered **per share**, never rejected wholesale, so an unbackable ancestor cannot suppress the tip that carries the PPLNS credit.
* `m_remote_txs` (the peer's have_tx advert) is still used only to choose hash-vs-bytes encoding, never to decide whether to send — the advert can be stale, and that staleness is exactly the disconnect.

**src/core/ hoist.** The gate predicate is coin-generic, so it now lives in `src/core/known_txs_backing.hpp`. `src/impl/dash/known_txs_retention.hpp` re-exports it (`using core::all_txs_backable;`) so every existing DASH call site resolves unchanged and no lane grows a second copy. The BTC/DGB/BCH sibling should consume this header rather than adding one.

### Additional finding: the tx-forwarding block was unreachable

While verifying the premise I found the block was **dead code**. It probed

```cpp
if constexpr (requires { obj->m_new_transaction_hashes; })
```

directly on the share object — and that expression is **false for every ltc share variant**:

| variant | version | new-tx list |
|---|---|---|
| `ltc::Share` | 17 | `m_tx_info.m_new_transaction_hashes` (nested) |
| `ltc::NewShare` | 33 | `m_tx_info.m_new_transaction_hashes` (nested) |
| `ltc::SegwitMiningShare` | 34 | none — `Formatter` serialises `m_tx_info` only for `version < 34` |
| `ltc::PaddingBugfixShare` | 35 | none |
| `ltc::MergedMiningShare` | 36 | none |

So this lane **never forwarded a single tx byte**, and a relayed v17/v33 share went out unbacked every time — a strictly worse form of the reported bug. The receive side (`protocol_actual.cpp:123`, `protocol_legacy.cpp:118`) spells it correctly, which is why only the send side was dead.

`src/impl/ltc/share_tx_refs.hpp` introduces one accessor handling both spellings and reporting "no list" for variants that have none, so a member-name mismatch can no longer disable the path silently.

**Safety for the shipping line:** the share versions this lane mints (v34/v35/v36) carry no new-tx list at all, so they are trivially backable — the gate is a strict no-op for locally minted shares and can never withhold our own work. Pinned by `LocallyMintedV36SharesAreNeverWithheld`, which runs the gate against an entirely empty known-tx cache.

*Not ported:* DASH's mint-time decline (`dash/node.cpp:1400-1410`). It has no LTC analogue (no `register_template_txs` here) and it would change minting semantics, which this PR does not do.

---

## FIX B — F2 mark only what was actually sent

`broadcast_share()` inserted into `m_shared_share_hashes` inside the chain-walk loop, **before** `send_shares()` ran. `send_shares` has early returns that abandon the whole batch (tracker-lock miss, empty batch), and with zero peers connected nothing is written at all — yet the shares were already marked broadcast. The next walk `break`s on that mark, there is no retry path, and the set is only ever cleared wholesale on overflow (`node.cpp:1442`). A chain **tip** lost this way is gone silently.

`send_shares` now returns `std::vector<uint256>` of the hashes it actually wrote; `broadcast_share` marks only the union across peers, after the loop. Anything skipped or abandoned stays unmarked and is retried next cycle. Mirrors DASH (`dash/node.cpp:1236-1241`, rationale at `:1216-1221`).

`readvertise_best_share()` deliberately ignores the return — it is advertise-only and never consults the de-dup set. Unchanged.

---

## Verification of the premise (corrections)

Every claim was checked against master `4d439b42`. Line numbers were off by a few in places; substance held except where noted:

| claim | actual | verdict |
|---|---|---|
| DASH gate `dash/node.cpp:1287-1309` | 1287-1310 | ✅ |
| DASH decline-mint `1400-1410` | 1398-1420 | ✅ |
| DASH missing-tx accounting `1338-1343` | 1336-1343 | ✅ |
| DASH F2 `1236-1241`, rationale `1216-1221` | exact | ✅ |
| LTC silent omission `ltc/node.cpp:745-749` | 745-749 | ✅ text, but **unreachable** (see above) |
| LTC `broadcast_share` @ 791 | 791 | ✅ |
| LTC premature mark @ 822 | 822 | ✅ |
| LTC early returns `687-695` / `713` | 687-695 / 712-713 | ✅ |
| wholesale clear @ 1442 | 1443-1444 | ✅ (off by one) |
| `#834` eviction `ltc/node.cpp:1450-1451` | 1450-1451 | ✅ |

**Unconfirmed / out of scope:** v34+ ltc shares carry no `new_transaction_hashes` in this implementation, while canonical p2pool's `share_info` carries the field for all versions ≥ 13. That may be a genuine wire divergence, but it is consensus surface and is **not** touched here. Flagged for a separate look.

Also noted: `test/test_dash_known_txs_retention.cpp` is absent from the `build.yml --target` allowlist and lives outside the per-coin tree the drift-guard audits, so it is never built in CI. Not fixable here (token lacks `workflow` scope).

---

## Tests

Folded into the **existing allowlisted** `core_test` and `share_test` targets — never a new `add_executable`, which would be absent from the `--target` list and reported "Not Run" (the #769 trap). `tools/ci/check_test_target_allowlist.py` passes: *119 per-coin test targets across 5 coin lanes all present*.

* `src/core/test/known_txs_backing_test.cpp` (10 cases) — gate and mark-after-send semantics generically, including `UnbackableShareIsNotBroadcast`, `GateFiltersPerShareAndKeepsTheBackableTip`, `AbandonedBatchIsNotMarkedBroadcast`, `ZeroPeersMarksNothing`.
* `src/impl/ltc/test/broadcast_tx_completeness_test.cpp` (9 cases) — the same, against the **real** `ltc::ShareType` variants, plus compile-time pins of the member-spelling bug.

### Local run (gcc-13, Release, conan `ci/conan/linux-gcc13.profile`)

```
core_test    : 99 tests from 14 test suites ran. [  PASSED  ] 99 tests.
share_test   : 55 tests from 11 test suites ran. [  PASSED  ] 55 tests.
```

DASH lane unaffected by the header hoist:

```
test_dash_node        : 8 tests ran. [  PASSED  ] 8 tests.
test_dash_broadcaster : 8 tests ran. [  PASSED  ] 8 tests.
```

`c2pool` binary builds clean.

### Negative control

Reverting the test's accessor to the pre-fix probe (`requires { obj->m_new_transaction_hashes; }`) fails 4 of the 9 LTC cases:

```
[  FAILED  ] LtcBroadcastTxCompleteness.AccessorReachesTheListOnV17AndV33
[  FAILED  ] LtcBroadcastTxCompleteness.UnbackableV17ShareIsNotBroadcast
[  FAILED  ] LtcBroadcastTxCompleteness.GateKeepsTheBackableTipWhenAnAncestorIsUnbackable
[  FAILED  ] LtcBroadcastTxCompleteness.GatedShareIsNotMarkedShared
```

Honest limitation: the node-level glue (`peer->write`, the peer loop) has no unit-test seam in this lane — there is no LTC `NodeImpl` harness. The tests pin the decision functions the node calls; the wiring is covered by review and by the build.

---

## Files touched

```
src/core/known_txs_backing.hpp              (new)  coin-generic gate + mark helpers
src/core/test/known_txs_backing_test.cpp    (new)  folded into core_test
src/core/test/CMakeLists.txt                       + source
src/impl/dash/known_txs_retention.hpp              re-export, de-dup
src/impl/ltc/share_tx_refs.hpp              (new)  new-tx-list accessor
src/impl/ltc/node.cpp                              gate, accounting, F2 return
src/impl/ltc/node.hpp                              send_shares signature
src/impl/ltc/test/broadcast_tx_completeness_test.cpp (new)  folded into share_test
src/impl/ltc/test/CMakeLists.txt                   + source
```

Untouched as instructed: `src/core/socket.cpp`, `src/core/web_server.cpp`, `src/impl/{btc,dgb,bch}/`, `protocol_legacy.cpp`, `.github/workflows/`.

**Draft — do not merge.**

### DIFF
--- a/src/c2pool/main_ltc.cpp
+++ b/src/c2pool/main_ltc.cpp
@@ -4938,7 +4938,14 @@ int main(int argc, char* argv[]) {
                     if (p.stale_info == 253)      stale = ltc::StaleInfo::orphan;
                     else if (p.stale_info == 254)  stale = ltc::StaleInfo::doa;
 
-                    // tracker_lock (blocking shared_lock) acquired at top of try block.
+                    // tracker_lock: EXCLUSIVE unique_lock(try_to_lock), acquired at
+                    // the top of this try block (rationale there — this path WRITES
+                    // the tracker via create_local_share). NOT a blocking
+                    // shared_lock, as this comment used to claim. It is held across
+                    // everything below, which is why the broadcast must be POSTED
+                    // rather than called inline: broadcast_share takes a shared
+                    // try-lock and a shared_mutex refuses that to a thread that
+                    // already holds it exclusively.
 
                     // Pass frozen fields from template time so ref_hash matches coinbase.
                     uint256 share_hash = ltc::create_local_share(
@@ -4979,11 +4986,22 @@ int main(int argc, char* argv[]) {
                     }
                     ++s_created;
 
-                    // Broadcast to all connected peers
+                    // Broadcast to all connected peers.
+                    //
+                    // POSTED, not called inline. We are holding tracker_lock
+                    // EXCLUSIVELY right here, and broadcast_share opens with a
+                    // shared try-lock on that same mutex — which a thread holding
+                    // the write lock can never obtain. Calling it inline took the
+                    // "tracker busy — deferring" early return on every single
+                    // share, so this node broadcast nothing it minted. Posting
+                    // runs it on the io thread after this scope has released the
+                    // lock, and keeps the m_known_txs read in send_shares on the
+                    // same thread as the unlocked remember_tx ingest that writes
+                    // that map.
                     try {
-                        p2p_node->broadcast_share(share_hash);
+                        p2p_node->post_broadcast_share(share_hash);
                     } catch (const std::exception& e) {
-                        LOG_ERROR << "broadcast_share failed: " << e.what();
+                        LOG_ERROR << "post_broadcast_share failed: " << e.what();
                     }
 
                     // CRITICAL: Update best share immediately so refresh_work()
--- a/src/core/known_txs_backing.hpp
+++ b/src/core/known_txs_backing.hpp
@@ -1,0 +1,99 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+//
+// Coin-generic tx-completeness gate for share broadcast ("F3"), hoisted out of
+// impl/dash/known_txs_retention.hpp so every node lane consumes ONE definition.
+//
+// Canonical p2pool broadcasts a share together with the bytes of every new tx it
+// references (remember_tx). A receiving canonical node that is handed a share
+// referencing a transaction it does not hold DROPS THE CONNECTION
+// ("referenced unknown transaction", p2p.py:404) -> sharechain isolation ->
+// our shares orphan -> direct PPLNS loss.
+//
+// So a share may only be put on the wire when we HOLD the bytes of every
+// referenced new tx (the node's m_known_txs relay cache), because that is the
+// only way remember_tx can always forward them. Gating on the peer's have_tx
+// advert instead is weaker than canonical: the advert can be stale (the peer may
+// have dropped the tx), and sending hash-only for a tx the peer no longer holds
+// is exactly the disconnect. The advert is therefore only good for choosing
+// hash-vs-bytes encoding, never for deciding whether to send at all.
+//
+// Withholding costs no propagation: a share we could not back is by definition
+// one we downloaded from the network (a sharereply carries no tx bytes), so the
+// network already has it. What withholding buys is that we stay connected.
+//
+// Reward-safety: these helpers decide WHETHER a share is broadcast. They never
+// touch share bytes, consensus, subsidy, coinbase, payee or minting state.
+
+#pragma once
+
+#include <cstddef>
+#include <utility>
+#include <vector>
+
+#include <core/uint256.hpp>
+
+namespace core {
+
+// True iff every hash in `tx_hashes` is present in `held` (the tx bytes we hold,
+// i.e. the node's m_known_txs). An empty list is trivially backable — a share
+// that references no new tx can never trigger the peer's unknown-tx disconnect.
+//
+// `held` needs only find()/end() (std::map, std::unordered_map, std::set...).
+template <typename Held>
+inline bool all_txs_backable(const std::vector<uint256>& tx_hashes,
+                             const Held& held)
+{
+    for (const auto& h : tx_hashes)
+        if (held.find(h) == held.end())
+            return false;
+    return true;
+}
+
+// F3 gate applied to a batch: drop from `shares` every entry we cannot back,
+// preserving order. Returns the number dropped (for logging / accounting).
+//
+// `tx_hashes_of(share)` must return a pointer to that share's new-tx-hash vector,
+// or nullptr when the share type carries no tx-hash list at all (then the share
+// is trivially backable). A pointer — not a copy — so the hot broadcast path
+// does no per-share allocation.
+//
+// The batch is filtered rather than rejected wholesale: one unbackable ancestor
+// in a chain walk must not suppress the backable tip, which is the share our
+// PPLNS credit actually depends on.
+template <typename Share, typename TxHashesOf, typename Held>
+inline std::size_t retain_backable_shares(std::vector<Share>& shares,
+                                          TxHashesOf&& tx_hashes_of,
+                                          const Held& held)
+{
+    std::vector<Share> keep;
+    keep.reserve(shares.size());
+    std::size_t dropped = 0;
+    for (auto& share : shares) {
+        const std::vector<uint256>* hashes = tx_hashes_of(share);
+        if (hashes == nullptr || all_txs_backable(*hashes, held))
+            keep.push_back(std::move(share));
+        else
+            ++dropped;
+    }
+    shares.swap(keep);
+    return dropped;
+}
+
+// F2 companion: fold the per-peer "these hashes actually reached the wire"
+// reports into the caller's already-broadcast de-dup set.
+//
+// The de-dup set must be advanced ONLY here, i.e. AFTER the peer loop. Marking a
+// share at walk time (before the send) permanently retires a share the send then
+// abandoned — a lock miss, an empty batch, zero peers connected, or the F3 gate
+// above — and nothing ever re-pushes it, because the next walk breaks on the
+// marked hash. That is a silently orphaned share and a direct PPLNS loss. An
+// unmarked share simply gets retried by the next broadcast cycle.
+template <typename MarkSet>
+inline void commit_broadcast_marks(MarkSet& marked,
+                                   const std::vector<uint256>& sent)
+{
+    for (const auto& h : sent)
+        marked.insert(h);
+}
+
+} // namespace core
--- a/src/core/test/CMakeLists.txt
+++ b/src/core/test/CMakeLists.txt
@@ -23,7 +23,14 @@ if (BUILD_TESTING AND (GTest_FOUND OR GTEST_FOUND))
     # every coin because all five node lanes share this one socket layer. Same
     # reasoning as above: FOLDED into the EXISTING allowlisted core_test target,
     # never a new add_executable (the #769 "Not Run" trap).
-    add_executable(core_test pack_test.cpp events_test.cpp chain_test.cpp packet_test.cpp block_broadcast_test.cpp broadcast_convergence_matrix_test.cpp core_merkle_branches_test.cpp version_gate_test.cpp filesystem_test.cpp web_server_submitblock_test.cpp known_txs_eviction_test.cpp tx_advertiser_test.cpp socket_write_queue_test.cpp)
+    #
+    # known_txs_backing_test.cpp: the coin-generic share-broadcast tx-completeness
+    # gate and the mark-only-what-was-sent de-dup rule (core/known_txs_backing.hpp),
+    # hoisted out of impl/dash/known_txs_retention.hpp so every node lane consumes
+    # ONE definition instead of each growing a copy. Same reasoning again:
+    # header-only over core, so it is FOLDED into the EXISTING allowlisted
+    # core_test target, never a new add_executable (the #769 "Not Run" trap).
+    add_executable(core_test pack_test.cpp events_test.cpp chain_test.cpp packet_test.cpp block_broadcast_test.cpp broadcast_convergence_matrix_test.cpp core_merkle_branches_test.cpp version_gate_test.cpp filesystem_test.cpp web_server_submitblock_test.cpp known_txs_eviction_test.cpp tx_advertiser_test.cpp socket_write_queue_test.cpp known_txs_backing_test.cpp)
     target_link_libraries(core_test PRIVATE GTest::gtest_main core c2pool_merged_mining GTest::gtest)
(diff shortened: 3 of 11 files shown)


frstrtr:
Review verdict: **CHANGES-REQUIRED**. Not a regression — the PR is mechanically safe and DASH is byte-identical — but its headline fix does not currently execute. Three items before this lands.

**1. HIGH — the F2 machinery is wired into unreachable code.** `broadcast_share` is only called from `main_ltc.cpp:4962`, inside a scope that holds an **exclusive** `unique_lock<shared_mutex>` on the tracker mutex taken at `main_ltc.cpp:4831` and never released before the call. `broadcast_share` then attempts `shared_lock(m_tracker_mutex, try_to_lock)` at `ltc/node.cpp:875`, which fails when the calling thread already holds the write lock. It always takes the "tracker busy — deferring" early return. So `commit_broadcast_marks` runs zero times in production — and symmetrically, the premature-marking bug this PR fixes never fired either, because the old insert sat after the same failing lock. Filed separately as #878.

Fix: release `tracker_lock` before the call, or post the broadcast to the io_context. **Read item 2 first — un-deadening this path exposes a race.**

**2. MED, blocks item 1 — `m_known_txs` race.** `HANDLER(remember_tx)` inserts into `m_known_txs` on the IO thread with no tracker lock at all (`protocol_actual.cpp:262-265`, `protocol_legacy.cpp:298-301`), while the compute-thread evictor holds exclusive. This is survivable today only because every live reader is on the same IO thread. The new gate adds a reader on the solve thread; the moment item 1 is fixed naively, it races an unlocked `emplace` on a `std::map` — the documented GP-fault class (#828). Posting the broadcast to the io_context keeps `m_known_txs` single-threaded and avoids it; otherwise bring the remember_tx insert under the try-lock discipline. A shared lock on the reader alone does not help against a writer that takes no lock.

**3. MED — `m_last_broadcast_to` records the pre-gate batch.** `broadcast_share` (`node.cpp:916`) and `readvertise_best_share` (`node.cpp:1097`) record the full `to_send` set even for peers where `send_shares` returned `{}` or the gate withheld shares. `error()`/`close_connection()` then permanently move those hashes into `m_rejected_share_hashes` (never cleared) on a disconnect within 10s, after which we also refuse to serve them to syncing peers. So a share the gate withheld — never a byte on the wire — can be branded "peer-rejected" forever. Record the per-peer `sent` vector instead of `to_send`; one line in each loop.

**Confirmed good, keep as-is:** the gate provably cannot withhold our own minted shares (v35/v36 are the only versions this lane mints, neither carries a tx list, and the Formatter serialises `m_tx_info` only below v34) — please add the tripwire comment that this holds *only* while that is true, since `m_known_txs` never contains template txs in this lane. The DASH re-export is behaviour-preserving (qualified lookup, token-identical template, single definition). Consensus surface untouched. Omitting DASH's mint-time decline is correct, not a gap.

**Also fix while here:** `main_ltc.cpp:4919` says "blocking shared_lock" — it is a unique try-lock. `ltc/node.cpp:1064-1066` "de-dup-marked at creation" is now false twice over.

Do not merge on the current check rollup — Linux/ASAN/Windows were still pending at review time.

frstrtr:
## Review follow-up — all three items addressed

Thanks, this was the right catch. Item 1 is worse than a wiring nit: it means the previous commit's headline fix was correct code in a function that never ran, and *symmetrically* that the premature-marking bug it fixed had never fired either. Both sat behind the same failing lock.

I verified the lock structure independently before changing anything:

* `main_ltc.cpp:4831-4832` — `std::unique_lock<std::shared_mutex> tracker_lock(p2p_node->tracker_mutex(), std::try_to_lock)`, **exclusive**, held unbroken through the `broadcast_share` call at `:4962` to the end of the lambda at `:5000`.
* `ltc/node.cpp:875` — `std::shared_lock(m_tracker_mutex, std::try_to_lock)`, which a thread already holding the write lock can never obtain.
* For contrast, `main_btc.cpp:1324` already does `lk.unlock();` with the comment *"drop the unique_lock BEFORE calling broadcast_share / notify_local_share"* — so the BTC lane does not have this bug and LTC is the outlier.

Confirmed. Fixed as follows.

---

### 1 — Un-deadened via `post_broadcast_share()` (io_context), not `unlock()`

Took the io_context option precisely for the reason you flagged in item 2. New public entry point on `NodeImpl`; the submit path calls it instead of `broadcast_share`. `broadcast_share` itself is now documented as "io thread, no tracker lock held by the caller".

Three robustness consequences of making the broadcast asynchronous, all handled:

* Added `IsNull()` + `m_chain->contains()` guards — a posted broadcast can outlive the share's presence in the chain, and `get_height`/`get_chain` on an absent hash is not a defined query. (DASH's `broadcast_share` already has the `contains` guard; LTC did not.)
* The posted handler swallows exceptions rather than unwinding into the io_context. The old inline call site had a `try`/`catch` that the post would otherwise have discarded.
* `pool::BaseNode`'s default ctor leaves `m_context` / `m_chain` / `m_config` **indeterminate** (its own `// todo: init` markers). Anything null-checking them on that path reads a garbage pointer. Nulled in `ltc::NodeImpl`'s default ctor rather than touching the shared base that the other coin lanes derive from.

**Self-diagnosis added.** A run of deferrals with zero successful acquisitions now logs a warning. Genuine contention with the compute thread is bursty and always interleaved with successes; a *pure* run of failures means a caller is holding the lock across the call and nothing is being broadcast at all. That is the signature this defect would have shown from day one, and it is now impossible to have silently again. Backed by `broadcast_deferred_count()` / `broadcast_acquired_count()` / `broadcast_reached_send_count()`.

### 2 — `m_known_txs` stays io-thread confined

Posting is what makes this safe: the gate's read now happens on the same thread as the unlocked `remember_tx` `emplace`, so no cross-thread reader is added. Had I used `tracker_lock.unlock()`, the read would have run on the submit thread against an unlocked `std::map::emplace` — and as you say, a shared lock on the reader alone would not have helped.

Verified the io_context is genuinely single-threaded: `main_ltc.cpp` runs `ioc.run_for()` from one loop; the only other `std::thread` is the external watchdog, which never touches `ioc`.

The discipline and its limits are now written at the declaration in `node.hpp`, including — explicitly labelled pre-existing and **not** fixed here — that `prune_shares()` evicts from this map on the compute thread under the exclusive lock, which the unlocked io-thread insert does not respect. That race exists independently of this PR and wants its own issue.

### 3 — `m_last_broadcast_to` records the per-peer `sent` vector

One line in each loop, `broadcast_share` and `readvertise_best_share`. Traced the consumer to confirm the severity: `error()` (`node.cpp:160-174`) and `close_connection()` (`:217-225`) move those hashes into `m_rejected_share_hashes` on a disconnect within 10s, and that set gates both re-broadcast and `readvertise`. Branding a never-sent share is permanent.

### Tripwire + stale comments

Gate tripwire now states the **failure mode**, not just the premise: if this lane ever mints a version carrying a new-tx list, the gate withholds *every locally minted share* — total broadcast suppression — because `m_known_txs` never holds our own template txs, only what peers pushed via `remember_tx`. It names the prerequisite (port DASH's `register_template_txs` **and** the mint-time decline together) and the two test cases that pin the premise.

`main_ltc.cpp:4919` "blocking shared_lock" → corrected. `ltc/node.cpp:1064-1066` "de-dup-marked at creation" → corrected, noting it is now false on both counts.

---

## Testing — the seam exists

You asked for a seam proving `broadcast_share` actually reaches `send_shares`. **Built one, against the real node.** `ltc::NodeImpl`'s default ctor takes no io_context, opens no LevelDB and starts no timers, so a test can own one; `ShareChain::add` takes a share pointer directly. `src/impl/ltc/test/broadcast_lock_discipline_test.cpp` builds a real two-share chain on a real `NodeImpl` and drives the real `broadcast_share`. Folded into the existing allowlisted `share_test`.

```
LtcBroadcastLockDiscipline.ExclusiveLockOnCallingThreadBlocksTheEntireBroadcast
LtcBroadcastLockDiscipline.WithoutTheCallersLockTheBroadcastReachesTheSendLoop
LtcBroadcastLockDiscipline.DeferredThenRetriedSucceeds
LtcBroadcastLockDiscipline.ZeroPeersMarksNothingOnTheRealNode
LtcBroadcastLockDiscipline.NullAndUnknownShareHashesAreRejectedNotWalked
```

The first reproduces the production defect exactly — caller holds the mutex exclusively, same thread — and asserts `acquired == 0` and `reached_send == 0`. The second asserts that without that lock the walk runs and the per-peer loop is entered.

**Negative control, on the real node:** reinstating the pre-fix walk-time `m_shared_share_hashes.insert(hash)` fails `ZeroPeersMarksNothingOnTheRealNode`. So the F2 fix is now genuinely regression-tested against `NodeImpl`, not only against the helper.

```
share_test : 60 tests, 12 suites — PASSED 60   (was 55; +5)
core_test  : 99 tests, 14 suites — PASSED 99
c2pool binary builds clean (gcc-13, Release)
CI drift-guard: 119 per-coin targets across 5 lanes, all allowlisted
```

Earlier negative control still holds: reverting the accessor to the pre-fix probe fails 4 of the 9 `LtcBroadcastTxCompleteness` cases.

### What I still cannot test, stated plainly

**No test can fail on a revert of the `main_ltc.cpp` call site.** `main_ltc.cpp` is a `main()` translation unit and is not linkable into a gtest binary, so the call site itself has no automated guard. What the suite does cover is the contract on both sides — call it under an exclusive lock and it does nothing; call it clean and it reaches the send loop — plus the runtime warning that now fires if the wiring regresses in production. That is the honest boundary.

Two further gaps I did not close:

* The **peer loop body** (`peer->write`) is still untested — constructing a `pool::Peer` needs a live socket. Three lines, exercised only in soak.
* **`notify_local_share` has the same defect and I did not fix it.** `main_ltc.cpp:4971` calls it inside the same exclusive scope, and it opens with `unique_lock(try_to_lock)`, which also always fails there. Its documented degradation is graceful ("the share is already in-chain and `run_think()` will score it next cycle"), it is a work-creation path rather than a broadcast path, and it is out of this PR's scope — but it is the same class of bug and is worth its own issue alongside #878.

Still **draft**, not merging. Check rollup needs to go green first — and per the earlier note, that means confirming the checks are *present*, not merely not-red.

Issue #884, opened 2026-07-26
dgb: set_mint_share_fn is never bound in main_dgb.cpp — DGB cannot mint local shares


frstrtr:
## Summary

`main_dgb.cpp` never calls `DGBWorkSource::set_mint_share_fn`, so the mint callback is unbound in production. A DGB node cannot record a locally-mined share — it can only relay shares received from peers.

## Verified

- `src/c2pool/main_dgb.cpp` contains **zero** occurrences of `set_mint_share_fn` and **zero** of `broadcast_share`.
- Repo-wide, the only DGB call sites are the definition (`src/impl/dgb/stratum/work_source.{cpp,hpp}`) and unit tests (`src/impl/dgb/test/work_source_test.cpp:149,188,285,312`).
- Compare: DASH binds it in `main_dash.cpp`; BTC binds its equivalent in `main_btc.cpp`.
- `dgb::mint_local_share_with_ratchet` (`src/impl/dgb/run_loop_mint.hpp:69`) — the intended adapter — is referenced only from `src/impl/dgb/test/share_test.cpp:657`.

Every share meeting share-target therefore takes the null-fn branch at `src/impl/dgb/stratum/work_source.cpp:810-814`:

```cpp
if (!fn) {
    LOG_WARNING << "[DGB-STRATUM] share met share-target but no mint callback "
                   "wired -- share NOT recorded (set_mint_share_fn unbound). "
                   "No silent drop: logged.";
    return uint256{};
}
```

Credit where due: this logs rather than dropping silently, so it is discoverable in any DGB node's log with `grep "set_mint_share_fn unbound"`.

## Consequences

1. `dgb::NodeImpl::broadcast_share` (`dgb/node.cpp:772`) and `notify_local_share` (`:817`) have **zero callers repo-wide** — dead code, not merely a dead path.
2. DGB's only outbound share push is `readvertise_best_share()` (`dgb/node.cpp:1001`) → `send_shares()` (`:1037`), which re-pushes the advertised head walk. That is peer-sourced shares only.
3. A DGB node contributes no hashrate to its sharechain.

## Relevance to PR #880

#880 wires a tx-completeness gate and mark-after-send into `send_shares`/`broadcast_share` for btc/dgb/bch. Per the caller-side lock trace:

- **BTC and BCH are reachable** and the PR is live code there (BTC releases at `main_btc.cpp:1323`; BCH at `pool_entrypoint.hpp:562`).
[...]

Issue #878, opened 2026-07-26
ltc: broadcast_share() is unreachable — caller holds the exclusive tracker lock its shared try-lock needs


frstrtr:
## Summary

`ltc::NodeImpl::broadcast_share()` can never execute past its first guard. Local share broadcast on LTC/DOGE relies entirely on `readvertise_best_share()` instead.

## Trace (verified)

- The **only** caller in the tree is `src/c2pool/main_ltc.cpp:4962`, inside the `set_create_share_fn` solve lambda.
- That lambda acquires an **exclusive** lock at `main_ltc.cpp:4831`:
  ```cpp
  std::unique_lock<std::shared_mutex> tracker_lock(p2p_node->tracker_mutex(), std::try_to_lock);
  ```
  and does **not** release it before line 4962 — the lock's scope covers the whole try-block.
- `broadcast_share` then attempts `std::shared_lock(m_tracker_mutex, std::try_to_lock)` at `src/impl/ltc/node.cpp:875`.
- `try_lock_shared` on a `shared_mutex` whose write lock is already held **by the calling thread** fails on every mainstream implementation (glibc `pthread_rwlock_tryrdlock` returns EBUSY when a writer holds it, including self; Windows `TryAcquireSRWLockShared` likewise).

So `broadcast_share` always takes the early return at `node.cpp:876-884` ("tracker busy — deferring") and never reaches the chain walk, `send_shares`, or the de-dup marking.

## Consequences

1. `m_shared_share_hashes` is never written by this path — its only inserter is inside the unreachable region.
2. Local share propagation happens **only** via `readvertise_best_share()` — best-change in the think IO-phase (`node.cpp:1794-1796`) plus a 10 s timer (`node.cpp:1801`) — which deliberately ignores the de-dup set. That is why production works today, but it means a freshly minted share waits for the next best-change or up to 10 s instead of going out immediately. On a 15 s `SHARE_PERIOD` that is a material propagation delay and a plausible contributor to orphan rate.
3. `notify_local_share` (`node.cpp:938`) has the same self-locked pattern, so its inline verify is also always skipped and only `run_think()` does the work.

## Empirical confirmation

On any production LTC node:
```
grep "\[broadcast_share\] tracker busy — deferring" <logfile>
```
It is logged every 50th occurrence (`node.cpp:878-882`). A steady stream confirms this; total absence would mean the function is never even called.

## Fix — and a hazard that comes with it

[...]

PR #888, opened 2026-07-26
coins: mint the block-winning share onto the sharechain (dash/btc/bch/dgb)

### LINKED ISSUE #887: dash/btc/bch/dgb: block-winning share is never minted into the sharechain (LTC is the only correct lane)
## Summary

On dash, btc, bch and dgb the block-target branch dispatches the block and **returns without minting the winning share into the sharechain**. Canonical p2pool adds the block-solving share. **LTC is the only lane that gets this right.**

Sites where the block arm returns before the mint arm:
- `src/impl/dash/stratum/work_source.cpp:954`
- `src/impl/btc/stratum/work_source.cpp:925`
- `src/impl/bch/stratum/work_source.cpp:738`
- `src/impl/dgb/stratum/work_source.cpp:613`

LTC is correct because `m_create_share_fn` (`src/core/web_server.cpp:7463`) runs **before and independently of** the block check, so the share is created either way.

## Why it matters

A share that meets the block target also, by definition, meets the share target — it is the **highest-work share we will ever produce**. Discarding it costs:

1. **Our own PPLNS weight** for that share, in the very window the block pays out from.
2. **Peers' view of our best work** — we never advertise the strongest share on the chain, which weakens our position in the sharechain race.

This is not block loss. The block is submitted correctly on all four coins (verified). It is share-weight loss, and it recurs on every block we win.

## Note on DGB

DGB additionally cannot mint any local share at all, because `set_mint_share_fn` is never bound in `main_dgb.cpp` (#884). Fixing that binding will not fix this issue on its own — the block arm still returns before reaching the mint seam.

## Suggested fix

Mirror LTC: mint the share first (or on both arms), then dispatch the block. The two operations are independent — the sharechain and the coin blockchain are separate systems, and a share can and should land in both.

Found in a cross-lane audit of block-submission paths. The primary question — "can a sharechain orphan still become a block" — came back **clean on all six coins**; this was surfaced by the same sweep.

frstrtr:
Fixes #887.

## The defect

`mining_submit` classifies a solve tighten-first: block target, then share target. On **dash, btc, bch and dgb** the block arm dispatched the won block and `return`ed, so the share arm — the *only* place the sharechain mint / create-share seam is invoked — was never reached.

Because `block_target <= share_target`, a solve that clears the block target clears the share target **by definition**. It is the highest-work share the node will ever produce, and it was discarded on every block won.

**This is not block loss.** The block submits correctly on all four coins (re-verified: every won-block KAT still passes). But "share-weight loss" undersells it — there are three consequences, and the third is the serious one.

### 1. Our own PPLNS weight

Forfeited, in the very window the block pays out from.

### 2. Peers never see our best work

We never advertise the strongest share on the chain, weakening our position in the sharechain race.

### 3. ★ Our blocks are INVISIBLE to every legacy p2pool node on the network

This surfaced after the PR was opened and is the reason it was escalated. The hotel won **two blocks today** on the live sharechain — `2511241` (64 confirmations) and `2511303`. Both accepted by dashd, both paid out correctly. **No legacy p2pool-dash node registered either one — including our own oracle**, whose `/recent_blocks` still shows a block from 21 hours earlier as the pool's most recent.

p2pool nodes do not learn about pool blocks from a block announcement. They detect them **through the sharechain**, by watching for a SHARE that satisfies `pow_hash <= header['bits'].target`. Verified against canonical p2pool-dash source (not taken on trust — read locally at `/home/ubuntu/p2pool-dash`):

`p2pool/node.py:145-147` — the detection watcher:

```python
@self.node.tracker.verified.added.watch
def _(share):
    if not (share.pow_hash <= share.header['bits'].target):
        return
    ...
    spread()
```

The same predicate is the block identity everywhere else in p2pool:

- `p2pool/data.py:436` — `if self.pow_hash <= self.header['bits'].target: return -1, 'block solution'`
- `p2pool/data.py:802` — `is_block = share.pow_hash <= share.header['bits'].target`

and `web.py:1663 get_recent_blocks()` states in its own docstring that it returns blocks *"from both persistent storage and sharechain ... merged with **current sharechain data**"*.

So the causal chain is: **block-winning share never minted → never propagates → no peer ever sees a share meeting the block target → `verified.added` watcher never fires → the block is never recorded by any p2pool node, ours included.** We are ~87% of the pool's hashrate and our blocks do not appear to exist. That lands immediately before we approach a peer operator about migrating to c2pool.

**Ruled out en route** (recorded so nobody chases it): `bestblock` is *not* the mechanism. `p2p.py:803-804`'s handler is a bare `print 'handle_bestblock', header` — verified — and it is sent on every coin-tip change with no pool attribution.

This makes the fix a **visibility and reputation** fix as much as a reward fix.

LTC is unaffected: the create-share hook runs before and independently of the block check.

### Citation check (issue line numbers verified against `3548684`)

| claim | verdict |
|---|---|
| `src/impl/dash/stratum/work_source.cpp:954` | exact — block-arm `return nlohmann::json(true);` |
| `src/impl/btc/stratum/work_source.cpp:925` | exact |
| `src/impl/bch/stratum/work_source.cpp:738` | exact |
| `src/impl/dgb/stratum/work_source.cpp:613` | exact |
| `src/core/web_server.cpp:7463` (LTC `m_create_share_fn`) | **off by one** — 7463 is the `if (!params.payout_script.empty() && params.bits != 0)` guard; the call `m_create_share_fn(params);` is 7464. (Its `if (m_create_share_fn)` outer guard is 7251.) Substance unchanged: LTC creates the share, then falls through to the block-submission block below. |

## The fix

Each coin's mint invocation is lifted **verbatim** out of the share arm into a local helper (`mint_solved_share` / `write_solved_share`) and invoked from **both** arms.

### Ordering — block submit stays absolutely first

LTC mints first. Here the helper runs on the won-block arm **strictly after** the block has been dispatched (on dash, also after the payee-guard verdict and the recent-blocks dashboard record). The invariant is *"never refuse a block the daemon would ACCEPT — a bad block is rejected free, a false refusal is guaranteed loss"*, so nothing may be inserted ahead of the submit: the mint walks the tracker, takes locks, can decline, and can throw. Running it downstream reaches the same end state as LTC with none of that exposure.

The helper also catches exceptions locally on every coin (dgb's `ShareAccept` previously had no `try`/`catch`), so a mint fault can never cost the block nor turn a won block into a stratum reject. There is a KAT per coin for exactly that.

### Per-coin

| coin | seam | change |
|---|---|---|
| **dash** | `mint_share_fn_` (bound `main_dash.cpp:1854`) | LIVE. The hotel's 9 won blocks each forfeited their own share; from here they mint. Runs **even after a payee-guard reject** — see below. |
| **btc** | `create_share_fn_` (bound `main_btc.cpp:1208`) | LIVE. |
| **bch** | `create_share_fn_` (bound `pool_entrypoint.hpp:423`) | LIVE. |
| **dgb** | `try_mint_share` | **PRE-WIRING ONLY (#884)** — see below. |

**dash + payee guard.** The mint runs even when the submit-time payee guard refused to dispatch. The guard is a *coin-side* verdict about a stale masternode payee; the sharechain is a different system with its own acceptance rule (the mint's X11-identity gate plus `pow <= own committed target`), and the share arm has never consulted the guard. The existing reject branch already says as much in its comment — *"the share still counts for the miner; the stale template was ours"*. Withholding here would be a second, self-inflicted loss on top of the forfeited block.

**dgb is pre-wiring.** Per #884 `main_dgb.cpp` never calls `set_mint_share_fn`, so dgb cannot mint **any** local share today — not this one, and not an ordinary `ShareAccept` either. The won-block arm now *reaches* the seam so dgb inherits this fix the moment #884 binds it. The dgb KAT is named `MiningSubmitWonBlockAlsoReachesMintSeam`, its comment carries an explicit SCOPE HONESTY paragraph, and it asserts only seam reachability — it does not claim a mint that cannot happen.

### No consensus change

Share construction, target derivation and serialisation live entirely inside the untouched mint / `create_local_share` seams. This changes only **whether** the seam is invoked, never what it builds. `git diff` touches no share struct, no target math, no serializer.

## ★ Caller-side lock trace

The #878/#881 hazard is: a caller holding the exclusive tracker lock across a call into a self-locking callee makes the callee permanently dead code, and a test that binds its own callback still passes. Traced explicitly:

**Caller side — zero locks held.**
`asio async_read_until` handler → `StratumSession::process_message` (`stratum_server.cpp:504`) → `handle_submit` (`:1036`) → `mining_interface_->mining_submit(...)` (`:1332` pool-share arm, `:1347` pseudoshare-is-a-block arm). Neither `process_message` nor `handle_submit` acquires any mutex; `handle_submit` deliberately *copies* the job snapshot rather than holding a reference, and holds no lock at either call site. **No tracker lock, no session lock, enters `mining_submit`.**

**Inside `mining_submit` — zero locks held across the seam, on both arms.**
Every mutex is scoped and released before the callback is invoked: dash copies `mint_share_fn_` under `mint_share_mutex_` then releases; btc/bch copy `create_share_fn_` under `callback_mutex_` then release; dgb's `try_mint_share` does the same. `workers_mutex_` is only taken *after*, in the `bump` lambda. `found_block_mutex_` (dash) is scoped and released before its callback too.

**The decisive point: the block arm and the share arm are in the SAME function, at the SAME lock depth (zero), with the SAME caller.** There is no lock-state difference between them, so the seam I call from the block arm is exactly as reachable as the one that already runs on the share arm.

**Callee side (dash, the live one).** `main_dash.cpp:1854` takes `node_ptr->read_tracker()` — a *try* `shared_lock` — inside a scoped block, **releases it**, and only then calls `add_local_share`, which takes `unique_lock(m_tracker_mutex, try_to_lock)`. No self-deadlock, and no held-lock-blocks-callee pattern. On the stratum IO thread both try-locks can genuinely succeed. btc/bch do the same (`unique_lock(tracker_mutex(), try_to_lock)`, released before `broadcast_share` / `notify_local_share`).

**How this was proved, not assumed.**
1. *Static:* the trace above — every mutex on the path is scoped, and both arms sit at lock depth zero in one function under one caller.
2. *Dynamic:* the KATs bind a real callback and assert **it ran** (`mint_called` / `share_written`), not merely that the code compiles. Deleting the one-line call reddens them (negative control below).
3. *Ordering:* each coin has an **ordering witness** — the bound callback records `submit_called` at the moment it runs and the test asserts it was already `true`. If anyone later reorders the mint ahead of the submit, that assertion goes red.
4. *Production corroboration:* the identical seam, from the identical function, at the identical lock depth, is what mints DASH shares on the live hotel sharechain today (share arm). That is the part a test cannot prove and reading can.

**Honest limit:** the KATs bind their own callback, so they prove work-source-level reachability, not that `main_dash.cpp`'s real lambda succeeds under production contention. That last step rests on the production share arm already working through the same lock topology. See "Not verified" below.

## Test evidence

Folded into **existing allowlisted targets** — never a new `add_executable` (which silently reports "Not Run"; PR #868 pattern). `tools/ci/check_test_target_allowlist.py` → `CI drift-guard OK: 120 per-coin test target(s) ... all present`. `.github/workflows/` untouched.

| target | file | result |
|---|---|---|
| `test_dash_stratum_work_source` | `test/test_dash_stratum_work_source.cpp` (+2 tests) | **46/46 PASSED** |
| `dgb_work_source_test` | `src/impl/dgb/test/work_source_test.cpp` (+2 tests) | **33/33 PASSED** |
| `btc_share_test` | new `src/impl/btc/test/won_block_mints_share_test.cpp` folded in (+`btc_stratum` link) | **116/116 PASSED** |
| `bch_g2_block_assembly_roundtrip_test` | `g2_block_assembly_roundtrip_test.cpp` (+3 checks, +`bch_stratum`/`bch_coin` link) | **all checks passed** |

Build: `Release`, gcc 13.3, `-DCOIN_DGB=ON -DCOIN_BCH=ON`. All four targets built with **zero errors**.

Representative live output (dash):

```
[DASH-STRATUM-BLOCK] *** BLOCK FOUND *** user=XfixtureMinerAddress height~=424242 pow_hash=1ce12ab821404976
[DASH-STRATUM-PAYEE-GUARD] coinbase pays all mandated payee scripts ... submitting
[DASH-STRATUM-BLOCK-SHARE] ACCEPTED + MINTED user=XfixtureMinerAddress share_hash=5555555555555555
[       OK ] DashStratumWorkSource.MiningSubmitWonBlockAlsoMintsTheShare
```

and the throw case — block already out, exception contained, miner still gets `true`:

```
[DASH-STRATUM-BLOCK-SHARE] mint_share_fn threw: mint blew up -- share not minted
[       OK ] DashStratumWorkSource.MiningSubmitWonBlockDispatchesBeforeAndDespiteAThrowingMint
```

### Negative control

Deleting **only** the one-line call from each block arm (`// NEGATIVE CONTROL: fix removed`), rebuilding, re-running:

```
dash: [  FAILED  ] MiningSubmitWonBlockAlsoMintsTheShare              (mint_called = false)
      [  FAILED  ] MiningSubmitWonBlockDispatchesBeforeAndDespiteAThrowingMint
dgb:  [  FAILED  ] MiningSubmitWonBlockAlsoReachesMintSeam            (minted = false)
btc:  [  FAILED  ] BtcWonBlockMintsShare.WonBlockAlsoWritesTheShare   (share_written = false)
bch:  FAIL: share_written @ 440 / saw_block_already_submitted @ 441 /
      seen_header_size == 80u @ 442  -> 3 CHECK(s) failed
```

Restored, rebuilt, re-ran → all four green again (numbers in the table above). Note the `SurvivesAThrowing*` KATs stay green under the negative control by design — they guard the reward invariant, not the fix; the fix-detectors are the ones listed here.

## Not verified / follow-ups

- **#889 is NOT closed by this PR.** `add_local_share` uses `try_to_lock` and *declines* if the compute thread is mid-think. For an ordinary share that is fine ("the next solve mints"); for a **block-winning** share there is no next solve on that job, so a busy tracker still forfeits it — and, per consequence 3 above, that also means the block stays invisible to p2pool peers in exactly that window. Making that one case blocking is a threading change beyond this PR's scope. **Tracked separately as #889 — do not read #888 as closing it.**
- **Production-contention behaviour of the real bound lambdas** cannot be exercised from a unit test (the KATs bind their own callback). That last step rests on the production share arm already working through the same lock topology.
- **bch** `create_share_fn_` binding lives in `pool_entrypoint.hpp`, not `main_bch.cpp`; I confirmed it is bound but did not soak the bch lane.
- **dgb** end-to-end minting remains impossible until #884.
- No live-node soak was run; this is a unit-level change verified by KAT plus negative control.

## CI: `Linux x86_64 (AsAN+UBSan)` red — diagnosed, NOT caused by this PR

The failure is **not** a crash and **not** a static-init-order abort. Full error:

```
CMake Error at /usr/share/cmake-3.28/Modules/GoogleTestAddTests.cmake:112 (message):
  Error running test executable.
    Path:   '.../build_asan/test/test_utxo'
    Result: Process terminated due to timeout
    Output:
```

`test_utxo` — a file this PR does not touch. `Result: Process terminated due to timeout` with **empty output** means `gtest_discover_tests(... DISCOVERY_MODE PRE_TEST)` ran `test_utxo --gtest_list_tests` and hit CMake's default 5 s `DISCOVERY_TIMEOUT`.

**Timing from the log proves no test had started yet:**

```
12:04:56.235  Test project .../build_asan     <- ctest begins
12:05:02.410  CMake Error ... test_utxo       <- 6.18 s later, zero tests run
```

PRE_TEST discovery happens during ctest **startup**. Nothing this PR adds was running.

### Reproduced locally under the identical flags

Same `-fsanitize=address,undefined -fno-sanitize=vptr -fno-omit-frame-pointer -fno-sanitize-recover=undefined -g`, same `ASAN_OPTIONS=detect_leaks=0:check_initialization_order=1:strict_string_checks=1:abort_on_error=0`, same `UBSAN_OPTIONS`:

```
run1: 0.19 s   (cold, first run after link)
run2: 0.06 s
run3: 0.05 s
run4: 0.06 s
exit = 0
```

and the binary itself is fully healthy — `14 tests from UTXOTest ... [ PASSED ] 14 tests`, exit 0, no ASAN or UBSAN diagnostic. **0.06 s of work against a 5 s budget.** Non-sanitized discovery is 0.00 s.

So the binary does not crash, does not abort on `check_initialization_order`, and does not need anywhere near 5 s. The CI process was starved of CPU, not broken.

### Proof the changed files cannot reach the failing binary

`test_utxo` links `core ltc ltc_coin pool c2pool_payout c2pool_merged_mining c2pool_hashrate c2pool_storage`. Mechanically, from its generated `link.txt`:

- `work_source` objects in the link line: **0**
- `dash_stratum` / `btc_stratum` / `bch_stratum` / `dgb_stratum`: **absent**
- the only `stratum` hit is `core/stratum_server.cpp.o`, which this PR does not modify

Every file in this PR's diff is either a coin `stratum/work_source.cpp` (in a lib `test_utxo` does not link), a per-coin test source, or a per-coin test `CMakeLists.txt`. None is compiled into `test_utxo`.

Additional corroboration: the sibling **`COIN_BCH Linux x86_64 (AsAN+UBSan)` job passed on this same PR**, and the same ASAN job is green on master runs `30191188565`, `30190238757` and on PR run `30197898389` — none of which contain this error string.

### The `check_initialization_order=1` hypothesis, ruled out

A plausible reading of "error running test executable" under ASAN is a static-init-order abort. It is not that, on four independent grounds:

1. The reported result is `Process terminated due to timeout`. An ASAN abort produces a diagnostic; **`Output:` was empty**. Confirmed at source: `GoogleTest.cmake:443` defaults `DISCOVERY_TIMEOUT` to 5, and `GoogleTestAddTests.cmake:105-112` raises `FATAL_ERROR` for any non-zero `execute_process(... TIMEOUT ...)` result — the timeout path.
2. This PR introduces **no new global or static with a dynamic initializer**. The only static initializer in the new TU is googletest's own per-`TEST` `test_info_` registration — `nm` shows exactly one `_GLOBAL__sub_I__ZN61BtcWonBlockMintsShare_...test_info_E`, the identical pattern to the pre-existing `share_test.cpp.o` (control: also exactly one). No ordering was changed.
3. That TU is not linked into `test_utxo` at all (`grep -c won_block_mints_share` on its link line: **0**).
4. `test_utxo` runs locally **with `check_initialization_order=1` active**, enumerating in 0.06 s and passing 14/14, exit 0, with no ASAN or UBSan output.

### Conclusion

Pre-existing latent CI-infra flake: a 5 s wall-clock `DISCOVERY_TIMEOUT` on an oversubscribed self-hosted host. Confirmed live while the re-run sat queued — `gh api .../actions/runners` shows **eight runners on VM905 (`c2pool-linux-905` … `-905-8`) all `busy=true` simultaneously**, each building `-j8`, with a 12-run queue and zero in progress. The failing job ran on `c2pool-linux-905-7`. **Not bundled into this PR** — filed separately as #898. The job is `continue-on-error: true` (informational, not gating) by design. Re-run of the identical commit is the environmental control.

Recommended separate fix, which does **not** weaken discovery in the #895 sense (the binary is still executed, and a genuine failure still hard-errors — it only stops a *slow but successful* enumeration being misread as a crash): add an explicit `DISCOVERY_TIMEOUT` to the three `gtest_discover_tests(... PRE_TEST)` calls in `test/CMakeLists.txt` (`test_dgb_subsidy`, `test_dgb_coinbase_value`, `test_utxo`).

## Scope

Untouched by design: `src/core/socket.cpp`, `src/core/web_server.cpp`, `src/impl/*/protocol_legacy.cpp`, `send_shares` / `broadcast_share` bodies, `.github/workflows/`.

### DIFF
--- a/src/impl/bch/stratum/work_source.cpp
+++ b/src/impl/bch/stratum/work_source.cpp
@@ -690,6 +690,52 @@ nlohmann::json BCHWorkSource::mining_submit(
 
     auto pow_hex_short = pow_hash.GetHex().substr(0, 16);
 
+    // -- Sharechain write (#887) --
+    // Dispatch to create_share_fn_ (v36 share add). SHARED by both accept
+    // classes: block_target <= share_target, so a solve that meets the block
+    // target meets the share target too -- it is the highest-work share this
+    // node will ever produce, and the sharechain and the coin blockchain are
+    // independent systems. Before #887 the won-block arm returned without ever
+    // reaching this seam, forfeiting our own PPLNS weight in the very window
+    // the block pays out from and denying peers our best work. LTC has always
+    // written on both classes (web_server.cpp).
+    //
+    // ORDERING -- reward invariant: on the won-block arm this runs AFTER the
+    // block has already been dispatched, never before. create_share_fn_ takes
+    // the exclusive tracker lock, can decline, and can throw; the block submit
+    // must never sit behind any of that.
+    //
+    // CONSENSUS: unchanged. Share construction, target derivation and
+    // serialisation live entirely inside the (untouched) create_local_share
+    // seam -- this changes only WHETHER it is invoked, not what it builds.
+    auto write_solved_share = [&](bool won_block) {
+        const char* tag = won_block ? "[BCH-STRATUM-BLOCK-SHARE]"
+                                    : "[BCH-STRATUM-SHARE]";
+        CreateShareFn create_fn;
+        { std::lock_guard<std::mutex> lk(callback_mutex_); create_fn = create_share_fn_; }
+
+        uint256 share_hash;
+        if (create_fn) {
+            auto payout_script = core::address_to_script(username);
+            try { share_hash = create_fn(coinbase, header, *job, payout_script); }
+            catch (const std::exception& e) {
+                LOG_WARNING << tag << " create_share_fn threw: " << e.what()
+                            << " -- share not added";
+            }
+        }
+
+        if (!share_hash.IsNull())
+            LOG_INFO << tag << " ACCEPTED + ADDED user=" << username
+                     << " share_hash=" << share_hash.GetHex().substr(0, 16)
+                     << " pow_hash=" << pow_hex_short << " job=" << job_id;
+        else if (create_fn)
+            LOG_INFO << tag << " accepted (deferred) user=" << username
+                     << " pow_hash=" << pow_hex_short << " job=" << job_id;
+        else
+            LOG_INFO << tag << " accepted (no-tracker) user=" << username
+                     << " pow_hash=" << pow_hex_short << " job=" << job_id;
+    };
+
     // -- Classify --
     if (!(pow_hash > block_target)) {
         // BLOCK FOUND.
@@ -731,6 +777,11 @@ nlohmann::json BCHWorkSource::mining_submit(
                       << height << " not broadcast -- lost subsidy!";
         }
 
+        // #887: the won block is a SHARE too. The dispatch above has already
+        // happened, so nothing here can delay, gate or endanger the block
+        // submit -- the sharechain write is strictly downstream of it.
+        write_solved_share(/*won_block=*/true);
+
         {
             std::lock_guard<std::mutex> lk(workers_mutex_);
             for (auto& [sid, w] : workers_) { (void)sid; if (w.username == username) { w.accepted++; break; } }
@@ -740,29 +791,7 @@ nlohmann::json BCHWorkSource::mining_submit(
 
     if (!(pow_hash > share_target)) {
         // Share meets sharechain target -> create_share_fn_ (v36 share add).
-        CreateShareFn create_fn;
-        { std::lock_guard<std::mutex> lk(callback_mutex_); create_fn = create_share_fn_; }
-
-        uint256 share_hash;
-        if (create_fn) {
-            auto payout_script = core::address_to_script(username);
-            try { share_hash = create_fn(coinbase, header, *job, payout_script); }
-            catch (const std::exception& e) {
-                LOG_WARNING << "[BCH-STRATUM-SHARE] create_share_fn threw: " << e.what()
-                            << " -- share not added";
-            }
-        }
-
-        if (!share_hash.IsNull())
-            LOG_INFO << "[BCH-STRATUM-SHARE] ACCEPTED + ADDED user=" << username
-                     << " share_hash=" << share_hash.GetHex().substr(0, 16)
-                     << " pow_hash=" << pow_hex_short << " job=" << job_id;
-        else if (create_fn)
-            LOG_INFO << "[BCH-STRATUM-SHARE] accepted (deferred) user=" << username
-                     << " pow_hash=" << pow_hex_short << " job=" << job_id;
-        else
-            LOG_INFO << "[BCH-STRATUM-SHARE] accepted (no-tracker) user=" << username
-                     << " pow_hash=" << pow_hex_short << " job=" << job_id;
+        write_solved_share(/*won_block=*/false);
 
         {
             std::lock_guard<std::mutex> lk(workers_mutex_);
--- a/src/impl/bch/test/CMakeLists.txt
+++ b/src/impl/bch/test/CMakeLists.txt
@@ -212,6 +212,14 @@ if(BUILD_TESTING)
     # Needs coin/transaction.cpp for the MutableTransaction ctors, same as
     # block_connector_test; link set and header-only-over-coin/*.hpp posture are
     # otherwise identical, so per-coin isolation stays clean.
+    #
+    # #887 fold-in: the same file now also drives BCHWorkSource::mining_submit
+    # to pin that a WON BLOCK is written to the sharechain as well (block
+    # dispatch first, share write strictly after). That needs bch_stratum
+    # (work_source.cpp) and its bch_coin dependency. FOLDED into this EXISTING
+    # allowlisted target rather than a new add_executable -- a new one would
+    # silently report "Not Run" (see merged PR #868). Still bch-only, so the
+    # per-coin isolation invariant holds.
     add_executable(bch_g2_block_assembly_roundtrip_test
         g2_block_assembly_roundtrip_test.cpp
         ${CMAKE_SOURCE_DIR}/src/impl/bch/coin/transaction.cpp)
@@ -219,6 +227,7 @@ if(BUILD_TESTING)
     target_link_libraries(bch_g2_block_assembly_roundtrip_test PRIVATE
         core pool sharechain
         c2pool_payout c2pool_merged_mining c2pool_hashrate c2pool_storage c2pool_difficulty
+        bch_stratum bch_coin
         btclibs)
     add_test(NAME bch_g2_block_assembly_roundtrip_test COMMAND bch_g2_block_assembly_roundtrip_test)
 endif()
--- a/src/impl/bch/test/g2_block_assembly_roundtrip_test.cpp
+++ b/src/impl/bch/test/g2_block_assembly_roundtrip_test.cpp
@@ -49,13 +49,20 @@
 #include <string>
 #include <vector>
 
+#include <memory>
+#include <stdexcept>
+
 #include "../coin/block.hpp"
 #include "../coin/block_assembly.hpp"
+#include "../coin/header_chain.hpp"
 #include "../coin/mempool.hpp"
 #include "../coin/merkle.hpp"
 #include "../coin/reconstruct_won_block.hpp"
 #include "../coin/transaction.hpp"
 #include "../stratum/coinbase_slot_guard.hpp"
+#include "../stratum/work_source.hpp"
+
+#include <core/stratum_types.hpp>
 
 namespace {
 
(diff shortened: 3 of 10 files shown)


frstrtr:
**INTEGRATOR REVIEW — diff read. VERDICT: APPROVE, merge on green.** Operator-escalated as high priority.

**Why this matters more than the original framing.** #887 was scoped as share-weight loss. The live consequence found today is worse: because the block-winning share is never minted, it never propagates, so **no p2pool node on the network can detect our blocks at all**. We won 2511241 (64 conf) and 2511303 today — both dashd-accepted and correctly paid out to all five pool participants — and NO legacy node registered either, including our own oracle. Verified against canonical source: `node.py:145-147` fires block detection from `tracker.verified.added` only for a share with `pow_hash <= header['bits'].target`, and `/recent_blocks` merges persistent history with **current sharechain data**. We are 87% of the pool's hashrate and invisible in its own records.

**The implementation is right, and better than what I asked for.** I specified "mirror LTC" (mint first); the PR does **block dispatch first, mint strictly after** — correct, because the mint walks the tracker, can decline, and can throw, and nothing may sit ahead of the submit. Same end state, none of the exposure. The local try/catch on every coin means a mint fault cannot turn a won block into a reject.

**The single-call contract is the detail I'd highlight**: moving `coinbase`/`branch_hashes` into the mint inputs makes it structurally impossible for both classify arms to invoke it. That's an invariant enforced by the type system rather than by comment, which is the right way to express it.

**The lock trace is what makes this trustworthy.** Caller holds zero locks; block arm and share arm sit in the same function at identical lock depth from the same caller — so the seam is provably as reachable as the one already minting in production. That distinguishes it from #873, where tests AND a negative control both passed against unreachable code.

**On the ASAN red — accepted as environmental, with proof.** `GoogleTestAddTests.cmake:112` reported `Process terminated due to timeout` with EMPTY output on `test_utxo`, i.e. CMake's default 5s `DISCOVERY_TIMEOUT`, not a crash. ctest started 12:04:56.235, errored 12:05:02.410 — 6.18s with zero tests run. Locally the same binary enumerates in 0.06s and passes clean under identical sanitizer flags. `test_utxo`'s link.txt has zero work_source objects, so this PR's files cannot reach it; the sibling COIN_BCH ASAN job passed on the same commit. Root cause is eight runners on one host each building -j8 (and Conan actually spawns 16 cc1plus). Filed separately as #898 rather than bundled — correct call.

**Explicitly NOT closed by this PR: #889.** A `try_to_lock` decline still forfeits the block-winning share, and a block has no "next solve". Worth noting the decline reproduces the full invisibility symptom, just less often — so #889 is now more than a share-weight follow-up.

Merging as soon as the re-run and attribution-gate land.

PR #890, opened 2026-07-26
ci: build and ship DASH with the real BLS backend linked (C2POOL_DASH_BLS=ON + non-hollow assertions)

frstrtr:
## Problem

`C2POOL_DASH_BLS` is set by **no** workflow in `.github/workflows/` — zero hits before this PR. Every DASH binary CI produces and every DASH package `release.yml` ships therefore compiles `dash_bls_verify` as the fail-closed stub.

The consequence is not cosmetic. The LLMQ/BLS commitment-verification core merged as #826 is soak-proven — 130 embedded serves, 0 bad-cbtx, 49 DKG-window heights served with real BLS-verified commitments byte-matching dashd — but it can never serve a real commitment in the field, because the shipped code path is the stub. This is on the DASH daemonless path.

**Confirmed empirically, not inferred:** the published `master` DASH artefact fails the link guard added here on both legs. Today's shipped binary is BLS-dark.

## The trap this PR is really about

Flipping the flag alone would have produced a green build that still ships dark binaries. Before this PR, `-DC2POOL_DASH_BLS=ON` with dashbls **missing** only emitted `message(WARNING)`: the option stayed on, the stub compiled, the build succeeded, and the package was byte-for-byte indistinguishable from a working one until a DKG-window height silently null-served in production. Same shape as a new test target quietly reporting "Not Run".

So the flag never travels alone here. It travels with three independent assertions.

### Assertion 1 — configure-time hard failure (root `CMakeLists.txt`)

`C2POOL_DASH_BLS=ON` + library not found is now a `FATAL_ERROR` that names every path it searched, instead of a warning. `OFF` (still the default) remains the supported way to build without BLS; `-DC2POOL_DASH_BLS_OPTIONAL=ON` restores warn-and-degrade for a local probe build.

### Assertion 2 — the artefact carries the backend (`scripts/ci/dash_bls_linked_guard.sh`)

Checks the produced binary, not the build flags, so a future refactor that reintroduces a soft fallback still cannot ship quietly:

- the dashbls BLS12-381 ciphersuite ID in rodata — a string constant, so it **survives `strip`** and the check applies to the packaged tarball users download;
- the `bls::G1Element::FromBytes` / `bls::G2Element::FromBytes` entry points `bls_verify.cpp` calls (skipped, explicitly and never silently, on a stripped binary).

### Assertion 3 — the BLS-only KATs actually ran (`build.yml`)

The six BLS-crypto KATs sit behind `#ifdef C2POOL_DASH_BLS`. On a stub build they do not fail — they **cease to exist**, and a bare `ctest` reports green over an empty set. The new job requires each one by name to appear as a passed line.

## What the runners need

`scripts/build_dashbls.sh` builds dash's own `src/dashbls` subtree at `v23.1.7`, i.e. byte-identical source to what dashd links. It is not on ConanCenter, so it is consumed as a prebuilt system library like secp256k1.

`.github/actions/provision-dashbls` (new composite action) handles it:

- parses the pinned tag **out of `scripts/build_dashbls.sh`** rather than duplicating it, so a re-pin cannot leave CI on the old library;
- installs `autoconf automake libtool libgmp-dev` under the same `flock` guard `release.yml` already uses for its apt step, then **asserts** `libgmp.so` (the `-dev` symlink `find_library(gmp)` needs) is present, so a host missing it says exactly that instead of surfacing as a confusing "dashbls not found";
- builds into `~/.c2pool-deps/dashbls` only when that prefix is cold or pinned to another tag — the heavy autotools+relic+mimalloc build happens **once per runner per tag**, not per job;
- restores the prefix from the Actions cache on github-hosted runners (fork PRs), which are ephemeral. Self-hosted keeps it on disk, which is faster and sidesteps the cache-service restore timeouts seen on the VM905 pool;
- fails the job if it cannot produce a usable prefix.

No manual host provisioning is required: the first `linux-dash-bls` run on `c2pool-build` self-provisions and every later run reuses the prefix.

## Scope — other lanes untouched

- `build.yml` is **purely additive**: `159` insertions, `0` deletions, one new job (`linux-dash-bls`). No existing job's flags, targets or steps changed.
- `release.yml`: the per-coin extra flags in the Configure step moved from an inline expression to a `case`, emitting byte-identical arguments for `btc`/`ltc`/`bch` (none) and `-DAUX_DOGE=ON` for `dgb` exactly as before. Every other new step is gated on `matrix.coin == 'dash'`.
- Windows / macOS release cells and `coin-matrix.yml` are not modified at all.
- `dash_bls_verify` is linked only by `c2pool-dash`, so no BLS symbol enters the shared core the Windows LTC launcher builds.

## Verification (Ubuntu 24.04, GCC 13, Conan)

| Check | Result |
|---|---|
| configure, library present | `dashbls found (E1 Phase-L BLS verify ENABLED)`, exit 0 |
| configure, `-DC2POOL_DASH_BLS=ON` and library absent | `FATAL_ERROR`, exit 1 |
| escape hatch `-DC2POOL_DASH_BLS_OPTIONAL=ON` | warning only, configure completes |
| link guard on BLS-linked `c2pool-dash` | PASS (both legs) |
| link guard on the same binary after `strip` | PASS (rodata leg; symbol leg reports SKIP) |
| link guard on a same-tree `C2POOL_DASH_BLS`-off build | **FAIL, exit 1** |
| link guard on the published `master` DASH artefact | **FAIL, exit 1** |
| the six BLS-only KATs | all run, all pass |
| run-assertion against a log with those KATs removed | reports MISS, non-zero |
| `scripts/build_dashbls.sh` from a cold prefix | exit 0 |

One finding worth recording: the first draft of the link guard used `grep -q` inside a pipeline under `set -o pipefail`. `grep -q` exits at the first match and SIGPIPEs its producer, so the pipeline reported failure *because the match succeeded* — the guard called a known-good BLS-linked binary dark. Caught by running it against a binary already proven good. Both the guard and the KAT run-assertion now use pure-shell `case` matching, with the reason written down where it would be reintroduced.

## Known gap (not addressed here)

Only the **Linux** DASH release cell gets the backend. The macOS and Windows DASH packages still ship BLS-dark: `src/dashbls` at this tag only supports the autotools path, which has no MSVC route at all, and the macOS runners would need their own validated prefix. Deliberately left out of scope rather than half-enabled — Linux is where the DASH daemonless path runs.

### DIFF
--- a/.github/actions/provision-dashbls/action.yml
+++ b/.github/actions/provision-dashbls/action.yml
@@ -1,0 +1,138 @@
+name: Provision dashbls
+description: >
+  Make the dashbls (BLS12-381) backend available to a DASH build, so
+  -DC2POOL_DASH_BLS=ON links the REAL verifier instead of failing to configure.
+
+  Builds dash's own src/dashbls subtree at the pinned release tag via
+  scripts/build_dashbls.sh (the single source of truth for the tag - this action
+  parses DASH_REF out of that script rather than duplicating it) and installs it
+  into a per-host prefix that is reused across runs. The build is heavy
+  (autotools + relic + mimalloc), so it happens ONCE per runner per pinned tag:
+  a prefix whose .dash-ref marker already matches is used as-is. On
+  github-hosted runners (fork PRs, ephemeral) the prefix is additionally
+  restored from the Actions cache.
+
+  Exports DASHBLS_ROOT to $GITHUB_ENV for the subsequent cmake configure, and
+  FAILS the job if the library or headers are not present when it finishes -
+  never leaves a job believing it has BLS when it does not.
+
+runs:
+  using: composite
+  steps:
+    - name: Resolve dashbls prefix + pinned tag
+      shell: bash
+      run: |
+        set -euo pipefail
+        # Fixed prefix (no input): the github-hosted cache path below must be a
+        # literal, and a configurable prefix that silently misses the cache is
+        # worse than none.
+        prefix="$HOME/.c2pool-deps/dashbls"
+        # Single source of truth for the pin: scripts/build_dashbls.sh. Parsing
+        # it here means a re-pin of the tag cannot silently leave CI on the old
+        # library (the prefix marker below is keyed on this value).
+        ref="$(sed -n 's/^DASH_REF="\${DASH_REF:-\(.*\)}"$/\1/p' scripts/build_dashbls.sh)"
+        if [ -z "$ref" ]; then
+          echo "::error::could not parse DASH_REF from scripts/build_dashbls.sh"
+          exit 1
+        fi
+        echo "dashbls prefix=$prefix pinned dash ref=$ref"
+        echo "DASHBLS_PREFIX=$prefix" >> "$GITHUB_ENV"
+        echo "DASHBLS_PIN=$ref" >> "$GITHUB_ENV"
+
+    # Ephemeral github-hosted runners only (fork PRs). Self-hosted runners keep
+    # the prefix on disk, which is both faster and immune to the cache-service
+    # restore timeouts seen on the VM905 pool.
+    - name: Restore dashbls cache (github-hosted only)
+      if: runner.environment == 'github-hosted'
+      uses: actions/cache@v5
+      with:
+        path: ~/.c2pool-deps/dashbls
+        key: dashbls-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('scripts/build_dashbls.sh') }}
+
+    - name: Install dashbls build dependencies
+      shell: bash
+      run: |
+        set -uo pipefail
+        # relic needs gmp, and the vendored subtree only supports the autotools
+        # path (see the note in scripts/build_dashbls.sh). libgmp-dev is also a
+        # LINK-time requirement of the c2pool find (find_library(gmp)), so it is
+        # ensured even when the prefix is already warm.
+        #
+        # flock: several coin cells run concurrently on the one self-hosted host
+        # and race on the apt lock (the same guard release.yml uses for its own
+        # apt step).
+        #
+        # Non-fatal on purpose: a transient apt hiccup on a host that ALREADY
+        # has these packages must not red the job. What is actually required is
+        # verified right after, and a genuinely missing gmp fails there.
+        exec {lockfd}>/tmp/c2pool-apt.lock
+        flock "$lockfd"
+        sudo apt-get update -qq || true
+        sudo apt-get install -y --no-install-recommends \
+          autoconf automake libtool libgmp-dev || true
+
+    - name: Assert libgmp development files are present
+      shell: bash
+      run: |
+        set -euo pipefail
+        # find_library(DASHBLS_GMP_LIB NAMES gmp) in the root CMakeLists needs
+        # the libgmp.so DEVELOPMENT symlink, not just the versioned runtime
+        # object -- i.e. libgmp-dev. Checked explicitly so a host missing it
+        # says so here instead of surfacing as a confusing "dashbls not found".
+        entries="$(ldconfig -p || true)"
+        case "$entries" in
+          *"libgmp.so ("*) echo "  ok   libgmp runtime present" ;;
+          *) echo "::error::libgmp is not installed on this runner (need libgmp-dev)"; exit 1 ;;
+        esac
+        found=""
+        for d in /usr/lib/x86_64-linux-gnu /usr/lib/aarch64-linux-gnu /usr/lib /usr/local/lib; do
+          if [ -e "$d/libgmp.so" ]; then found="$d/libgmp.so"; break; fi
+        done
+        if [ -z "$found" ]; then
+          echo "::error::libgmp.so development symlink not found (install libgmp-dev on this runner)"
+          exit 1
+        fi
+        echo "  ok   $found"
+
+    - name: Build dashbls if the prefix is cold or pinned to another tag
+      shell: bash
+      run: |
+        set -euo pipefail
+        marker="$DASHBLS_PREFIX/.dash-ref"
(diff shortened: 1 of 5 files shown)


frstrtr:
## Independent re-verification (Ubuntu 24.04, GCC 13, Conan 2, self-built dashbls)

Every row of the verification table in the description was re-run from scratch in a clean tree, plus two new proofs that were not previously established. Transcript:

### Configure-time assertion (Leg 1)

| Configure | Result |
|---|---|
| `-DC2POOL_DASH_BLS=ON -DDASHBLS_ROOT=/nonexistent-prefix` | **exit 1**, `CMake Error at CMakeLists.txt:179`, naming each searched path (`DASHBLS_INCLUDE_DIR-NOTFOUND`, `DASHBLS_LIB-NOTFOUND`) |
| `-DC2POOL_DASH_BLS=ON -DDASHBLS_ROOT=<real prefix>` | exit 0, `dashbls found (E1 Phase-L BLS verify ENABLED)` |
| `-DC2POOL_DASH_BLS=ON -DC2POOL_DASH_BLS_OPTIONAL=ON` + library absent | exit 0 with exactly one `CMake Warning` — escape hatch behaves as documented |

### Artefact assertion (Leg 2)

| Binary | Guard |
|---|---|
| freshly built `c2pool-dash` (BLS on) | **PASS** — rodata leg + both `bls::G{1,2}Element::FromBytes` symbols |
| same binary after `strip` | **PASS** — rodata leg holds, symbol leg reports `skip` explicitly |
| a known BLS-off `c2pool-dash` built from this tree lineage | **FAIL, exit 1** on all three legs |
| **the published `v0.2.4` Linux DASH release tarball** | **FAIL, exit 1** — `no BLS12-381 ciphersuite ID ... this binary is BLS-DARK` |

The last row is the premise of this PR confirmed against the artefact users actually download today, not against a reconstruction. That tarball also ships no `libgmp` in `lib/`, consistent with a BLS-off build — which is what the new packaging assertion checks for on the BLS-on path.

### KAT run-assertion (Leg 3) — and why it is load-bearing, not belt-and-braces

The description said a bare `ctest` would "report green over an empty set". That is now demonstrated concretely, and the mechanism is worse than it sounded:

`test/CMakeLists.txt:524` uses `gtest_add_tests(test_dash_bls_verify "" AUTO)`, which discovers cases by **parsing the source at configure time**. It does not evaluate the preprocessor. So in a stub configure the BLS-only cases are still registered — verified directly in a stub-configured build tree:

```
/tmp/probe_opt/test/CTestTestfile.cmake:1168:
add_test([=[DashBlsVerify.RealQuorumSigAcceptsAndTamperRejects]=] ".../test_dash_bls_verify" \
  "--gtest_filter=DashBlsVerify.RealQuorumSigAcceptsAndTamperRejects")
```

…while the case itself is inside `#ifdef C2POOL_DASH_BLS` (`test_dash_bls_verify.cpp:128–274`). And a gtest binary whose filter matches nothing exits **0**:

```
$ ./test/test_dash_bls_verify --gtest_filter='DashBlsVerify.ThisCaseDoesNotExist'
[==========] 0 tests from 0 test suites ran. (0 ms total)
[  PASSED  ] 0 tests.
exit=0
```

So on a BLS-dark build `ctest` does not skip the BLS KAT — it reports it as **`Passed`**. Leg 3 is the only thing standing between that and a green light.

Both directions of Leg 3 exercised:

- against the real log — all six cases `ok`, exit 0;
- against the same log with those six lines removed (the stub shape) — six `MISS` lines, **exit 1**.

Full run of the three targets under `ctest`: `100% tests passed, 0 tests failed out of 21`.

### Runner provisioning

`scripts/build_dashbls.sh` was run from a **completely cold prefix** into the exact path the composite action uses (`~/.c2pool-deps/dashbls`): exit 0, producing `include/dashbls/bls.hpp` and `lib/libdashbls.a` — the two files the action asserts on. No separate `librelic_s.a` / `libmimalloc-secure.a` are produced, confirming the note in the root `CMakeLists.txt` that the autotools install folds relic and mimalloc into `libdashbls.a` and that those `find_library` results are correctly treated as optional. No manual host provisioning is required.

### Scope

- `master` has **zero** `C2POOL_DASH_BLS` occurrences under `.github/` — the stated premise, confirmed.
- `dash_bls_verify` is linked by exactly one executable (`src/c2pool/CMakeLists.txt:325`, `c2pool-dash`) and the three DASH tests. It is absent from `core`, `pool`, `sharechain` and from every other coin target, so no BLS symbol can reach the shared core the Windows LTC launcher builds.
- All `release.yml` hunks fall inside the `linux` job (lines 56–273). The `macos`, `macos-universal` and `windows` jobs are not modified.
- `build.yml` is purely additive — one new job, no edits to existing ones.

frstrtr:
Non-destructive rebase path (b): the real-BLS content is now carried on fresh signed branch ci-steward/dash-bls-enable as #978, rebased onto current master (inherits the .github-preserving prune fix #916). This draft (#890) stays open for history; merge proceeds via #978. Cross-link: #978.

frstrtr:
Superseded by #978 (merged 2026-07-30, a05e1e72), which landed this exact deliverable — linux-dash-bls in build.yml (-DC2POOL_DASH_BLS=ON + dashbls provision + dash_bls_linked_guard.sh + six named BLS KATs report-as-run), release.yml shipping the dash cell with the link guard on built and packaged binary, and fail-closed CMakeLists. Rebasing #890 would only reproduce merged work and re-conflict. Closing as superseded; branch retained. Verified by integrator.

frstrtr:
Cross-link: the two net-new commits on the pr-890 rebase branch (fail-closed dashbls dedup + REAL/STUB startup banner) are carried forward in #1013 as clean GPG-signed commits off current master. This PR remains superseded by #978 for the real-BLS provision itself.

Issue #883, opened 2026-07-26
test: VerifyShareThreading.Phase2HoldsLockAgainstComputePrune flakes on its own self-check under sanitizer load


frstrtr:
## Summary

`VerifyShareThreading.Phase2HoldsLockAgainstComputePrune` (`test/test_threading.cpp`) fails intermittently on loaded CI runners, on its own **self-check** rather than on the invariant it is meant to verify.

Observed on PR #872, `Linux x86_64 (AsAN+UBSan)`:

```
io_ops == 0  -- "IO body never acquired the lock — invariant not exercised"
```

## Why it is a flake, not a defect

Three independent confirmations that the failure is unrelated to the change under test:

1. PR #877, which touches a strictly **larger** surface, passes the identical `Linux x86_64 (AsAN+UBSan)` job.
2. PR #872's own `COIN_BCH Linux x86_64 (AsAN+UBSan)` job passes.
3. `test/test_threading.cpp` includes neither `core/random.hpp` nor any protocol file, so #872's edits (guard inversion in five `protocol_legacy.cpp` + `random_choice` bounds) cannot reach it.

It also passes locally on the same branch.

## Root cause

Both IO threads acquire with `std::try_to_lock` and `continue` on failure. The test then asserts that at least one acquisition happened (`io_ops > 0`) to prove the invariant was actually exercised. Under sanitizer slowdown on a contended runner, the compute thread can hold the lock for the whole test window, so every `try_lock` fails, `io_ops` stays 0, and the self-check trips — even though nothing is wrong.

The test is timing-dependent by construction, and its failure mode is indistinguishable from a real regression at a glance, which makes it actively expensive: it will keep sending people to investigate changes that are innocent.

## Suggested fix

Make the exercise deterministic rather than opportunistic — e.g. have the IO body block (or retry with a bounded deadline) so at least one acquisition is guaranteed, and keep `try_to_lock` only where the test is specifically asserting the non-blocking property. Failing that, retry the whole scenario until `io_ops > 0` with a generous cap, and only then assert the invariant.

Found while reviewing #872 (issue #869).

frstrtr:
[...]

PR #892, opened 2026-07-26
test(threading): deterministic exercise for Phase2HoldsLockAgainstComputePrune (#883 flake)

frstrtr:
Fixes the 3x flake on #872 that is holding a clean PR out of master.

## Root cause
`VerifyShareThreading.Phase2HoldsLockAgainstComputePrune` asserts `io_ops > 0` as a self-check that the invariant was exercised. Both IO bodies acquire with `std::try_to_lock` and `continue` on failure. Under sanitizer slowdown on a contended runner the compute thread can hold the exclusive lock for the entire window, so every try_lock fails, `io_ops` stays 0, and the self-check trips while the discipline under test is correct.

Environmental, cross-confirmed by integrator: #877 (larger surface) passes the identical Linux x86_64 AsAN+UBSan job; #872 own BCH AsAN job passes; test_threading.cpp includes no protocol/random header #872 touched.

## Fix
First iteration acquires BLOCKING so at least one real under-lock deref is guaranteed (io_ops>0). Remaining iterations keep `try_to_lock` so the non-blocking defer path is still exercised. The concurrent-prune UAF property under ASan/TSan is unchanged.

Draft for review, no self-merge.

### DIFF
--- a/test/test_threading.cpp
+++ b/test/test_threading.cpp
@@ -557,8 +557,18 @@ TEST(VerifyShareThreading, Phase2HoldsLockAgainstComputePrune)
     auto io_body = [&] {
         for (int iter = 0; iter < 20000 && !stop.load(std::memory_order_relaxed); ++iter) {
             // processing_shares_phase2: hold lock across the mutation body.
+            // The first iteration acquires BLOCKING so at least one real
+            // acquisition + under-lock deref is GUARANTEED to exercise the
+            // invariant even when the compute thread monopolises the lock under
+            // sanitizer slowdown on a contended runner (io_ops would otherwise
+            // stay 0 and trip the self-check with nothing actually wrong —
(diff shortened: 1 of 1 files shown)


frstrtr:
**INTEGRATOR REVIEW — APPROVE.** 11 lines, test-only, and it fixes the false alarm without weakening the assertion.

The first iteration acquires BLOCKING, guaranteeing at least one real acquisition and under-lock deref, so `io_ops` can never be 0 for environmental reasons. Later iterations keep `try_to_lock`, so the non-blocking defer path — the thing the test actually exists to cover — is still exercised. That is the right shape: it removes the flake, not the coverage.

This has cost three investigations today (it held #872 out of master twice) and its failure mode is indistinguishable from a real regression at a glance, which is the expensive part. Merging.

Issue #886, opened 2026-07-26
ltc/doge: two false-refusal paths lose valid blocks — process-lifetime prev-hash dedup + non-advisory merkle recompute


frstrtr:
## Summary

Two independent false-refusal paths on the LTC/DOGE block-submit route, both violating the invariant this codebase already learned the hard way: *"NEVER refuse a block the daemon would ACCEPT... a bad block is rejected FREE, a false refusal is guaranteed loss."* A prior payee-guard false-refusal cost ~0.89 DASH before it was caught.

`src/core/web_server.cpp:7502-7517` (pool path; the solo path at `:7100-7130` has the same shape):

```cpp
if (header_merkle != expected_merkle) {
    LOG_ERROR << "Pool block merkle_root mismatch!" ...   // <-- logs, never submits
} else {
    uint256 prev_block;
    std::memcpy(prev_block.data(), block_bytes.data() + 4, 32);
    static uint256 s_last_pool_submitted_prev;             // <-- process-lifetime
    if (prev_block != s_last_pool_submitted_prev) {
        s_last_pool_submitted_prev = prev_block;           // <-- set BEFORE submit
        submitblock(block_hex);
    }
}
```

## Bug 1 — same-prev-hash dedup permanently retires a prev-hash (HIGH)

`s_last_pool_submitted_prev` is a **process-lifetime static**, assigned **before** `submitblock` runs and never reconciled with the result.

**Failure scenario.** Miner A solves at prev=P. The static is set to P, then `submitblock` fails — stale template, transient RPC error, or a `-1` verdict. Seconds later miner B solves a **valid** block at the same prev=P. `prev_block == s_last_pool_submitted_prev`, so it is **silently never submitted**. There is no log line on that branch at all, so the loss is invisible.

Also fires after a 1-block reorg back to P: our block at the re-org'd prev is now unsubmittable for the life of the process.

Additionally it is an **unsynchronised static written from multiple stratum session threads** — a data race independent of the logic bug.

Dedup is reasonable; retiring a prev-hash forever on an *attempted* submit is not. A duplicate submission is rejected by the daemon for free.

## Bug 2 — merkle recompute refuses to submit (HIGH)

`:7502` logs `Pool block merkle_root mismatch!` and falls through **without submitting**.

BTC runs the identical check at `src/impl/btc/stratum/work_source.cpp:802-818` and treats it as **explicitly advisory — it never rejects the share**. LTC/DOGE is the outlier.

[...]

PR #894, opened 2026-07-26
#886 core(web_server): stop LTC/DOGE submit path from silently losing valid blocks

frstrtr:
Fixes two false-refusal bugs in `core::MiningInterface::mining_submit` (the LTC/DOGE submit path; BTC/DASH/DGB/BCH use the separate `IWorkSource::mining_submit` family and are unaffected).

BUG 1 (dedup): process-lifetime static keyed on prev-hash replaced with a bounded (256) thread-safe recent-submission set keyed on the BLOCK HASH, recorded ONLY on successful submit. Distinct blocks sharing a prev both submit; a failed submit stays resubmittable.

BUG 2 (merkle): self-check is now advisory — LOG_WARNING then submit anyway (daemon is authority; mirrors BTC work_source.cpp:802). Never refuse a block the daemon might accept.

Scope: src/core/web_server.cpp + .hpp only (shared file, LTC/DOGE-scoped behavior). Isolation-reviewed: no other coin submit semantics change; auto-detect fallback can only make submits safer.

Reachability (caller-side lock trace): mining_submit is dispatched per-share from stratum_server.cpp:1332/1347, NOT inside a held tracker exclusive lock — the block branch runs lock-free, so the fix is live code (structurally unlike #878/#881). No empirical [BLOCK] fired (0 LTC blocks in 30d at current hashrate) — structural trace, not live-fired.

Tests: 5 new RecentBlockDedup KATs folded into core_test (web_server_submitblock_test.cpp). core_test 98/98 green locally on Linux x86_64.

GPG-signed 16be9867 (key 50AB1379).

### DIFF
--- a/src/core/test/web_server_submitblock_test.cpp
+++ b/src/core/test/web_server_submitblock_test.cpp
@@ -5,6 +5,7 @@
 #include <cstring>
 #include <stdexcept>
 #include <string>
+#include <thread>
 #include <vector>
 
 #include <nlohmann/json.hpp>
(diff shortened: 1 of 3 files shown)


frstrtr:
**INTEGRATOR REVIEW — diff read, not just checks. VERDICT: APPROVE, merge on green.**
(Formal approve is blocked — same account authored it — so recording here.)

**BUG 2 (merkle) — correct.** `LOG_ERROR`+refuse becomes `LOG_WARNING`+submit-anyway, and critically the dedup block moved OUT of the `else`, so a mismatch can no longer suppress the submit at all. Mirrors BTC `work_source.cpp:802`, which has always treated this check as advisory. The comment states the invariant explicitly — good thing to leave for the next reader.

**BUG 1 (dedup) — correct, and I checked the success predicate specifically.** Marking on an ambiguous verdict would reintroduce the exact loss being fixed, so `!submit_res.contains("error")` needed verifying. It holds: `submitblock()` returns `{{"error", ...}}` on every failure path — daemon rejection (:102), exception (:111), short data (:8). Marking therefore happens ONLY on genuine success; rejection, exception and daemon-duplicate all stay resubmittable. That errs toward resubmitting, which is the correct direction — a redundant submit is free, a suppressed one is a lost block.

Keying on block hash rather than prev-hash is the substantive fix: many distinct valid blocks share one prev, which is precisely what made the old static drop the second one.

**Tests pin the right things.** `DistinctBlocksSamePrev_BothSubmittable` is the actual regression. `UnmarkedBlockStaysSubmittable` covers the failed-submit case that lost blocks silently. `BoundedFifoEvictsOldest` pins the 256 cap. `ConcurrentMarksAreSafeAndBounded` (8 threads) targets the unsynchronised-static data race — worth a TSan run as the comment suggests.

**The caller-side lock trace is what makes this trustworthy.** Proving `mining_submit` is not under the tracker exclusive lock is exactly what distinguishes this from #873, where tests AND a negative control both passed against unreachable code. That gate earned its keep today.

Non-blocking follow-up: the linear scan is fine now (won-block path is rare, cap 256), but if the recent-set is ever reused on a hotter path it wants a real hash set.

Confirmed both copies fixed (solo ~7164, pool ~7562) and that these were the only two — BTC/DASH/DGB work_sources were already clean.

PR #912, opened 2026-07-27
web: gate every merged-child element behind the topology class (no phantom DOGE on standalone coins)

frstrtr:
## What / why

A live paying tenant on the standalone DASH pool saw a phantom `+N/A DOGE` on the **Miners Block Value** card, sitting next to the real DASH block value — inviting the reading that a second payout stream exists. Same defect class as the merged worker-username instructions #896 fixed: the dashboard must never imply revenue that does not exist.

#896 gated only the worker-username line. The block-value sub, the merged-block ETA leg, and the node merged-payout sub were bare hardcoded-`DOGE` elements the standalone-hide path never touched.

## The one mechanism (not a second conditional)

The topology callback already shows/hides every `.merged-child-hint` on `currency_info.merged_child_symbol`, and fills `.merged-child-symbol-hint` with the symbol. This change brings the three orphaned merged elements under that **same existing class** — no new per-site conditional (that is how we got here).

- `web-static/dashboard.html:1852` block-value sub → gated + symbol from `.merged-child-symbol-hint`
- `:1853` merged-block ETA leg → gated
- `:1860` node merged-payout sub → gated + symbol from topology
- JS `#node_merged_symbol` write → sourced from `currency_info.merged_child_symbol`, not hardcoded

## Failure-direction (per review constraint)

- **Standalone coin** (no `merged_child_symbol`): every merged element hidden. ✔
- **Failed `currency_info` fetch**: falsy → treated as standalone → merged elements hidden. Safe direction. ✔
- **LTC** (has `merged_child_symbol`): DOGE elements still render. ✔ (KAT-independent; reverse case preserved.)

## KAT — fails without the fix

`test/test_dashboard_merged_gating.cpp` (static-source honesty assertion, mirrors `test_web_honesty_regression`):
1. every merged HTML element must sit inside a `.merged-child-hint` gate;
2. the standalone-hide path must exist;
3. no hardcoded merged symbol fused to `merged_block_value`.

Verified: **3/3 pass** on this branch; runs against the pre-fix source produce **3 ungated violations** (fails without the fix).

## Scope note

This lands the operator-facing gating (the phantom-DOGE bug). The related explorer-URL sweep (routing hardcoded `blockchair.com/dogecoin|litecoin/...` links through `currency_info.*_explorer_url_prefix`) needs a `merged_child_*_explorer_url_prefix` field on `currency_info` — a distinct backend touch — and follows as a stacked slice.

Display-only, non-consensus, reversible.


frstrtr:
---
### 07-27 update — #769 Not-Run trap fixed (build-wiring, not logic)

The 3 reds were one cause: `DashboardMergedGating` was a standalone `add_executable` missing from the build.yml `--target` allowlist, so CI never compiled it → ctest reported all 7 cases `(Not Run)` → exit 8, and the allowlist drift-guard failed closed on exactly that. Fixed per integrator preference (a): folded the 7 cases into the already-allowlisted `test_web_honesty_regression` target (same charter, where #896 lives), added `DASHBOARD_HTML_PATH` to that target, deleted the standalone target + file. No new CMake target, no build.yml edit.

**Non-vacuous, verified locally (built + run):**
- All 7 PASS on fixed HTML; folded honesty suite 49/49 green.
- All 7 **FAIL** on pre-fix HTML (`git merge-base HEAD master:web-static/dashboard.html`) — proven, not assumed.

**Coverage confirmed (the operator-reported phantom):**
- `NoHardcodedMergedSymbolOnBlockValue` ↔ dashboard.html:1852 `+<merged_block_value> DOGE` card — fails pre-fix.
- :1860 Node Fee card: `node_merged_symbol` is gated by containment in `node_merged_payout` (the one `.merged-child-hint` mechanism); `node_merged_payout` is in the tests gated-id set.
- :3049 `+N/A` JS fallback: `MergedNaWritesTargetGatedElements` asserts the exact `merged_block_value`/`time_to_merged_block` N/A writes exist AND their target elements are gated — couples the JS render path to the gate.

**Scope (explicit split, not a silent partial):** this PR is deliberately **gating-only**. The full branding sweep the operator asked for — ~105 DOGE hits, LTC 14 / Ltc 18 / LtcBlock 8 / litecoin 5, and the hardcoded blockchair explorer URLs — is **out of scope here** and gets its own follow-up PR (topology-driven symbol/explorer sourcing across all sites).

Issue #910, opened 2026-07-27
p2p: hostname addnodes are advertised to peers as 127.0.0.1 — DNS-named seeds are unreachable for everyone downstream


frstrtr:
## Observed on the live DASH hotel (production, 2026-07-27)

The primary hotel node is started with seven `--addnode` entries, two of which are DNS names:

```
--addnode rov.p2p-spb.xyz:8999
--addnode usa.p2p-spb.xyz:8999
```

Its log emits, repeatedly:

```
[warning] Write_IPV4: non-dotted address 'rov.p2p-spb.xyz', substituting 127.0.0.1
[warning] Write_IPV4: non-dotted address 'usa.p2p-spb.xyz', substituting 127.0.0.1
```

**Both hostnames resolve correctly from that host:**

```
rov.p2p-spb.xyz  -> 83.221.211.116
usa.p2p-spb.xyz  -> 66.151.242.154
```

So this is not a DNS failure. c2pool has the name, could resolve it, and instead **writes `127.0.0.1` onto the wire**.

## Root cause

`src/core/netaddress.cpp:11` `NetAddress::Write_IPV4`:

```cpp
std::vector<std::string> split_res;
boost::algorithm::split(split_res, ip, boost::is_any_of("."));
if (split_res.size() != 4)
{
    // Not a dotted-decimal IPv4 address (e.g. "localhost") — use loopback 127.0.0.1
    LOG_WARNING << "Write_IPV4: non-dotted address '" << ip << "', substituting 127.0.0.1";
    split_res = {"127", "0", "0", "1"};
}
```

The serializer splits on `.` and demands exactly 4 numeric parts. `rov.p2p-spb.xyz` splits into 3 non-numeric parts, so it silently becomes loopback. **`NetAddress` stores the hostname as a string and never resolves it before serialization.**

## Why this matters beyond a cosmetic warning

The value written by `Write_IPV4` is what we **advertise to other peers** (addrs/addrme gossip). Every peer that learns a node address from us via a DNS-named seed receives **`127.0.0.1`** — an address that, for them, points at *themselves*.

Consequences:
1. **DNS-named seeds do not propagate.** They work for the operator who typed them (the outbound connect path resolves normally — we do have live connections to `66.151.242.154`), but they are poisoned the moment they are gossiped onward.
[...]

Issue #875, opened 2026-07-26
ltc: tx-forwarding send path is dead code (requires-probe misses nested m_tx_info); v34+ shares carry no tx list — canonical parity unconfirmed


frstrtr:
## Summary

Two related findings on the shipping LTC/DOGE lane, surfaced while adding the F3 tx-completeness gate (PR #873).

## 1. VERIFIED — the tx-forwarding send path on LTC is dead code

`src/impl/ltc/node.cpp:721` gates the whole `remember_tx`/`forget_tx` send block on:

```cpp
if constexpr (requires { obj->m_new_transaction_hashes; })
```

No LTC share type has that as a direct member. `Share`(v17) and `NewShare`(v33) nest it one level down — `src/impl/ltc/share.hpp:94,105` declare `ltc::ShareTxInfo m_tx_info;`, and the vector lives at `src/impl/ltc/share_types.hpp:107`. So the probe is **false for every LTC share variant** and the block is compiled out entirely.

**LTC has therefore never forwarded a single transaction byte to a peer.** The receive side spells the path correctly (`protocol_actual.cpp:123`), which is why only the send half was silently dead.

Fixed in #873 via a single accessor (`src/impl/ltc/share_tx_refs.hpp`); without it the F3 gate would have been a literal no-op.

## 2. NEEDS CANONICAL REFERENCE — v34+ shares carry no tx list

`src/impl/ltc/share.hpp` serialises the tx list only below v34:

```cpp
if constexpr (version < 34)
{
    READWRITE(obj->m_tx_info);
}
```

and `SegwitMiningShare`(34) / `PaddingBugfixShare`(35) / `MergedMiningShare`(36) do not declare `m_tx_info` at all. So c2pool LTC v34/35/36 shares carry **no `new_transaction_hashes` and no `transaction_hash_refs`**.

Canonical p2pool's `share_info` carries `new_transaction_hashes` for share versions >= 13 (confirmed in the DASH fork at `p2pool/data.py:98`, which is v16). If the LTC-side canonical (jtoomim p2pool / p2pool-merged-v36) also carries it for v34+, then **every v34+ share c2pool mints is a wire divergence** — canonical peers re-derive `share_info` on receipt and reject on mismatch, which would mean silent rejection with no PPLNS credit.

**This could not be settled here** — no LTC-side canonical python is checked out locally, and the DASH fork is not the right reference for v34+. Someone with the jtoomim/merged-v36 source should confirm.

[...]

Issue #863, opened 2026-07-26
core(socket): no outbound write queue — overlapping composed writes can silently lose shares (all coins)


frstrtr:
## Summary

`core::Socket::write` (`src/core/socket.cpp:110-137`) starts a composed `boost::asio::async_write` immediately with **no outbound queue**. Initiating a second composed write before the first drains splices bytes mid-message, producing a bad checksum and a disconnect from canonical peers.

`send_shares` (`src/impl/dash/node.cpp`) issues **three back-to-back writes** on the same socket:

- `remember_tx` — node.cpp:1346
- `shares` — node.cpp:1357
- `forget_tx` — node.cpp:1362

`handle_version` also writes `getaddrs` (node.hpp:661) then `have_tx` (node.hpp:678) on the same socket. The comment at node.hpp:666-677 argues the `have_tx` is safe because nothing *follows* it — but something *precedes* it, which violates the same invariant. Once `m_known_txs` is non-empty this becomes the ~32 KB multi-round write the header at `tx_advertiser.hpp:43-70` explicitly warns about.

## Why this is reward-relevant, not cosmetic

`send_shares` returns its hashes as "sent" (node.cpp:1367-1373) **regardless of whether the write framed correctly**. `broadcast_share` then marks them in `m_shared_share_hashes` (node.cpp:1239-1240) and **never retries**. A corrupted frame therefore loses that share to that peer permanently — silent PPLNS loss with no counter and no log line.

## Fix

The outbound write queue in `core::Socket` already prescribed at `tx_advertiser.hpp:81-87`: queue composed writes per socket and drain strictly in order; only mark hashes as shared on completion, not on submission.

## Scope

`tx_advertiser.hpp:84-87` names this as pre-existing and affecting **every coin**, not just DASH.

PR #915, opened 2026-07-27
dash(p2p): originate addrme and fix the Legacy loopback self-probe (#882)

### LINKED ISSUE #882: p2p(addrme): legacy self-probe checks 127.0.0.0 instead of 127.0.0.1; DASH never originates addrme
### 1. Self-probe compares against `127.0.0.0`, not `127.0.0.1`

All five `protocol_legacy.cpp` implementations gate the addrme self-probe branch on:

```cpp
if (peer->addr().address() == "127.0.0.0")
```

`protocol_actual.cpp` and canonical p2pool both use **`127.0.0.1`**. `127.0.0.0` is the loopback *network* address, not the loopback *host* address — a peer will never present it. So the self-probe branch is unreachable in the Legacy generation, which is the only live generation on ltc/btc/dgb/bch (`handle_version` returns legacy unconditionally there).

Not fixed in #872 because changing it alters p2p address handling behaviour, which is a wider surface than the guard inversion that PR is scoped to.

### 2. DASH never originates an `addrme`

`handle_version` writes a `message_addrme` on:
- ltc `node.cpp:343`
- btc `node.cpp:344`
- bch `node.cpp:344`
- dgb `node.cpp:351`

**DASH has no equivalent.** `dash::message_addrme` is registered as a handler, but nothing in the DASH lane ever writes one, so DASH peers never learn our listen port from us directly.

This corrects a claim in #869, which asserted the origination gap was universal. It is DASH-only.

## Why these matter together

With #872 fixing the inverted relay guard, addrme relay starts working. But on DASH the message is still never originated, and on the Legacy path the self-probe branch still cannot fire. Both degrade inbound peer discovery rather than costing rewards directly — worse topology means slower share propagation and a higher orphan rate.

Suggested order: land #872 first (guard inversion, all coins), then treat these two as a follow-up so each behavioural change is separately bisectable.

frstrtr:
Closes #882. Last p2p protocol gap in the DASH lane — the other 12 of the 13 canonical pool messages are already aligned.

## Verified root cause

Both premises in the issue hold. Two line citations in the dispatch brief were off and are corrected below; nothing else changed.

### (a) DASH never ORIGINATED an addrme

`dash::message_addrme` is registered as a handler on both generations — **`src/impl/dash/node.hpp:1272`** (Legacy) and **`:1291`** (Actual), *not* 1173/1192 as cited — but a grep of the whole lane finds no writer. Every other lane writes one from `handle_version`:

| lane | origination site |
|---|---|
| ltc | `src/impl/ltc/node.cpp:346` |
| btc | `src/impl/btc/node.cpp:346` |
| bch | `src/impl/bch/node.cpp:344` |
| dgb | `src/impl/dgb/node.cpp:351` |
| **dash** | **absent** |

DASH's `handle_version` is header-inline (`src/impl/dash/node.hpp:699`); it writes `getaddrs` and the `have_tx` advert and nothing else. So a DASH peer could only ever learn our listen port second-hand, from a third party's `addrs` gossip.

### (b) The Legacy self-probe compared the loopback NETWORK address

**`src/impl/dash/protocol_legacy.cpp:55`** (pre-fix):

```cpp
if (peer->addr().address() == "127.0.0.0")
```

`127.0.0.0` is the `/8` network address. `peer->addr()` is `Socket::m_addr`, latched in `Socket::init()` (`src/core/socket.cpp:78`) from the real `remote_endpoint()` — a peer physically cannot present it. Canonical (`p2p.py:281`) and our own `src/impl/dash/protocol_actual.cpp:57` both compare `127.0.0.1`. Legacy is the **only** generation live on DASH today: `handle_version` returns `legacy` unless the operator has ratcheted the floor above the 1700 baseline.

The same `127.0.0.0` constant is present in the other four `protocol_legacy.cpp` files (btc/ltc/dgb `:55`, bch `:84`). **Deliberately left alone** — this PR is DASH-lane-scoped, and those lanes are live networks whose p2p behaviour should change in their own bisectable commit. #882 should stay open, or spawn a follow-up, for them.

## What canonical does, and how this matches it

`p2pool/p2p.py:109-134` `sendAdvertisement()`:

```python
def sendAdvertisement(self):
    if self.node.serverfactory.listen_port is not None:
        host = self.node.external_ip
        port = self.node.serverfactory.listen_port.getHost().port
        if host is not None:
            ...  send_addrs([our external ip:port])
        else:
            self.send_addrme(port=port)
```

Two things are load-bearing:

1. **The payload is our listen port.** The peer can already see our IP; the port is the one thing it cannot infer. Matched: the addrme carries `core::Server::listen_port_or_none()`.
2. **The whole advertisement is gated on `listen_port is not None`.** Matched — and this is the one place the PR does **not** copy ltc/btc verbatim, because copying them would have been a live regression:

   `core::Server::listen_port()` (`src/core/factory.hpp:90`) is `m_acceptor->local_endpoint().port()` with no guard. DASH has a real non-listening mode: `--connect` without `--listen` skips `listen()` entirely (`src/c2pool/main_dash.cpp:566`), leaving the acceptor **engaged but closed**, on which `local_endpoint()` throws. A throw out of `handle_version` makes the bridge drop the connection — so an unguarded port read would have broken `--connect` mode outright. That is not a theoretical reading: `DashAddrme.ConnectOnlyNodeHasNoListenPortAndUnguardedReadWouldThrow` constructs exactly that node and asserts the throw.

   The fix adds `core::Server::listen_port_or_none()` — canonical's `is not None`, expressed as `std::optional`. Purely additive; `listen_port()` is untouched, so no existing caller in any lane changes.

Two deliberate, documented divergences from canonical, both inherited from the existing ltc/btc port rather than introduced here:

* Canonical re-runs `sendAdvertisement()` on a repeating expovariate timer (`p2p.py:212-214`, mean `100 * len(peers) + 1` s). c2pool sends it **once at handshake**, in all four existing lanes. This PR matches c2pool, per "port the pattern, do not invent one." Adding the timer would be a cross-coin change, not a DASH gap-closure.
* Canonical prefers `send_addrs([external_ip])` when `--external-ip` is configured. c2pool has no `external_ip` knob on this path in any lane, so the `else` arm is the only reachable one and is what all five now implement.

Ordering: the addrme goes immediately after the existing `getaddrs`, **before** `advertise_known_txs()`. The #863 comment block right there is explicit that the (up to ~32 KB) `have_tx` advert must be the last write in the handler; a two-byte message inserted ahead of it preserves that.

## Why this is interop-safe

Our DASH node is joined to the live p2pool-dash sharechain with third-party operators on it, so this was the governing constraint.

* **No wire byte moves.** The fix adds a *call site*. `message_addrme` (`src/impl/dash/messages.hpp:47-55`) is untouched: still `command="addrme"`, still a single little-endian `uint16` port. Pinned by `DashAddrme.WireFormatByteIdentical` (payload bytes + NUL-padded 12-byte command field + `length == 2` inside the full framed packet) and by the pre-existing `DashPoolNodeMessages.Message_Addrme_LayoutPinned`.
* **The message was already in the protocol both ways.** DASH has parsed and handled inbound `addrme` since the reception slice; every canonical p2pool-dash node has originated one at us all along. This makes us *reciprocate* a message the network already speaks. It cannot surprise a peer.
* **Nothing consensus-adjacent is touched.** No share, target, coinbase, payee, subsidy or PPLNS path is read or written. Peer discovery only.
* **(b) is a strict reduction in what we emit and store.** See the behaviour question below.

## Does fixing (b) change observable behaviour toward real peers?

**Yes — but only for addrme sent from a loopback source, and strictly in the direction of emitting less garbage.**

For any peer with a routable address the two constants are indistinguishable: `x != "127.0.0.0"` and `x != "127.0.0.1"` are both true, both take the `else` arm, and the handler behaves identically byte for byte. **The overwhelming majority of live traffic is unaffected.**

When the sender *is* on loopback — a peer on the same host, or anything reaching us through a local proxy/tunnel — the pre-fix behaviour was:

1. `got_addr(NetService{"127.0.0.1", port})` — **loopback entered our AddrStore as a routable peer**, and
2. with p=0.8, an `addrs` message carrying that loopback record was **relayed to a random peer**, where it points at *themselves*.

Post-fix it takes canonical's self-probe arm: record nothing, and relay the `addrme` onward with p=0.8 so a peer that *can* see our public address is the one to answer it (`p2p.py:282-283`).

So the observable change toward real peers is: **we stop injecting `127.0.0.1` into other operators' address books.** That is the same class of harm #910 reports from the other direction, and it is why the issue calls the address-gossip subsystem wrong at both ends.

## Coordination with #910 / PR #911

Read #911 before writing this. The two are complementary and land on opposite ends of the same subsystem:

* **#911** fixes *serialization*: a DNS-named addnode must not be written to the wire **as** `127.0.0.1`. It touches `NetAddress::Write_IPV4`, adds `is_wire_advertisable()`, and guards the ten per-coin `HANDLER(getaddrs)` sites.
* **This PR** fixes *reception*: an addrme arriving **from** loopback must not be recorded and re-gossiped as routable. It touches `HANDLER(addrme)` in `src/impl/dash/protocol_legacy.cpp` only.

No duplication and no contradiction. #911 explicitly preserves `address()` / `to_string()` / the AddrStore key, which is exactly the accessor this fix's comparison reads, so the semantics it depends on are the ones #911 promises to keep.

Overlapping files are `src/core/factory.hpp` (#911 edits `Client::resolve()` ~`:157`; this adds a method to `Server` ~`:90`) and `src/impl/dash/protocol_legacy.cpp` (#911 edits `HANDLER(getaddrs)`; this edits `HANDLER(addrme)`). Verified clean: `git merge-tree --write-tree HEAD origin/fix/910-resolve-hostname-before-advertising` exits 0 — no conflict in either direction, whichever merges first.

## What the tests prove

`test/test_dash_addrme.cpp`, folded in as a **third TU of the existing allowlisted `test_dash_node` target** — no new `add_executable`, so it cannot land as a "Not Run" (#769). `tools/ci/check_test_target_allowlist.py` passes.

These drive the **real** `NodeImpl::handle_version` and the **real** `dash::Legacy::handle(message_addrme)` against a **real** loopback TCP socket, then read back the frames that actually went on the wire — not a re-implementation, and not a mock that could pass over unreachable code (#875/#878).

| test | proves |
|---|---|
| `WireFormatByteIdentical` | the addrme payload, command field and framed length are unchanged |
| `HandleVersionWritesAddrmeCarryingOurListenPort` | a listening node emits an addrme carrying its **actual ephemeral** bound port; `getaddrs` is asserted alongside it as a harness sanity check, so a silent harness failure cannot masquerade as a pass |
| `NonListeningNodeAdvertisesNothingAndDoesNotThrow` | canonical's `is not None` gate: no listen port, no advertisement, no throw |
| `ConnectOnlyNodeHasNoListenPortAndUnguardedReadWouldThrow` | the `--connect` shape (context present, `listen()` never called): `listen_port_or_none()` is `nullopt` **and** the unguarded `listen_port()` really does throw |
| `LoopbackPeerPresentsHostAddressNotNetworkAddress` | a real accepted loopback connection presents `127.0.0.1`, never `127.0.0.0` — the old guard was dead code |
| `LoopbackAddrmeIsNotRecordedAsARoutablePeer` | 32 real loopback addrmes leave the AddrStore free of `127.0.0.1:41337`, emit **no** `addrs` relay, and **do** emit the canonical addrme relay |

### Negative control (run on this tree)

Reverting each fix independently and rebuilding:

```
[  FAILED  ] DashAddrme.HandleVersionWritesAddrmeCarryingOurListenPort
             #882(a): handle_version never originated an addrme on the DASH lane
[  FAILED  ] DashAddrme.LoopbackAddrmeIsNotRecordedAsARoutablePeer
             #882(b): a loopback addrme was recorded in the AddrStore as a routable peer
             #882(b): loopback was gossiped to peers in an addrs relay
[  PASSED  ] 3 tests   (wire pin + both non-listening cases — correctly insensitive to the defects)
```

Each defect fails only its own case. With both fixes in place: **6/6 pass, full `test_dash_node` 24/24.**

## Build verification

Local Release build, gcc-13 + the CI conan profile:

* `c2pool-dash` — clean (the lane that changed)
* `c2pool` — clean (proves the shared `src/core/factory.hpp` addition breaks no other lane)
* `test_dash_node` 24/24, `test_dash_poolnode_messages` 10/10, `test_dash_p2p_messages` 8/8

Nothing was deployed anywhere. No live node was touched.

### DIFF
--- a/src/core/factory.hpp
+++ b/src/core/factory.hpp
@@ -88,6 +88,33 @@ public:
     }
 
 	uint16_t listen_port() const { return m_acceptor->local_endpoint().port(); }
+
+	/// Canonical p2p.py:110 — `if self.node.serverfactory.listen_port is not None`.
+	///
+	/// listen_port() above is only valid on a node that actually listens: it
+	/// dereferences the optional acceptor unconditionally (empty for the
+	/// rig-free `Server(nullptr, ...)` construction used by the KAT targets)
+	/// and calls local_endpoint() on it, which THROWS boost::system::system_error
+	/// (bad_descriptor) when the acceptor was never opened. c2pool has a real
+	/// non-listening mode — DASH's `--connect` suppresses the inbound listener
+	/// (src/c2pool/main_dash.cpp:566) — so an advertisement path that needs our
+	/// listen port must be able to ask "are we listening at all?" without
+	/// throwing. That is exactly the question canonical asks before deciding
+	/// whether to advertise, and this is the accessor that answers it.
+	///
+	/// Additive: listen_port() is untouched, so no existing caller changes.
+	std::optional<uint16_t> listen_port_or_none() const noexcept
+	{
+		if (!m_acceptor || !m_acceptor->is_open())
+			return std::nullopt;
+
+		boost::system::error_code ec;
+		const auto ep = m_acceptor->local_endpoint(ec);
+		if (ec)
+			return std::nullopt;
+
+		return ep.port();
+	}
 };
 
 class Client
--- a/src/impl/dash/node.hpp
+++ b/src/impl/dash/node.hpp
@@ -743,6 +743,31 @@ public:
         // consensus/mint state touched.
         peer->write(dash::message_getaddrs::make_raw(8));
 
+        // #882 — ORIGINATE addrme. Canonical p2p.py:109-134 sendAdvertisement():
+        // a listening node with no configured external IP asks the peer to
+        // advertise back what IT believes our address to be, carrying our listen
+        // port as the only thing the peer cannot observe for itself. Port of the
+        // ltc reference (src/impl/ltc/node.cpp:343-348; btc node.cpp:344-348,
+        // bch node.cpp:342-346, dgb node.cpp:349-353). DASH registered the
+        // handler on both generations (node.hpp:1272 Legacy / 1291 Actual) but
+        // nothing in the lane ever WROTE one, so a DASH peer could only ever
+        // learn our listen port second-hand from someone else's addrs gossip.
+        // Discovery only; no consensus/mint state touched.
+        //
+        // GUARDED, unlike the ltc/btc port. Canonical gates the whole
+        // advertisement on `if self.node.serverfactory.listen_port is not None`
+        // (p2p.py:110), and DASH has a real non-listening mode: `--connect`
+        // without `--listen` suppresses the inbound listener
+        // (src/c2pool/main_dash.cpp:566). core::Server::listen_port() reads
+        // m_acceptor->local_endpoint() unconditionally, which throws on an
+        // acceptor that was never opened — and a throw out of handle_version
+        // makes the bridge drop the connection, so an unguarded read would break
+        // --connect mode outright. listen_port_or_none() is canonical's
+        // `is not None` test; when we are not listening we advertise nothing,
+        // exactly as canonical does.
+        if (auto our_listen_port = core::Server::listen_port_or_none())
+            peer->write(dash::message_addrme::make_raw(*our_listen_port));
+
         // Canonical p2p.py:276 — one FULL have_tx as soon as the handshake
         // completes, so the peer's view of our tx pool starts correct (and its
         // dashboard stops showing us as txpool = 0). Subsequent sweeps send
--- a/src/impl/dash/protocol_legacy.cpp
+++ b/src/impl/dash/protocol_legacy.cpp
@@ -52,7 +52,17 @@ void Legacy::HANDLER(addrs)
 
 void Legacy::HANDLER(addrme)
 {
-    if (peer->addr().address() == "127.0.0.0")
+    // #882 — the loopback HOST address, not the loopback NETWORK address.
+    // Canonical p2p.py:281 compares `host == '127.0.0.1'`, and so does the
+    // Actual generation (protocol_actual.cpp:57). Legacy shipped "127.0.0.0",
+    // the /8 network address, which no peer can ever present as its source IP,
+    // so this branch was unreachable and EVERY addrme — including one arriving
+    // over loopback — fell into the else arm below: it recorded
+    // 127.0.0.1:<port> in our AddrStore as if it were a routable peer and then
+    // gossiped that loopback record onward in an addrs message. Legacy is the
+    // only generation live on DASH today. Related to #910, which fixes the
+    // other end of the same subsystem (a hostname serialising AS 127.0.0.1).
+    if (peer->addr().address() == "127.0.0.1")
     {
         if (!m_peers.empty() && (core::random::random_float(0, 1) < 0.8))
         {
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -304,7 +304,15 @@ if (BUILD_TESTING AND GTest_FOUND)
     # target would need a build.yml --target edit and, until it had one, would
     # trip CI's NOT_BUILT sentinel (the drift-guard audit, #883/#893). It needs
     # exactly this link set — NodeImpl + ShareTracker + core MiningInterface.
-    add_executable(test_dash_node test_dash_node.cpp test_dash_dashboard_found_block.cpp)
+    # test_dash_addrme.cpp (#882): the addrme origination + Legacy self-probe
+    # KATs. Drives the REAL NodeImpl::handle_version and the REAL
+    # dash::Legacy::handle(message_addrme) over a REAL loopback TCP socket, so it
+    # needs exactly this link set (NodeImpl + the `dash` OBJECT lib carrying
+    # protocol_legacy.cpp + core Socket/Packet). FOLDED in as a third TU for the
+    # same reason as the second: a standalone target would need a build.yml
+    # --target edit and would trip CI's NOT_BUILT sentinel until it had one
+    # (#769/#883/#893).
+    add_executable(test_dash_node test_dash_node.cpp test_dash_dashboard_found_block.cpp test_dash_addrme.cpp)
     target_link_libraries(test_dash_node PRIVATE
         GTest::gtest_main GTest::gtest
         dash_x11 core
--- a/test/test_dash_addrme.cpp
+++ b/test/test_dash_addrme.cpp
@@ -1,0 +1,461 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+//
+// DASH addrme KAT — issue #882.
+//
+// Two independent defects, both proven here against the REAL code paths rather
+// than against a re-implementation:
+//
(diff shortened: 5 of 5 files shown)


frstrtr:
**Review gate — one test required before merge.** Otherwise this is close and the work is good.

## The gap (raised by code review, and I agree)

Your negative control proves **loopback** senders now take canonical's self-probe arm. It does **not** prove that **routable** senders are byte-identical before and after the `127.0.0.0` → `127.0.0.1` change.

That is the interop question, and it is the one that matters most here: our DASH node is joined to the **live public p2pool-dash sharechain right now**, alongside third-party operators we do not control. The reasoning that routable peers take the `else` arm identically under both constants is sound — but it is reasoning, not evidence, and this repo has repeatedly punished the difference.

**Please add:** one test driving a **routable-source** addrme through the real `Legacy::handle(message_addrme)` and asserting the stored/relayed bytes are identical under both constants. If one of the existing 24 already does this, say which and I will withdraw the ask.

## What I verified myself

- `factory.hpp:90` — `listen_port()` is a bare `m_acceptor->local_endpoint().port()`, unguarded ✓
- `ltc/node.cpp:346` calls it bare ✓
- `main_dash.cpp:566` — `connect_only` skips `listen()`, leaving the acceptor engaged-but-closed ✓

So your refusal to do a verbatim port was correct, and **my instruction was the wrong one**. I told you "port the ltc/btc pattern, do not invent one" while also saying "match canonical exactly"; those conflicted and canonical wins. A verbatim port would have thrown out of `handle_version` and failed every `--connect` handshake. Gating on `listen_port_or_none()`, mirroring canonical's `if serverfactory.listen_port is not None` (`p2p.py:110`), is right — and writing a test that **empirically confirms the throw** rather than asserting it from docs is exactly the standard I want.

Thank you also for correcting my line numbers (handlers are `node.hpp:1272`/`:1291`, not 1173/1192).

## Two follow-up conditions, recorded so they are not lost

1. **The other four lanes carry the same unguarded `listen_port()` call.** I described that as "latent"; review pushed back and is right — latent-vs-live must be **enumerated per lane**, not inferred from the absence of a known non-listening mode. The follow-up issue's acceptance criterion is *"each lane's connect/listen paths audited"*, not *"swap in the guarded accessor."*
2. **Scope decision endorsed:** leaving `127.0.0.0` in btc/ltc/dgb (`protocol_legacy.cpp:55`) and bch (`:84`) for per-lane bisectable commits on live networks is better than one sweep. #882 stays open, and the unguarded-throw audit folds into the same follow-up since it has identical lane-by-lane shape.

## Coordination

#911 is reviewed and near-green. `git merge-tree` exits 0 against it in either order, and the two are complementary — #911 fixes serialization (hostname written **as** loopback), this fixes reception (addrme **from** loopback). No action needed, just do not rebase past each other unnecessarily.

Add the routable-sender test and I will merge on green.

PR #916, opened 2026-07-27
ci(prune): preserve .github so local composite actions survive post-step prep (fixes #890 red leg)

frstrtr:
Root cause: prune-runner-workspace was deleting .github/ so the runner cannot re-read provision-dashbls action.yml during post-step prep. That is what turned #890 red on the dash-bls leg — a failure with nothing to do with libgmp.

The prune step now preserves .github/ (composite action definitions) alongside _actions/_tool, so post-step prep can re-read local action.yml files. Workspace-size reclaim is otherwise unchanged.

Note: #890 still needs a rebase onto this branch once merged.

### DIFF
--- a/.github/actions/prune-runner-workspace/action.yml
+++ b/.github/actions/prune-runner-workspace/action.yml
@@ -37,9 +37,14 @@ runs:
         # Everything under the workspace except the checkout's .git. Keeping .git
         # (~230 MiB) lets actions/checkout's post-step unset its auth header
         # instead of erroring, and lets the next job fetch incrementally. The
-        # build trees, which are the actual ~5 GiB, all go.
+        # build trees, which are the actual ~5 GiB, all go. .github is also
+        # kept: local composite actions (./.github/actions/*) resolve from
+        # the workspace, not _actions/, and the runner re-reads their
+        # action.yml during end-of-job post-step preparation. Deleting it
+        # made provision-dashbls unresolvable and failed every DASH-impl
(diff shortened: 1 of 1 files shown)


frstrtr:
**Reviewed — correct, and it fixes a fault I misattributed.**

```diff
-find "$GITHUB_WORKSPACE" -mindepth 1 -maxdepth 1 ! -name .git -exec rm -rf -- {} +
+find "$GITHUB_WORKSPACE" -mindepth 1 -maxdepth 1 ! -name .git ! -name .github -exec rm -rf -- {} +
```

That is precisely the failure seen on #890 and on the `SHA256SUMS + draft release` leg of run 30221138155:

```
Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under
'.../.github/actions/prune-runner-workspace'. Did you forget to run actions/checkout?
```

The prune deleted `.github`, which contains the local composite action the post-step then tries to run — self-defeating on a reused self-hosted workspace. Preserving `.github` alongside `.git` is the minimal correct fix, and the ~5 GiB of build trees this exists to reclaim still go.

Worth noting the blast radius this had: because that job failed, run 30221138155 produced **no draft release at all** despite 12 of 14 packages building fine. A green matrix was withheld by a workspace-prune bug.

I originally attributed #890's red to the missing `libgmp-dev` alone. That was incomplete — there were two independent faults on that leg. Please confirm whether `libgmp-dev` is now present on the runners taking the DASH+BLS job, since this PR does not address it.

Merging on green.

PR #917, opened 2026-07-27
build(windows): add /bigobj for MSVC so c2pool-dash links (fixes #909)

### LINKED ISSUE #909: build(windows): c2pool-dash fails C1128 — main_dash.cpp exceeds MSVC object section limit, needs /bigobj
## Symptom

`dash package (Windows x86_64)` fails deterministically on the Release matrix:

```
C:\actions-runner\_work\c2pool\c2pool\src\c2pool\main_dash.cpp(1,1):
error C1128: number of sections exceeded object file format limit: compile with /bigobj
[build_ci\src\c2pool\c2pool-dash.vcxproj]
```

Observed on run **30221138155** (master `07084676`). Every other package in the matrix succeeded — linux/macOS for all five coins, and Windows for btc/ltc/dgb/bch. **`c2pool-dash` is the only Windows target affected.**

## ⚠️ This is NOT the known Conan flake — do not re-run and expect green

There is a documented pattern of Windows CI reddening on Conan dependency-resolution flakes that clear on re-run. **This is a different failure and it will recur on every build until the flag is added.** C1128 is a hard compiler limit: MSVC's default COFF object format caps a translation unit at 65,279 sections, and `main_dash.cpp` has crossed it. Nothing about a re-run changes that.

Evidence it is not a flake: the Conan stage of this very job completed successfully (`Conan: Target declared 'ZLIB::ZLIB'`, `'BZip2::BZip2'`, `'nlohmann_json'`, `'yaml-cpp'`, LevelDB found), and the `boost_sentinel` test passed 1/1. The failure is squarely in compilation of one TU.

## Why now

`main_dash.cpp` is the largest per-coin main in the tree and has been growing steadily — the embedded-daemon wiring, the mint seam, the block-relay arms, and the recently merged #888 / #901 / #903 all add code or heavily-templated call sites to it. Heavy `if constexpr` / template instantiation is what generates sections, so this threshold was always going to be crossed by accretion rather than by any single commit. Do not treat any one of those PRs as "the cause".

## Fix

Add `/bigobj` for MSVC on this target. Narrow form, preferred:

```cmake
if(MSVC)
    target_compile_options(c2pool-dash PRIVATE /bigobj)
endif()
```

`/bigobj` raises the section cap to 2^32. It has **no effect on the emitted code** — same semantics, same behaviour, larger object file only. It is not a workaround that hides a defect; it is the documented and intended flag for large TUs.

**Consider applying it to every `c2pool-*` target rather than dash alone.** The other coins are on the same growth curve and will each hit this in turn; fixing them one at a time as they break costs a red matrix every time. A single `if(MSVC)` block covering all coin targets is the same diff size and prevents four repeats.

## Scope / severity

- **Does NOT affect Linux or macOS.** Both build `c2pool-dash` cleanly on this same commit; only MSVC has this object-format limit.
- **Production is unaffected.** The DASH hotel runs Linux, and `pkg-linux-dash` from this run is verified good (binary self-reports `c2pool-dash 07084676d`, `/P2Pool-DASH/` marker and `[MINT-BLOCK]` paths present, all shared-lib deps resolved).
- Blocks Windows releases of the DASH lane, and reddens the whole Release matrix — which also fails the downstream `SHA256SUMS + draft release` job, so **no draft release is produced at all** even though 12 of 14 packages built fine.

## Secondary, unrelated failure in the same run

`SHA256SUMS + draft release` additionally failed with:

```
Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under
'.../.github/actions/prune-runner-workspace'. Did you forget to run actions/checkout?
```

That is a self-hosted-runner checkout-ordering problem — the local composite action is referenced before `actions/checkout` has materialised it. Same class of failure seen on #890. Worth a separate issue; noted here so the two are not conflated.

frstrtr:
MSVC hits the 65279-section object-file limit compiling the dash translation units; /bigobj raises it. No effect on non-MSVC toolchains.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,11 @@ if(MSVC)
     # toolchain does not guarantee CMake's default /EHsc, so set it explicitly
     # for every MSVC target — the coin binaries and the CI boost_sentinel alike.
     add_compile_options(/EHsc)
+    # /bigobj: main_dash.cpp -- and other coin main_*.cpp as template
+    # instantiation grows -- exceeds MSVC's 65,279-section object limit
+    # and fails C1128. /bigobj raises the cap; it is link-compatible and
+    # has no runtime cost, so apply it to every MSVC target.
+    add_compile_options(/bigobj)
 endif()
 
 # Handle CMake 3.30+ policy for FindBoost removal


frstrtr:
**Reviewed — correct, and better scoped than what I asked for.**

`add_compile_options(/bigobj)` sits inside the existing `if(MSVC)` block next to `/EHsc`, so gcc/clang are untouched. I suggested `target_compile_options(c2pool-dash PRIVATE /bigobj)`; directory-scoped is the better call — it covers **every** coin target at once, which is exactly the "do not fix these one at a time as each crosses the limit" point from #909. `main_dash.cpp` crossed first; the others are on the same template-instantiation growth curve.

Verified the premise myself before filing #909: run 30221138155 failed `C1128: number of sections exceeded object file format limit` on `main_dash.cpp`, while Conan resolution and `boost_sentinel` both **succeeded** in that same job — so it was never the known Conan flake, and no number of re-runs would have cleared it.

`/bigobj` changes the object format only: no codegen change, no runtime cost, link-compatible. The comment says as much and is accurate.

Merging on green.

PR #821, opened 2026-07-23
dash: T1c byte-parity KAT vs live p2pool-dash oracle corpus (share-v16)

frstrtr:
## T1c byte-parity — the last gate before ★M1 JOIN

Closes the serialization-parity gap the 2026-07-21 join-lane audit flagged:
`test_dash_conformance.cpp` only exercised synthetic / out-of-band vectors —
there was no byte-parity test against a **real** p2pool-dash node. This adds one.

### What
- **`test/data/dash_oracle_corpus.hpp`** — 29 distinct canonical shares captured
  off a controlled p2pool-dash node **JOINED to the live public sharechain**
  (share-v16 / min-proto-1700) via the oracle-gate on VM210. Each raw wire entry
  `[VarInt type=16][VarStr contents]` was cross-validated **byte-identical across
  4 independent peers**, deduped by `share_hash`.
- **`test/test_dash_oracle_byte_parity.cpp`** — for every canonical `SHARE_RAW`:
  round-trips it through the production `load_share` → `DashFormatter::Read` →
  `DashFormatter::Write` path and asserts **`serialize(parse(raw)) == raw`** — the
  exact condition for a c2pool-relayed share to `VERIFY_PASS` on the fork — plus
  semantic cross-checks (`prev_hash` / `donation` / `new_tx_count` /
  `pubkey_hash → payout_script`) against the fork's independently-parsed
  `SHARE_RECV` fields.

### Result
- **30/30 tests pass.** c2pool's DASH v16 wire layout **already byte-matches** the
  canonical p2pool-dash fork — **zero serialization divergences found**. The gap
  was test coverage (synthetic vectors), now closed with the real captured corpus.

### Live corroboration (out-of-band, not in this PR)
A `c2pool-dash --run --connect <oracle>:8999` node was **accepted as a peer by the
canonical fork** (`Incoming connection ... p2pool version: 1700 'c2pool'`, peers
4→5), **downloaded the full live sharechain and parsed every share with 0 parse
failures / 0 VERIFY_FAIL**, and disconnected cleanly — **0 VERIFY_DIFF / 0 NOBAN**
logged by the oracle. The offline round-trip KAT here pins that parity as a
repeatable gate with no live dependency.

### Not yet exercised
The `c2pool mints a fresh share → oracle byte-VERIFY_PASSes it` leg needs a
stratum miner + coin RPC on the joined chain (the mint-and-relay step); the
connect-only download node above does not produce shares. Serialization parity
[...]

PR #918, opened 2026-07-27
test(ci): discover guarded dash-bls/int128 cases so they skip, not falsely-pass (fixes #895)

### LINKED ISSUE #895: ci: gtest_add_tests(AUTO) registers #ifdef-guarded cases that cannot run — ctest reports them as Passed, not skipped

frstrtr:
Narrowed per review: only the three targets that actually register compile-guarded TEST cases — test_dash_bls_verify and test_dash_quorum_members (#ifdef C2POOL_DASH_BLS) and test_dash_coinbase_muldiv (#if defined(__SIZEOF_INT128__)/#else).

gtest_add_tests(... AUTO) source-scans at configure time and registers guarded-out cases unconditionally; on a build with the guard off, ctest invokes a --gtest_filter that matches nothing and reports success — a false pass masking absent coverage. gtest_discover_tests enumerates the built binary, so a compiled-out case is simply not registered.

Suites with runtime GTEST_SKIP (test_threading) or guards over includes only (test_coin_broadcaster) are correct as-is and left unchanged. The earlier blanket 96-target version is preserved on ci-steward/gtest-discover-blanket-superseded.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -567,7 +567,7 @@ if (BUILD_TESTING AND GTest_FOUND)
         ${Boost_LIBRARIES}
     )
     target_link_libraries(test_dash_bls_verify PRIVATE c2pool_payout c2pool_hashrate c2pool_merged_mining)  # OBJECT-lib SCC direct-naming (#22/#39)
-    gtest_add_tests(test_dash_bls_verify "" AUTO)
+    gtest_discover_tests(test_dash_bls_verify DISCOVERY_MODE PRE_TEST)
 
     # E1 Phase-L member-set selection: deterministic ComputeQuorumMembers over
     # the SML AS OF the WORK block (base - 8, per v23.1.7 GetAllQuorumMembers —
@@ -588,7 +588,7 @@ if (BUILD_TESTING AND GTest_FOUND)
         ${Boost_LIBRARIES}
     )
     target_link_libraries(test_dash_quorum_members PRIVATE c2pool_payout c2pool_hashrate c2pool_merged_mining)  # OBJECT-lib SCC direct-naming (#22/#39)
-    gtest_add_tests(test_dash_quorum_members "" AUTO)
+    gtest_discover_tests(test_dash_quorum_members DISCOVERY_MODE PRE_TEST)
 
     # E1 Phase-L member-set SOURCING PLUMBING (quorum_member_source.hpp): the
     # #814 re-review findings — R1 send-once-per-work-block dedup + strict
@@ -1125,7 +1125,7 @@ if (BUILD_TESTING AND GTest_FOUND)
         ${Boost_LIBRARIES}
     )
     target_link_libraries(test_dash_coinbase_muldiv PRIVATE c2pool_payout c2pool_hashrate c2pool_merged_mining)  # OBJECT-lib SCC direct-naming (#22/#39)
-    gtest_add_tests(test_dash_coinbase_muldiv "" AUTO)
+    gtest_discover_tests(test_dash_coinbase_muldiv DISCOVERY_MODE PRE_TEST)
 
     add_executable(test_dash_cb_payee test_dash_cb_payee.cpp)
     target_link_libraries(test_dash_cb_payee PRIVATE


frstrtr:
**Reviewed — this is the right fix for the right reason, and it is the most valuable of the three.**

```diff
-gtest_add_tests(test_dash_bls_verify "" AUTO)
+gtest_discover_tests(test_dash_bls_verify DISCOVERY_MODE PRE_TEST)
```
(and the same for `test_dash_quorum_members`, `test_dash_coinbase_muldiv`)

`gtest_add_tests(... AUTO)` scans **source text** at configure time and registers every `TEST(...)` it finds — whether or not the case survives compilation. Cases behind `#ifdef C2POOL_DASH_BLS` / int128 guards therefore get registered but never built, and ctest reports them **"Not Run"**. `gtest_discover_tests(... PRE_TEST)` queries the **actual binary** at test time, so a guarded-out case simply does not exist and is skipped honestly.

That is the exact mechanism behind #895 and #769 — and it is not academic. PR #821 (the T1c byte-parity corpus, 29 shares captured off a live p2pool-dash node and cross-validated across 4 peers) is red for this and nothing else. PR #912 was red for the same reason this morning. It is also the same shape as the dead-probe class in #875/#880/#905: **code that is registered, reported, and never executed.**

The distinction that matters: this converts a false-pass into an honest skip. It does **not** make guarded tests run — `#890` is still the PR that has to make the BLS build actually happen. Both are needed, and neither substitutes for the other.

Merging on green.

Issue #961, opened 2026-07-29
core(address): address_to_hash160 never checks the version byte against the coin's network — a BTC/LTC/DOGE address is silently accepted and paid to a DASH script derived from it (all lanes share this)


frstrtr:
## Defect

`core::address_to_hash160()` (`src/core/address_utils.cpp:103-128`) decodes
base58check and reads the version byte **only to choose P2SH vs P2PKH**. It
never validates the version against the coin's own network:

```cpp
uint8_t version = decoded[0];
if (version == 0x32 || version == 0x05 || version == 0x3a ||
    version == 0xc4 || version == 0x16) {
    addr_type = "p2sh";
} else {
    addr_type = "p2pkh";
}
return h160;
```

Any base58check-valid address from **any** chain — a Bitcoin `1...`, a
Litecoin `L...`, a Dogecoin `D...` — passes, yields a 20-byte hash160, and
`address_to_script()` builds a well-formed 25-byte DASH P2PKH script from it.

## Consequence

A miner who authorizes with an address for the wrong chain is **silently
paid to a DASH script derived from that chain's hash160**. No warning, no
rejection — the share is accepted and the coinbase is built.

The funds are only recoverable if the miner happens to hold the same private
key, and they have no indication of where the payout went. From the miner's
side this is indistinguishable from working correctly, right up until they
look for coins that are not at the address they expected.

This also means the "invalid address" path is much narrower than it appears:
only genuinely malformed or empty usernames produce an empty script. A
wrong-network address is NOT caught anywhere.

## Scope: all lanes

`src/core/address_utils.cpp` is shared. LTC, DGB, BCH and BTC have the same
hole. The merged-mining lanes are of particular concern because operators
routinely paste two addresses for two different chains, and a swap is an easy
mistake — p2pool-merged-v36 explicitly added a swapped-order guardrail for
exactly this reason, which suggests the mistake is common in practice.

## Proposed

1. Validate the decoded version byte against the active coin's
   `address_version` / `address_p2sh_version` (already available in
   `core::CoinParams`, used by `decode_payee_script` in
   `src/impl/dash/share_producer.hpp`).
2. On mismatch, reject at `mining.authorize` with a clear error naming the
   expected network — fail loudly at connect time, not silently at payout
   time.
3. Add tests: a valid DASH address accepted; a BTC/LTC/DOGE address rejected
   with the network named; a malformed string rejected.

Rejecting at authorize is safe here in a way that rejecting ownerless work
[...]

PR #962, opened 2026-07-29
Fix: Validate address version byte against the coin's network

nivokvo:
Closes #<issue_number>

This PR fixes the address validation to check the version byte against the coin's network. It adds a check in `address_to_hash160` to ensure that the version byte of the address matches the expected network. If the version byte does not match, it throws an `std::invalid_argument` exception with a message indicating the expected network.

It also adds tests to verify that valid addresses are accepted, invalid addresses are rejected, and wrong-network addresses are rejected with a clear error message.


frstrtr:
Thank you for looking at this — the underlying diagnosis is correct. `address_to_hash160` really does read the base58 version byte only to choose between P2PKH and P2SH, and never checks it against the active coin's network. That is a real defect and it is tracked as #961.

Unfortunately this PR cannot be merged, and the reasons go beyond style.

## It replaces the file rather than changing the function

`src/core/address_utils.cpp` goes from 450 lines to 17, keeping only `address_to_hash160`. The following functions are deleted, and these are the call sites that remain in the tree:

| deleted | call sites |
|---|---|
| `address_to_script` | 40 across 15 files |
| `script_to_address` | 33 across 10 files |
| `is_address_for_chain` | 11 across 5 files |
| `address_to_hash160` | 10 across 5 files |
| `classify_script` | 9 across 5 files |
| `hash160_to_merged_script` | 9 across 5 files |
| `base58check_to_hash160` | 8 across 4 files |
| `register_address_decoder` | 4 across 3 files |
| `pubkey_to_p2pkh_address` | 4 across 2 files |

That is roughly 128 call sites. The tree does not compile. `register_address_decoder` is also the pluggable per-coin decoder registry, and `hash160_to_merged_script` is on the merged-mining path.

## The signature changes silently

Existing (`src/core/address_utils.hpp:40`):

    std::string address_to_hash160(const std::string& address, std::string& addr_type);

This PR:

    uint160 address_to_hash160(const std::string& address, const CoinParams& params);

Different return type and different parameters, so all existing callers break even setting the deletions aside.

## It will not compile for other reasons too

* `#include "core/coin_params.h"` and `#include "core/address_utils.h"` — this tree uses `.hpp` throughout. A new `address_utils.h` is added alongside the existing `address_utils.hpp` rather than editing it.
* `DecodeBase58` is not available in this translation unit. `core` provides `DecodeBase58Check`.

No CI ran on the branch, so none of the above was caught before opening.

## Two defects in the new code itself

**Checksum validation is dropped.** The replacement uses `DecodeBase58`, not `DecodeBase58Check`. The original validated the base58check checksum. This function feeds payout script construction, so accepting an address whose checksum does not verify means paying to a mistyped address. That is a more serious problem than the one being fixed.

**Out-of-bounds read.** The guard is `decoded.size() < 1`, and the code then copies 20 bytes:

    if (decoded.size() < 1) throw std::invalid_argument("Address too short");
    std::copy(decoded.begin() + 1, decoded.begin() + 21, h160.begin());

Any address decoding to fewer than 21 bytes reads past the end.

## Licensing

`// SPDX-License-Identifier: AGPL-3.0-or-later` is removed with nothing in its place. This project is AGPL-3.0-or-later with an Apache-2.0 engine split, and contributions cannot strip the identifier.

## What the fix should look like

The change belongs inside the existing `address_to_hash160`, at roughly ten lines: after `DecodeBase58Check` succeeds, compare `decoded[0]` against the active `CoinParams::address_version` and `address_p2sh_version` and reject on mismatch, naming the expected network in the error. Keep the checksum path, keep the `addr_type` out-parameter, keep every other function in the file, keep the SPDX line.

Note also that `src/core/address_utils.cpp` is shared by every coin lane, so the fix affects LTC, DOGE, DGB, BCH and BTC as well as DASH — #961 is written up with that scope.

Closing this. #961 stays open and will be fixed there.

PR #975, opened 2026-07-30
btc(compact): merkle backstop on reconstructed blocks, discard-not-deliver

frstrtr:
## Problem

`ReconstructBlock` (`src/impl/btc/coin/compact_blocks.hpp`) declares a compact block
**complete** the instant every slot is filled and delivers it downstream into the UTXO
cache — with **no** check that the reconstructed transactions actually hash to the
header's committed merkle root. There were **zero** `merkle` references in the 319-line
file.

BIP 152 short IDs are **48-bit SipHash over wtxid**. A short-ID collision can silently
substitute the *wrong* transaction into a filled slot; the block still looks complete,
and the pre-fix code delivered that **forged block** as real.

## Fix

Add `ReconstructedMerkleRoot()`, which folds the **non-witness** txids of the
reconstructed vector through the **sealed** `btc::coin::compute_merkle_root`
(`template_builder.hpp`) — the *same* walk that validates mined blocks, **not** a second,
divergent merkle path. `ReconstructBlock` verifies that root against
`header.m_merkle_root` before completing; on mismatch it sets a new `merkle_mismatch`
flag, leaves `result.block` empty, and returns. **Discard, do not deliver.**

Both delivery sites in `coin/p2p_node.hpp` honour it:

- the `cmpctblock` handler, on `merkle_mismatch`, requests the **full block via getdata**
  (`request_full_block`) rather than a getblocktxn with no missing indexes;
- the `blocktxn`-completed path runs the same check before delivering — even once the
  missing txns are authoritative from `blocktxn`, a *short-ID-matched* slot can still
  collide.

`merkle_mismatch` is kept **distinct** from `missing_indexes`, so a genuinely incomplete
block still takes the `getblocktxn` path, not the getdata full-block fallback.

## Scope

`src/impl/btc/**` only. The file is one of **four** per-coin copies; the `ltc/` copy is
dispatched to the LTC/DOGE lane and the `dgb/`/`bch/` copies are **not** touched here.

## Note (stated, not fixed)

`main_btc.cpp` around `:366` uses `submit_block` (always full-block) while the path around
`:738` reaches the compact-capable one — same function family, different coverage. Left
un-generalised per the deliberate no-generalise policy between those two sites.

## Test

`reconstruct_merkle_backstop_test.cpp` — three KATs:

- **HonestReconstructionCompletes** — correct slot tx verifies → `complete=true`,
  `merkle_mismatch=false`.
- **CollisionSubstitutesWrongTx** — a slot filled by a tx whose txid differs from the
  committed one → `complete=false`, `merkle_mismatch=true`, **empty `result.block`**
  (discard-not-deliver). Deleting the merkle check flips this red — pre-fix, all-slots-filled
  meant `complete=true` and delivery.
- **MissingTxIsNotAMerkleMismatch** — a genuinely missing tx reports `missing_indexes`
  with `merkle_mismatch=false`, keeping the getblocktxn path distinct from the getdata one.

Full suite: `btc_share_test` **148/148** green.


frstrtr:

PR #981, opened 2026-07-30
bch(p2p): merkle-root backstop on compact-block reconstruction

frstrtr:
BIP152 reconstruction declared a block complete once every slot was filled, with no comparison against the header merkle root. 48-bit short-id collisions can substitute the wrong tx; the header commitment is the only backstop.

- ReconstructedMerkleRoot() over the sealed bch::coin::compute_merkle_root (same walk GBT/emit_full_block use) gates result.complete.
- On mismatch: merkle_mismatch=true, discard, fall back to full getdata. Belt-and-suspenders re-check at the p2p_node delivery site.
- Test case models a 48-bit short-id collision substituting a wrong tx -> complete=false, merkle_mismatch=true, empty block (no forgery). CI-wired (build.yml x2). ctest bch_compact_block_connector_test PASS.

Reachability confirmed: p2p_node.hpp includes compact_blocks.hpp and p2p_node.cpp compiles directly into the c2pool-bch executable (src/c2pool/CMakeLists.txt). Per-coin isolation held: src/impl/bch only.

### DIFF
--- a/src/impl/bch/coin/compact_blocks.hpp
+++ b/src/impl/bch/coin/compact_blocks.hpp
@@ -28,6 +28,7 @@
 #include "block.hpp"
 #include "transaction.hpp"
 #include "mempool.hpp"  // compute_txid()
+#include "merkle.hpp"   // compute_merkle_root() -- sealed consensus merkle SSOT
 
 #include <core/uint256.hpp>
 #include <core/pack.hpp>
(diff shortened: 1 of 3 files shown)


frstrtr:
ID form: txid-only. BCH has no segwit, so there is no witness-stripped/wtxid distinction — the coinbase merkle commitment is over plain txids. ReconstructedMerkleRoot() reuses bch::coin::compute_merkle_root (the same double-SHA256 txid walk GBT and emit_full_block use), so BTC-style wtxid handling is deliberately not ported. Confirming per integrator acceptance criterion 3.

frstrtr:

Issue #984, opened 2026-07-30
dash(embedded): PoSe validity is never reconciled into the payee axis — sync_validity_from_sml() has ZERO production callers, wedging all daemonless template production


frstrtr:
## Summary

Daemonless DASH cannot serve a single block template. Root cause is a **dead-code wiring gap**, not a logic error: the function that reconciles PoSe validity into the payee axis has **zero production callers**.

## Confirmed against a live dashd

The masternode projected at the block that fail-closes the bridge:

```
proTxHash          e8626fcd57f6394db6669ad6db6fd44c
PoSeBanHeight      2513418      <- banned 71 blocks before we projected it
PoSePenalty        2972         (= registered count, maxed)
PoSeRevivedHeight  2454297      <- an EARLIER revival: the trap
lastPaidHeight     2511423
```

Per-height check vs `protx list valid`:

| height | projected MN in valid set? |
|---|---|
| 2513168 | yes |
| 2513261 | yes |
| **2513489** | **NO — registered but banned** |

Population at 2513489: **valid 2066 / registered 2972**.

The pinned anchor is **correct**: `protx list valid @ h=2513000` = 2068, matching the anchor's 2068 exactly.

## Mechanism

The validity filter **exists and is correct** — `if (!st.isValid) continue;` at `mn_state_machine.hpp:350` and `:256`. The struct carries the full PoSe triple at `mn_state_db.hpp:79-84`.

The defect is that `MnStateMachine::sync_validity_from_sml()` (`mn_state_machine.hpp:468`) is **never called in production**. Only the unit tests call it.

Two axes, both live, never joined:

- **payee axis** (`MnStateMachine`) — seeded once from RPC `protx list valid true`
- **SML axis** (`CSimplifiedMNList`) — carries authoritative per-entry `isValid`, advances on every `mnlistdiff`, and never touches `m_mnstates`

**PoSe bans are consensus-driven, not transaction-driven.** They never appear as a tx, so `apply_block()` structurally cannot observe one.

Because the seed uses `protx list valid true`, banned nodes are pre-filtered at startup — which is exactly why this stays invisible until a ban lands *after* seeding.

## Timeline

```
h=2513000  anchor seeded; node genuinely still valid -> seed CORRECT
h=2513418  PoSe ban by consensus. No tx. apply_block() cannot see it.
           Our queue keeps isValid=true.
h=2513489  node comes up as argmin -> projected -> dashd paid someone else
[...]

PR #985, opened 2026-07-30
dash(payee): wire SML->payee validity reconcile on the maintainer axis (live-run half; checkpoint lane still open)

frstrtr:
## Summary

Fixes the daemonless payee-desync (the block-viability gate for daemonless mining). The fix is a narrow two-line wiring change: call the already-existing, already-tested, but **never-invoked** reconciler `MnStateMachine::sync_validity_from_sml()`.

## Root cause (ninth instance of the wired-but-never-called class)

PoSe bans are **consensus-driven, not transaction-driven**. dashd's `CDeterministicMNManager` bans an MN when it crosses `MAX_PoSe_PENALTY`; that ban never appears as a special tx in any block. `apply_block()` walks special txs, so it can **never** observe one — the PAYEE `MnStateMachine` keeps a genuinely-banned MN at `isValid=true` forever, and `find_expected_payee()` (which correctly filters on `isValid`) deterministically projects the phantom-eligible node. The embedded template's payee then disagrees with dashd, the bridge cross-check fail-closes, the MN set never publishes, and the daemonless arm never becomes viable.

The two axes were both live and never joined:

| axis | source | carries |
|---|---|---|
| PAYEE (`MnStateMachine`) | seeded once from `protx list valid true` | full PoSe triple |
| SML (`CSimplifiedMNList`) | advances on **every** mnlistdiff | authoritative per-entry `isValid` |

The reconciler that fixes exactly this — `sync_validity_from_sml()` at `mn_state_machine.hpp:468` — existed, was unit-tested (`DashMnState.SyncValidityFromSml*`), was correct, and had **zero production callers**. Its own doc block promised the call "after every SML root MATCH and once at startup" — which never happened.

Because the seed uses `protx list valid true`, banned nodes are pre-filtered at startup; the desync only surfaces once a ban lands **after** seeding — which is why it looked height-specific and sparse (live-observed ~h2513489 against the h2513000 anchor).

## The fix — two call sites, both where the maintainer owns both axes

1. **`on_mnlistdiff()`** — after each accepted SML apply, reconcile the payee axis from the freshly-applied SML at `m_sml_current_height`. Catches a ban landing **after** the queue was seeded (the live defect).
2. **`on_mn_list_update()`** — after the payee axis is (re)seeded, reconcile against an already-present SML. Handles the seed-arrives-after-SML ordering (warm-loaded / live-advanced SML), so the join does not depend on a fragile fixed startup line racing the async seed.

Trap-compliance:
- **Only the boolean is reconciled.** The SML wire entry carries only `isValid`, not ban/revive heights; those are bounded by `current_height` per the function's field-ownership contract. No ban-height reconstruction.
- **Position is preserved.** `sync_validity_from_sml()` early-continues on entries whose `isValid` did not flip, so queue position for unchanged nodes is never perturbed; the reconcile never adds/removes entries.
- Guarded on a non-empty SML (an empty set is a gap, handled downstream).

Kept narrow — no state-machine restructuring. The `dashd` RPC fallback path is untouched.

## Tests

Two maintainer-level regressions pin the wiring end-to-end (both were impossible to pass before, since the reconciler was never called):

- `MnlistdiffBanReconcilesOntoPayeeAxis` — a ban arriving as a post-seed mnlistdiff flips the payee entry and drops it from `find_expected_payee` (call site 1).
- `SeedReconcileAppliesAlreadyPresentSmlBan` — an SML ban already applied before the seed is applied at seed time (call site 2).

Local (Debug, system deps): **17/17** `test_dash_coin_state_maintainer` + **30/30** `test_dash_mn_state` green. Log lines confirm firing: `[SML->PAYEE] validity reconcile: -1 banned` and `[SML->PAYEE] seed reconcile: -1 banned`.

## Gate

No self-merge — requesting integrator verify (oracle parity + full Linux x86_64 CI, no false-green) before an operator tap. Full-matrix CI has not run in the authoring sandbox; only the two affected targets were built/run locally.

### DIFF
--- a/src/impl/dash/coin/coin_state_maintainer.hpp
+++ b/src/impl/dash/coin/coin_state_maintainer.hpp
@@ -234,6 +234,24 @@ public:
         // fetch and the first live full-block ingest were the soak's
         // 2-slot cursor lag -> bad-cb-payee at the address-group boundary).
         m_state.mnstates().load(std::move(mnstates), as_of_height);
+        // Startup / reseed join (2026-07-30): the PAYEE axis was just (re)seeded
+        // -- reconcile it against an already-present SML so a warm-loaded or
+        // live-advanced SML's authoritative isValid lands on the fresh queue
+        // immediately, without waiting for the next mnlistdiff. Handles the
+        // seed-arrives-after-SML startup ordering (mirror of the on_mnlistdiff
+        // reconcile, which handles SML-arrives-after-seed). No-op when no SML has
+        // applied yet (m_have_mn_sml false) or nothing flipped. Height falls back
+        // to the seed's as_of_height when no live diff has set m_sml_current_height.
+        if (m_have_mn && m_have_mn_sml) {
+            const uint32_t vh = m_sml_current_height ? m_sml_current_height
+                                                     : as_of_height;
+            const auto vr = m_state.mnstates().sync_validity_from_sml(
+                m_state.sml(), vh);
+            if (vr.flipped_to_invalid || vr.flipped_to_valid)
+                LOG_INFO << "[SML->PAYEE] seed reconcile: -"
+                         << vr.flipped_to_invalid << " banned +"
+                         << vr.flipped_to_valid << " revived @ h=" << vh;
+        }
         if (!m_have_mn)
             demote();
         else
@@ -418,6 +436,30 @@ public:
                  << " MNs; quorums +" << q_added << " -" << q_deleted
                  << " => " << m_state.qmgr().active_count()
                  << " active; have_sml=" << (m_have_mn_sml ? "yes" : "no");
+
+        // -- PAYEE-axis validity reconcile (Bug 12/14 wiring, 2026-07-30) ------
+        // The SML just advanced and carries the AUTHORITATIVE per-MN isValid.
+        // PoSe bans are CONSENSUS-driven, not tx-driven, so apply_block() can
+        // NEVER observe them -- the PAYEE MnStateMachine keeps a phantom-eligible
+        // banned MN as isValid=true forever and find_expected_payee determinist-
+        // ically projects it (daemonless payee-desync, live-observed ~h2513489).
+        // sync_validity_from_sml() is the reconciler that fixes exactly this and
+        // was DEAD CODE (zero production callers) until this line. It reconciles
+        // the BOOLEAN off the freshly-applied SML (ban/revive heights stay SML-
+        // approximate, bounded by m_sml_current_height -- see the function's
+        // field-ownership contract) and early-continues on entries whose isValid
+        // did NOT flip, so queue POSITION for unchanged nodes is never perturbed.
+        // Guarded on a non-empty SML (an empty set is a gap, handled below).
+        if (m_have_mn_sml) {
+            const auto vr = m_state.mnstates().sync_validity_from_sml(
+                m_state.sml(), m_sml_current_height);
+            if (vr.flipped_to_invalid || vr.flipped_to_valid)
+                LOG_INFO << "[SML->PAYEE] validity reconcile: -"
+                         << vr.flipped_to_invalid << " banned +"
+                         << vr.flipped_to_valid << " revived (scanned "
+                         << vr.scanned << ", matched " << vr.matched
+                         << ") @ h=" << m_sml_current_height;
+        }
         // SML/quorum persistence (SMLDb/QuorumDb): the applied state is now
         // current AT diff.blockHash — flush it so a restart resumes from this
         // tip incrementally instead of a cold mnlistdiff(zero, tip). Only when
--- a/src/impl/dash/coin/mn_state_machine.hpp
+++ b/src/impl/dash/coin/mn_state_machine.hpp
@@ -254,6 +254,20 @@ public:
         int best_h = std::numeric_limits<int>::max();
         for (const auto& [hash, st] : m_entries) {
             if (!st.isValid) continue;
+            // Bug 14 guard: nPoSeBanHeight is write-only on the selection
+            // path, so a divergent persisted snapshot could carry
+            // isValid==true alongside a live ban height and pay a banned MN
+            // on the boolean alone. Reconciled invariant: isValid==true =>
+            // ban height clear (0, or the -1/UINT32_MAX never-banned
+            // sentinel). Fail-closed (a crash here would kill the daemonless
+            // payee projection outright, worse than a skip) with a logged
+            // warning so the invariant violation is never silent.
+            if (sane(st.nPoSeBanHeight) != 0) {
+                LOG_INFO << "[MNS] fail-closed: valid MN " << hash.GetHex().substr(0,16)
+                         << " carries live PoSe banHeight=" << st.nPoSeBanHeight
+                         << " (divergent snapshot) -- excluded from selection";
+                continue;
+            }
             if (st.scriptPayout.m_data != script) continue;
             uint32_t lastPaid = sane(st.nLastPaidHeight);
             uint32_t revived  = sane(st.nPoSeRevivedHeight);
@@ -348,6 +362,14 @@ public:
         };
         for (const auto& [hash, st] : m_entries) {
             if (!st.isValid) continue;
+            // Bug 14 guard (see pick_paid_mn): fail-closed on the write-only
+            // ban field so a divergent snapshot cannot project a banned MN.
+            if (sane_height(st.nPoSeBanHeight) != 0) {
+                LOG_INFO << "[MNS] fail-closed: valid MN " << hash.GetHex().substr(0,16)
+                         << " carries live PoSe banHeight=" << st.nPoSeBanHeight
+                         << " (divergent snapshot) -- excluded from projection";
+                continue;
+            }
             uint32_t lastPaid = sane_height(st.nLastPaidHeight);
             uint32_t revived  = sane_height(st.nPoSeRevivedHeight);
             int h = static_cast<int>(lastPaid);
--- a/test/test_dash_coin_state_maintainer.cpp
+++ b/test/test_dash_coin_state_maintainer.cpp
@@ -734,3 +734,87 @@ TEST(DashCoinStateMaintainer, StaleZeroBaseSnapshotDoesNotCorruptTipSml) {
     EXPECT_EQ(st.sml_current_hash(), raw256(0x60))
         << "post-reorg cold resync must not be blocked by the stale-guard";
 }
+
+// ════════════════════════════════════════════════════════════════════════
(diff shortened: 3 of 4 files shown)


frstrtr:
This is the merge candidate for the daemonless payee wedge. I duplicated it in a separate PR before spotting yours — that was my error for not re-checking the PR list in the ten minutes between your open and mine. Yours landed first, is broader, and I am closing mine in favour of it.

**Where yours is better than mine, and I want that on the record:**

- The selection-path guard covers the write-only-`nPoSeBanHeight` trap in `pick_paid_mn()` *and* `find_expected_payee()`. I only flagged that hazard in prose; you closed it.
- Sentinel-normalising so dashd's `-1`/`UINT32_MAX` never-banned marker is not misread as a live ban. I did not think of this, and it is exactly the kind of thing that would have produced a silent, wrong exclusion — a *worse* failure than the one we are fixing, because it would under-pay eligible nodes with no error surfaced anywhere.
- Making it a logged fail-closed rather than a hard assert. Right call for a consensus path.

**Two things from my branch that yours does not appear to cover.** Both are additive; please fold them in rather than taking my PR:

1. **Revival propagation on a later diff.** I see ban coverage and seed coverage, but no test asserting that an SML revival restores eligibility, clears the recorded ban height, and records the revived height. That last part matters beyond eligibility: the revived height feeds the queue-position key, so a missed revival shifts *payment order*, not just inclusion — and it would under-pay a masternode the network has already restored. My version:

```
m.on_mnlistdiff(diff_with_seed(uint256::ZERO, raw256(0xC1), 2513489, 100'000'000LL, banned));
ASSERT_FALSE(st.mnstates().entries().at(mn).isValid);
m.on_mnlistdiff(diff_with_seed(raw256(0xC1), raw256(0xC2), 2513600, 100'000'000LL, revived));
EXPECT_TRUE (st.mnstates().entries().at(mn).isValid);
EXPECT_EQ   (st.mnstates().entries().at(mn).nPoSeBanHeight, 0u);
EXPECT_EQ   (st.mnstates().entries().at(mn).nPoSeRevivedHeight, 2513600u);
```

2. **An end-to-end assertion on the projection itself**, not just on the stored flag. Seed the doomed masternode with the lowest queue key so it is the unambiguous argmin, assert it *does* win the slot pre-ban (or the test proves nothing), then assert that after the ban `find_expected_payee()` returns the *other* node. That is the production failure stated directly, and it is the assertion that fails loudest if someone later refactors the filter out of selection while leaving the reconciliation intact.

**One question, and it is the one I care most about.** Did you verify these fail *without* the fix? I reverted the wiring, kept the tests, rebuilt, and confirmed 3/3 red — the decisive line being `find_expected_payee()` returning the banned node rather than the healthy one. A regression test that has only ever been observed green is not yet evidence, and this is the exact defect class where a test can pass for the wrong reason. If you have not run that negative pass, please do it before merge and post the output.

**Still open, and not blocking this PR:**

- h=2513168 and h=2513261 are not covered by either of our test sets. If this fix closes them, the single-defect reading holds; if either survives, there is a second independent ordering defect that must surface rather than be masked by a green suite.
- A live cold-start soak on this branch is the only thing that will prove a daemonless node actually publishes a masternode set and serves a real template. Both of our evidence is unit-level.
- The reward-unsafe inversion — serving `h=0` with no payee instead of refusing — remains separate and should not wait on this.

This has gone to Fable for an adversarial advisory review. I will point that review at this PR rather than mine.

frstrtr:
## Adversarial review returned — verdict: **merge this, but do NOT close #984 on it**

Full review ran at high effort against the PR head with the surrounding source, not the diff. Summarising the findings that change what we do next.

### MAJOR: this fixes the live-run half, not the cold-start half — and the incident died on the cold-start half

`MnCheckpointLane` owns a **private, third `MnStateMachine` instance**. It replays checkpoint→tip through `apply_block` and **never consults `m_state.sml()`**. Its own doc says any payee desync during the bridge is *terminal* — the lane fails closed permanently and never publishes.

This PR wires the reconciler into the **maintainer's** machine only.

So a daemonless node cold-starting (or desync-recovering) after a consensus ban landing between the pinned checkpoint height and the tip will still desync mid-replay → terminal fail → never publish → fall into the h=0 fail-open. **That is exactly the path that produced the eleven `arm=dashd-fallback h=0 mn_payee=(none)` serves.**

The case this PR genuinely and correctly fixes is a node that was *already published* when the ban arrived: the flip lands before the banned node reaches the queue head, projection matches dashd, and lastPaid attribution stays in sync. Real value — but the title "fixes daemonless payee-desync" overclaims.

And note the shape: a reconciler that exists but isn't invoked **on the third machine that needs it**. Same class again.

**Required follow-up before #984 closes:** reconcile the checkpoint lane, or apply a lane-level SML validity mask at seed time.

### The two other heights probably are a second defect

**h=2513168 and h=2513261 both PRECEDE the ban at h=2513418.** Whatever diverged there was not *this* ban. I should have caught that arithmetic myself.

Code alone can't adjudicate. Check both heights against dashd's protx ban/revive events — **if either has no corresponding PoSe event, that is the second independent defect.**

### Negative pass — static, not executed, and honestly labelled

The reviewer could not build (its sandbox rejects all non-git filesystem writes). Instead it traced each test's assertions with the hunks mentally reverted, helper by helper, and confirmed all four fail without their respective hunk — including that each test isolates its own call site. It verified `single_mn` keys the entry to match `sml_entry`, and that `diff_with_seed` carries a parseable type-5 cbTx so `m_sml_current_height` is genuinely set.

That matches my empirical 3/3-fail result on the closed duplicate branch. It explicitly flags this is **not** a green-red run and asks for one on a machine that can build. **That request stands — please run the maintainer suite with the two `coin_state_maintainer.hpp` hunks reverted and post the output.**

### Attack vectors — mostly clean

- **Sentinel normalisation:** the guard's normalisation is exactly consistent with the queue-position scoring — same lambda, same constant. Neither silent-failure direction exists in the guard. But `sync_validity_from_sml` itself does **not** normalise: a sentinel-carrying entry would keep `UINT32_MAX` through a ban flip and never get its revive height bumped. **Latent, not live** — every ingress derives `isValid` from the ban height, and the one path that could load a divergent pair is unwired.
- **Selection guard:** no reachable state where `isValid==true` + live ban height coexist for a node dashd would pay. Pure defence-in-depth. Fail-closed over assert confirmed correct — a crash takes down all lanes, a skip degrades one projection loudly.
- **Ban-height precedence:** sound. The upper bound has no magnitude consumers — `nPoSeBanHeight` is only ever compared to 0. Every revive clears it, so a re-ban re-stamps fresh. No stale-precise-height trap.
- **Revival:** propagates, clears, bumps monotonically. Error direction is the safe one — our approximation is ≥ dashd's, so the node sits *further back*, and the failure mode is a visible desync, never a silent wrong payment.

### Counterexample to my scope claim — I was wrong

"Cannot alter behaviour for any node whose eligibility did not change" is **false** in one reachable window: a **re-seed while the SML lags the seed**. The RPC re-seed is fetched at dashd's live tip Y; if the SML is current at X < Y and an MN was revived in (X, Y], the seed correctly lists it valid and the seed-join reconcile flips it **invalid** off the stale SML — a node whose real eligibility never changed, silently excluded until the SML catches up, with its exact RPC-provided revive height then clobbered by the approximation.

Availability-class, not payment-class, and the window is one diff cadence. Cheap hardening: **skip the seed-join reconcile when `m_sml_current_height < as_of_height`.**

### Three fixes worth folding in

1. **Move the guard after the script match in `pick_paid_mn`**, or rate-limit its log. `find_paid_in_block_first` calls `pick_paid_mn` once per coinbase output, so one persistent divergent entry emits 10+ INFO lines per block, forever.
2. **Fix the warm-restart comment.** The warm SML load in `main_dash.cpp:2634-2637` writes `node_coin_state` directly and never sets the maintainer's `m_have_mn_sml`, so the seed-join reconcile is silently skipped on exactly the case its comment claims to handle. It's *accidentally* safe — per the counterexample above, reconciling there would be dangerous — so fix the comment, not the guard.
3. **Fix the blast-radius comment.** It claims worst case is "a transient under-payment of 1-2 blocks"; the true worst case is a desync-triggered wipe/re-seed cycle.

### Instance #10, found incidentally

**`MnStateDb`** — the payee-axis persistence store, with atomic-batch writes, a BEST sentinel and round-trip serialisation tests — **is instantiated nowhere.** The state-machine comment even describes main_dash wiping it, which main_dash never does. Found by a one-line grep while checking guard reachability.

That also means the persisted-snapshot divergence I asked about **does not exist**: the payee axis is never persisted or restored.

### The mechanical detector

Proposed and worth doing across all lanes: for every public method/class under `src/impl/*/coin/`, grep the non-test tree for its name excluding the defining file; zero hits → flag. Crude at the text level, but this codebase's distinctive snake_case names and header-only components make it unusually reliable — a ~20-line CI script. Complementary structural fix is what this PR's tests already model: **assert the call, not the callee.**

frstrtr:
## Green-red proof of the regression tests (integrator ask)

Method: on HEAD `72a5ff01`, reverse-applied **only** `abd1b08b`s source hunk to
`src/impl/dash/coin/coin_state_maintainer.hpp` (strips both `sync_validity_from_sml()`
call sites), leaving the two new tests in place. Rebuilt, ran, then restored + rebuilt.
Fixup `72a5ff01` does not touch this file, so the reverse-apply is clean.

### 1. GREEN @ HEAD (wiring present)
```
[==========] 17 tests from 1 test suite ran.
[  PASSED  ] 17 tests.
```

### 2. RED (wiring stripped, tests intact) — the tests bite
```
[  FAILED  ] DashCoinStateMaintainer.MnlistdiffBanReconcilesOntoPayeeAxis (0 ms)
[  FAILED  ] DashCoinStateMaintainer.SeedReconcileAppliesAlreadyPresentSmlBan (0 ms)
[  PASSED  ] 15 tests.
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] DashCoinStateMaintainer.MnlistdiffBanReconcilesOntoPayeeAxis
[  FAILED  ] DashCoinStateMaintainer.SeedReconcileAppliesAlreadyPresentSmlBan

test_dash_coin_state_maintainer.cpp:786: Failure
Value of: st.mnstates().entries().at(raw256(0x01)).isValid
  Actual: true   Expected: false
  SML ban must reconcile onto the payee axis (dead-code wiring)
test_dash_coin_state_maintainer.cpp:788: Failure
Value of: st.mnstates().find_expected_payee().has_value()
  Actual: true   Expected: false
  a banned MN must never be projected as expected payee
```
Both failures reproduce the exact live defect: banned MN stays `isValid=true`
and `find_expected_payee()` projects a phantom. Only the two new tests fail;
the other 15 are unaffected.

### 3. GREEN restored (wiring back)
```
[==========] 17 tests from 1 test suite ran.
[  PASSED  ] 17 tests.
```

Clear to merge on this axis.

PR #989, opened 2026-07-31
feat(dgb): coin-peer-manager seed-gate behind --coin-p2p-discover (opt-in, off by default)

frstrtr:
## Summary
Ports the dash coin-peer-manager seed-gate into DGB as a new in-fence header
`src/impl/dgb/coin/coin_peer_manager.hpp`, wired into `main_dgb.cpp` behind a
new **`--coin-p2p-discover`** flag. The flag is **OPT-IN / off by default** —
`DEFAULT_BOOTSTRAP_HOSTS` behavior is unchanged when the flag is absent.

## Scope (3 files, all in the DGB fence)
- `src/impl/dgb/coin/coin_peer_manager.hpp` (new, ported from dash)
- `src/impl/dgb/coin/CMakeLists.txt` (+2, wire the new header)
- `src/c2pool/main_dgb.cpp` (+43/-3, flag + seam)

No changes to shared base, `src/impl/doge/`, or any other coin.

## Static review (author)
Compile + scope + static pass: PASS. GPG-signed (Good signature).
The 3-file fence holds and the `DEFAULT_BOOTSTRAP_HOSTS` loop is untouched.

## Merge gate vs on-gate — explicit distinction
- **CI is the missing evidence** for this change; that is what this PR exists
  to surface. The static pass is not sufficient on its own.
- The DGB testnet **discovery soak is NOT a merge gate** for a flag that is
  off by default — nothing changes at runtime unless an operator passes
  `--coin-p2p-discover`.
- The soak **IS a gate before anyone recommends turning the flag ON**. It is
  running in parallel; result will be posted as a comment on this PR.

## CI note
Expect `Initialize CodeQL` to fail on egress — that is the fleet-wide blocker
(ci-steward tracked separately), NOT this diff.

### DIFF
--- a/src/c2pool/main_dgb.cpp
+++ b/src/c2pool/main_dgb.cpp
@@ -54,6 +54,8 @@
 #include <impl/dgb/coin/p2p_node.hpp>
 #include <impl/dgb/coin/rpc.hpp>        // NodeRPC — external-daemon submitblock arm (#82)
 #include <impl/dgb/coin/rpc_conf.hpp>   // digibyte.conf creds resolution (rpcpassword off argv)
+#include <impl/dgb/coin/coin_peer_manager.hpp> // dgb::coin::DgbCoinPeerManager — DGB-ISOLATED scored/diverse peer discovery (--coin-p2p-discover; network-standalone gate)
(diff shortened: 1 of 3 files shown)


frstrtr:

PR #994, opened 2026-07-31
bch(p2p): progress-gate the block-download idle watchdog

frstrtr:
### Problem
The liveness/idle timer was restarted on ANY inbound message. A peer emitting
keepalive chatter (pings, addr, inv) while stalling actual block delivery would
keep resetting the watchdog and never trip the idle limit. The block-download
window then requeued the same heights to the same non-delivering peer forever —
a silent stall with no recovery.

### Fix
Arm the idle watchdog only while block bodies are in flight — it counts
*download progress*, not raw inbound traffic. Keepalive chatter no longer
resets it, so a stalling peer trips the idle limit and the window can requeue
elsewhere. Because the gate is armed only when bodies are outstanding, an
at-tip node with the normal ~10 min gap between BCH blocks never churns
(nothing in flight → nothing to time out).

### Non-overlap with the compact-block merkle backstop (#981)
This is orthogonal to the merkle backstop, not stacked on it. My anchors are
in the idle/progress path (p2p_node.hpp lines 54, 151, 575, 825); the merkle
backstop lives in the compact-block reconstruct paths (~1030-1180). Different
code, different failure mode — one recovers from a stalled download, the other
rejects a mis-reconstructed block. A reviewer should not assume they interact.

### Scope
src/impl/bch only — one coin tree, no shared-base / bitcoin_family change. A
single BCH smoke gate suffices; no cross-coin review needed. New KAT
(idle_watchdog_kat_test.cpp) registered in both COIN_BCH build.yml lists.


frstrtr:
DGB-lane adjudication of the "DGB Phase-B smoke — per-coin pillar gtests" red (run 30609726784, sha 934e728e): **CI-side, not a DGB regression.** Evidence:

1. Failure step annotation is "The operation was canceled." at 4m34s — a cancellation, not a gtest assertion failure. "Check DGB Phase-B source presence" passed and logged "DGB Phase-B sources present — running per-coin smoke"; the smoke step was then canceled mid-run.
2. The same gate is GREEN on master across four runs today (30618614297 @09:05, 30619785080/30619791418/30619799205 @09:23–09:24) and GREEN on other PR branches (dash/t1c-oracle-byte-parity, ci-steward/codeql-bundle-prestage) on the same self-hosted harness.
3. This PR (bch/compact-relay-idle-progress-gate) diff touches **zero** src/impl/dgb / Phase-B / smoke files — a bch-only diff cannot regress the DGB per-coin pillar gtests.

Recommendation: re-run the job; it will go green. @bch-embedded-steward — this red is not yours to carry as a DGB code issue. — dgb-scrypt-steward

frstrtr:
CI triage — the two red legs are Analyze (c-cpp) and Analyze (python), CodeQL run 30609726843. Both are environmental, NOT findings:

- Both jobs ran 4h28m and terminated with the "Initialize CodeQL" step still incomplete; "Set up job" and "Checkout repository" passed, but "Perform CodeQL Analysis" never executed. gh run view --log-failed returns empty because no analysis ran — there is no finding, the bundle download hung and timed out the runner.
- This is the known Initialize/bundle-download failure mode, not a code issue in this diff. Scope is src/impl/bch only plus a 2-line build.yml KAT registration; there is nothing CodeQL-actionable here.
- Structural fix is ci-steward #988 (pinned local CodeQL bundle). Not chasing further.

Re-running the two failed jobs now.

PR #1014, opened 2026-08-01
test(ltc): real Merkle root in compact-block fixture — fixes #1010 CompactBlockTest reds

frstrtr:
Fixes the 2 CompactBlockTest FAILUREs (FullReconstruction, EmptyBlock_OnlyCoinbase) that #1010 un-masked once real test execution was restored.

Root cause: make_block() called m_merkle_root.SetNull(). The #977 BIP152 reconstruction backstop (ltc compact_blocks.hpp:341) recomputes the Merkle root over the reconstructed txs and DISCARDs on mismatch — correct production behaviour. Against a null header root every valid reconstruction mismatched, so complete=false and m_txs empty. The fixture predated the backstop.

Fix (test-only, shared BIP152 fixture): compute the real root via ComputeTxMerkleRoot(txs) so the happy path is actually exercised.

Verified on 288ffd6 (=#1010 head): before, 2 FAILED / 13 passed at exactly the reported lines 237/239, 274/275; after, 15/15 PASSED. Drives the real ltc ReconstructBlock, no mock. #1010 goes green once it carries this.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1481,6 +1481,16 @@ jobs:
             exit 1
           fi
 
+      - name: Upload real-BLS binary (deployable candidate)
+        if: success()
+        uses: actions/upload-artifact@v7
+        with:
+          name: c2pool-dash-linux-x64-realbls
+          path: build_dash_bls/src/c2pool/c2pool-dash
+          retention-days: 3
+          overwrite: true
+          if-no-files-found: error
+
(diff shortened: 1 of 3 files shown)


frstrtr:
Closing as duplicate. The identical fix already merged to master as #1012 (c3bebb72b) at 15:14Z. The reds on #1010 persist only because its head (288ffd6) predates that merge — updating #1010 onto master clears them. No separate change needed.

PR #1021, opened 2026-08-02
dgb: Phase B pool/share-layer port (B1+B2, Scrypt-only) + AutoRatchet threshold seam

frstrtr:
## Phase B pool/share-layer port — c2pool-dgb (DigiByte, Scrypt-only)

Ports the Phase B pool/share layer (B1 + B2) into `src/impl/dgb/`, plus the
`dgb::AutoRatchet` threshold-constant seam (slice #2, SAFE-ADDITIVE). All
consensus/share logic is carried byte-for-byte from `src/impl/ltc/`
(== `p2pool-merged-v36` reference); DGB-Scrypt shares LTC ratchet/PPLNS math,
so these are namespace ports, not re-derivations. **p2pool-merged-v36
compatibility preserved.**

### Isolation verdict (self-verified)
**24 of 26 files under `src/impl/dgb/`.** Zero changes in
`src/impl/bitcoin_family/`, `src/core/`, or any second-coin tree
(ltc / doge / btc / bch / dash). **ISOLATION-CLEAN** — non-exceptional,
single-coin smoke suffices.

### Files (26)
src/impl/dgb/CMakeLists.txt
src/impl/dgb/auto_ratchet.hpp
src/impl/dgb/b1_smoke.cpp
src/impl/dgb/config.hpp
src/impl/dgb/config_coin.cpp
src/impl/dgb/config_coin.hpp
src/impl/dgb/config_pool.cpp
src/impl/dgb/config_pool.hpp
src/impl/dgb/donation_consensus.hpp
src/impl/dgb/messages.hpp
src/impl/dgb/node.cpp
src/impl/dgb/node.hpp
src/impl/dgb/params.hpp
src/impl/dgb/peer.hpp
src/impl/dgb/pool_monitor.hpp
src/impl/dgb/protocol_actual.cpp
src/impl/dgb/protocol_legacy.cpp
src/impl/dgb/redistribute.hpp
src/impl/dgb/share.hpp
src/impl/dgb/share_check.hpp
src/impl/dgb/share_messages.hpp
src/impl/dgb/share_tracker.hpp
src/impl/dgb/share_types.hpp
src/impl/dgb/whale_departure.hpp

### The two workflow files (branch-trigger-allowlist only, moot post-merge)
- `.github/workflows/build.yml` — adds `dgb/pool-pillars-port` to the push + pull_request branch filters so CI actually runs on this branch. Allowlist entry only; moot after merge.
[...]

PR #1022, opened 2026-08-02
dgb: fix silently-dead tx-relay probe in send_shares (#905)

### LINKED ISSUE #905: bch,dgb: tx-forwarding send path is dead code — same requires-probe defect as #875 (ltc)
Found while auditing the DASH p2p message layer. **BCH and DGB carry the identical defect already filed for LTC in #875**, and neither is covered by that issue's scope.

## The defect

Both lanes gate the `remember_tx`/`forget_tx` send block on a `requires`-probe for a **top-level** member that does not exist:

```cpp
if constexpr (requires { obj->m_new_transaction_hashes; })   // ALWAYS FALSE
```

- `src/impl/bch/node.cpp:725`
- `src/impl/dgb/node.cpp:702`

The member is actually **nested inside `ShareTxInfo`**:

- bch: `src/impl/bch/share.hpp:95` and `:106` declare `bch::ShareTxInfo m_tx_info;`; the vector lives at `src/impl/bch/share_types.hpp:116`
- dgb: `src/impl/dgb/share_types.hpp:110`, reached via `m_tx_info`

So the correct expression is `obj->m_tx_info.m_new_transaction_hashes`. The probe as written is false for **every** share variant in both lanes, which means the entire tx-forwarding SEND path is unreachable — these nodes have never forwarded a transaction byte to a peer.

## Cross-lane state (all five verified on master `22e8c692`)

| lane | probe | member location | verdict |
|---|---|---|---|
| btc | `requires { obj->m_tx_info; }` (`node.cpp:778`) | nested | ✅ **already fixed** — use as the reference |
| dash | no probe, direct access | **top-level** (`share.hpp:43`, `:153`) | ✅ correct by construction |
| ltc | top-level probe (`node.cpp:721`) | nested (`share_types.hpp:107`) | ❌ dead — **#875** |
| bch | top-level probe (`node.cpp:725`) | nested (`share.hpp:95`, `:106`) | ❌ dead — **this issue** |
| dgb | top-level probe (`node.cpp:702`) | nested (`share_types.hpp:110`) | ❌ dead — **this issue** |

BTC already carries the fix *and* an explanatory comment at `node.cpp:767` describing the old broken probe. **The fix is a port, not a design.**

## Why it is invisible

`if constexpr` on a false `requires` compiles cleanly and silently discards the branch. No warning, no crash, no counter — the node simply never sends. This is the same reason #875 went unnoticed, and the same class of silent-unreachability as #878/#881.

## Suggested fix

Port BTC's `node.cpp:778` form to both lanes. Coordinate with #875 so ltc/bch/dgb land consistently rather than diverging further.

⚠️ **Do not fix by unconditionally dereferencing `m_tx_info`** — verify per share variant that the member exists before assuming it, and add a test that would FAIL if the probe went false again. A passing test suite proved nothing here precisely because the code under test never ran.

## Scope note

Reward-neutral: this is peer tx relay, not the payout path. But it means BCH/DGB peers never learn our transactions, which affects block propagation and the `known_txs` accounting that #834's wholesale-clear bug also touches.

frstrtr:
## dgb: fix silently-dead tx-relay probe in send_shares (#905)

Fixes #905 — the DGB port of the BTC tx-relay dead-requires-probe fix.

### What was broken
`send_shares()` collected the new-tx hashes each share references so it could `remember_tx`-forward them to a peer. The pre-fix probe guarded on a **top-level** `obj->m_new_transaction_hashes` that **no dgb share type declares** — that field is nested inside `m_tx_info` (`dgb::ShareTxInfo`). The `if constexpr (requires { ... })` therefore compiled to **always-false**, so the collection was empty and DGB **never relayed a tx byte to a peer for any share version**.

### Fix
Introduce `src/impl/dgb/coin/share_tx_relay_refs.hpp` — SSOT `append_share_tx_refs(obj, refs)` that guards on the real nested `m_tx_info`:
- v17/v33 (carry `m_tx_info`) → referenced hashes ARE collected, in order
- v34/v35/v36 + merged-mining (no `m_tx_info`) → compiled out, vacuously empty

`send_shares()` now calls the SSOT instead of the dead top-level probe.

### Isolation verdict
All changes confined to `src/impl/dgb/` + a 2-line `.github/workflows/build.yml` test-allowlist entry (adds `dgb_share_tx_relay_refs_test` to build+run lists — same pattern as #1021).
- **ZERO** hits in `bitcoin_family/`, `src/core/`, or any second coin tree (ltc/btc/bch/doge/dash/nmc).
- `git diff --name-only origin/master...` = 5 files: `coin/share_tx_relay_refs.hpp`, `node.cpp`, `test/CMakeLists.txt`, `test/share_tx_relay_refs_test.cpp`, `build.yml`.

### Test — failing-if-dead regression
`dgb_share_tx_relay_refs_test` (4 cases): v17/v33 collect nested hashes in order; segwit + merged-mining variants vacuously empty. Reverting the SSOT to the old top-level probe collapses the v17/v33 expectations to 0 (the historical dead path).

Local DGB smoke: **built clean, 4/4 PASS.**

Commits GPG-signed (EDDSA key 50AB1379285EFE76, Good). No self-merge — integrator merges on operator approval.


frstrtr:
Closing as SUPERSEDED: this exact fix already landed on master as PR #914 (37515dc7d) — SSOT header src/impl/dgb/coin/share_tx_relay_refs.hpp, the append_share_tx_refs call in node.cpp, and the KAT are all live. A rebase of this branch drops its single commit as previously-applied and yields a zero-file diff. Branch is untouched (no force-push); reopen if this reading is wrong.

PR #1024, opened 2026-08-02
ci(dash-bls): keep the real-BLS binary as an artifact

frstrtr:
## Why

`linux-dash-bls` builds `c2pool-dash` with `C2POOL_DASH_BLS=ON`, runs `dash_bls_linked_guard.sh` against the binary and executes the six BLS KATs — then **throws the binary away**.

That binary is exactly what a daemonless DASH deploy needs. Today it gets rebuilt by hand, and the build host has no conan cache, so "by hand" is not straightforward. Every green run already produces a deployable candidate; we just don't keep it.

## What

One step appended to `linux-dash-bls`:

```yaml
- name: Upload the real-BLS binary
  if: success()
  uses: actions/upload-artifact@v4
  with:
    name: c2pool-dash-real-bls-linux-x86_64
    path: build_dash_bls/src/c2pool/c2pool-dash
    retention-days: 7
    if-no-files-found: error
```

`if-no-files-found: error` is deliberate: a silently-missing binary should fail the job, not publish an empty artifact. Publishing nothing and publishing something are otherwise indistinguishable downstream — the failure mode that cost four days on this same file.

No change to the build, the link guard, or the KATs. Verified the step parses and lands inside `linux-dash-bls` (18 steps, last one is this).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1511,3 +1511,12 @@ jobs:
 
   # Releases are built and published manually (not via CI).
   # See installer/ for packaging scripts (DMG, ZIP, Inno Setup).
+      - name: Upload the real-BLS binary
+        if: success()
+        uses: actions/upload-artifact@v4
+        with:
+          name: c2pool-dash-real-bls-linux-x86_64
+          path: build_dash_bls/src/c2pool/c2pool-dash
+          retention-days: 7
+          if-no-files-found: error
+


frstrtr:
Closing as redundant — verified after the fact.

An upload step for this exact artifact already exists in `linux-dash-bls` and is a better version than the one I proposed:

```yaml
- name: Upload real-BLS binary (deployable candidate)
  uses: actions/upload-artifact@v7      # mine was @v4
  with:
    name: c2pool-dash-linux-x64-realbls
    path: build_dash_bls/src/c2pool/c2pool-dash
    retention-days: 3
    overwrite: true                     # mine lacked this
    if-no-files-found: error
```

And it is working — six binaries published today (01:08, 04:35, 05:37 ×2, 05:41, 06:31), 5MB each.

Why I opened this: I queried the artifacts API earlier, found only `pkg-windows-dash`, and concluded nothing retained the DASH binary. That was true *at the time* — build.yml could not parse, so no `linux-dash-bls` run had produced anything. I then carried that conclusion past #1010 landing, which is the event that invalidated it.

The failure on this PR was also correct behaviour: it only touches build.yml, so the PR path filter skips the build, no binary exists, and `if-no-files-found: error` failed the job rather than publishing an empty artifact. The guard worked on its own author.

Branch retained.

PR #1026, opened 2026-08-02
dash(embedded): MN diff-ladder recovery — daemonless re-seed key for the payee-desync latch

frstrtr:
Implements the storage layer and recovery function from
`DASH_MN_DIFF_LADDER_RECOVERY.md` (frstrtr/the).

## The defect

`coin_state_maintainer.hpp` latches `m_mn_needs_reseed` on a masternode payee
desync. The latch is deliberate and correct — a desynced payee projection would
mint a consensus-invalid coinbase, so refusing is right. The defect is that the
**only key that opens the latch is an authoritative re-seed obtained via the
`protx` coin RPC**. A daemonless node has no coin RPC, so the latch is permanent
and the node is demoted to an arm that does not exist. The contabo smoke rig
(`c2pool-dash-smoke0802`, 2026-08-02) served 639 consecutive empty set-gap
templates over 76 minutes for exactly this reason.

The diffs that could rebuild the state are already received, parsed and
validated — and then discarded, because the store does an atomic full-rewrite
per accepted `mnlistdiff`. This PR retains them.

## (a) The re-arm is behind a flag that DEFAULTS OFF, and why

`--embedded-mn-ladder-reseed`, default **off**.

* The **ladder is written unconditionally**. Retaining data we already receive,
  parse and validate cannot change behaviour, and the write rides the SAME
  `WriteBatch` that already rewrites `'S'`/`'B'` — one atomic commit point, no
  new failure mode.
* What the flag gates is only the **RE-ARM**. With it off, a payee desync
  latches *exactly* as it does on master; the replay is exercised by the KATs
  and by nothing else.
* Two independent reviews are in flight — a test-harness review and a
  minority report — that could change the RECOVERY SEMANTICS. They cannot change the STORAGE layer. Gating
  this way makes the PR merge-safe regardless of what they conclude, and
  flipping the default becomes a separate one-line change.

Every failure path — and the disabled path — lands on **today's** behaviour:
latch stays set, cold `mnlistdiff(zero, tip)` fallback. There is no new way to
serve a template.

## (b) The three open gates, verbatim from the design doc

> **G1 — does replay reproduce the same divergence?**
> If the projection desynced because a diff was *missed*, replaying the diffs we
> hold rebuilds the same wrong queue and the latch re-fires. Two mitigations,
> and the second is unresolved:
>
> - *Gap detection*: step 3 aborts on a missing height rather than skipping.
>   This catches the missed-diff case **provided** we know a height is missing —
>   i.e. `'D'` keys must be dense over the window, and density must be checked,
>   not assumed.
> - *Anchor trust*: `load_verified()` recomputes `merkleRootMNList` and compares
>   it to a root **that was itself supplied by the peer's diff**. A
>   consistently-lying peer produces a self-consistent chain of roots that
>   passes every check. dashd is immune because it derives the root from `cbTx`
>   in a validated block. **We hold headers, not blocks — `cbTx` is not
>   available to us.** This is the same trust exposure the arm already carries
>   today, so the change does not worsen it; but it is the reason this ladder
>   cannot be called "consensus-anchored", and it must not be described as such.
>   *Open: whether a `merkleRootMNList` commitment reachable from the header
>   chain, or the existing checkpoint bridge, can serve as an independent
>   anchor.*
>
> **G2 — retention.**
> dashd keeps diffs indefinitely. We want a bounded window: large enough to
> cover any plausible desync-to-recovery span, small enough that the store does
> not grow without limit. `N` (anchor interval) and the window length are both
> unset pending review. Note the smoke rig's two desync events were 155 heights
> apart, which is a data point of exactly one.
>
> **G3 — does the ladder survive the fail-closed wipe?**
> `load_verified()` wipes the store on corrupt load, and `set_on_sml_clear`
> wipes on reorg/H-1 heal. If those wipes also destroy `'A'`/`'D'`, the ladder
> is unavailable in precisely the situations that need it. Reorg-wipe is
> *correct* (an orphaned-branch ladder is self-consistent and wrong).
> Corrupt-wipe scope needs deciding.

Where this PR lands against them:

* **G1** — density is REQUIRED, not assumed: the replay walks every integer
  height in `(anchor, target]` and aborts on the first one with no `'D'`
  record. It additionally requires base-hash continuity at each step. Anchor
  trust is UNRESOLVED and is documented as such in the store header; the ladder
  is nowhere described as consensus-anchored.
* **G2** — `MN_LADDER_ANCHOR_INTERVAL = 32` and
  `MN_LADDER_RETENTION_HEIGHTS = 4096` are labelled **PENDING REVIEW SIZING**
  placeholders in the source, with the reasoning (and its weakness) written
  next to them. They must be revisited before the default is flipped.
* **G3** — decided conservatively for now and stated in the code: `clear()`
  takes the ladder with it. The reorg case is unambiguous. The corrupt-load
  case is the one the doc leaves open; wiping means a store we have just proven
  untrustworthy cannot supply the re-seed key, at the cost of recovery in
  exactly that case. Flagged for the reviews to revisit.

## (c) Reviews in flight

Both in-flight reviews — the test-harness review and the minority report — bear
on the RECOVERY SEMANTICS — including the recommended values for `N` and the
retention window, and whether strict height-density is the right density rule.
**Those reviews gate flipping the default**, not this PR.

## What is in the change

**Storage** (`sml_quorum_db.hpp`, additive keyspaces):

```
'D' + height(4B BE) -> pack(CSimplifiedMNListDiff) ++ resulting_root(32B)
'A' + height(4B BE) -> pack(vector<entry>) ++ root(32B) ++ block_hash(32B)
```

Big-endian heights so LevelDB key order IS height order. `'S'`/`'B'` and their
full-rewrite path are untouched. The root is a fixed-width SUFFIX because
`CSimplifiedMNListDiff`'s unserializer drains the remaining stream as its opaque
quorum tail — the reader hands it a body slice that stops 32 bytes short. The
anchor additionally records its block hash so the first replayed diff's
`baseBlockHash` can be bound to the anchor rather than trusted to start in the
right place.

**Recovery** — `SMLDb::replay_from_ladder(target_h)`:

1. greatest `'A'` anchor at height `<= target_h`; none -> `anchor-missing`
2. load it, RECOMPUTE the merkle root, mismatch -> `anchor-root-mismatch-at-h`
3. for `h in anchor_h+1 .. target_h`: `'D'@h` MUST EXIST (`diff-gap-at-h`),
   must chain onto the previous block (`diff-base-mismatch-at-h`), apply,
   RECOMPUTE the root, compare to the stored resulting root
   (`diff-root-mismatch-at-h`)
4. success -> the rebuilt state is RETURNED; the function never mutates live
   state and never clears a latch by itself
5. any abort -> no state returned, caller stays on today's path

**Observability** (this is a standing project law — a gate that can silently
refuse must say WHY, where a human looks):

* `classify_decline()` gains `mn-needs-reseed`, checked FIRST because it is a
  strictly more informative cause of the `not-populated` it used to report. The
  smoke rig reported "not-populated" 639 times while the actual cause appeared
  nowhere in the decline classification.
* Every ladder outcome logs its reason: `anchor-missing` /
  `anchor-root-mismatch-at-h` / `diff-gap-at-h` / `diff-root-mismatch-at-h` /
  `ok-replayed-N-diffs`.
* A startup banner reports ladder depth: anchors, diffs, height range, top
  anchor, and the sizing constants in force.

**Anti-mint interlock**: the ladder clears the payee latch only while
`require_seeded_mn_set` is ACTIVE. That requirement is what still makes it
impossible for block-connect alone to arm MN-readiness off an incidental
ProRegTx; without it, clearing the latch would reopen the exact E2d hole the
latch was added to close.

## Honest scope note (a place the design doc is wrong about the code)

The design doc's step 4 says "on success: `mnstates().load(rebuilt)`". That is
not achievable from this ladder and the PR does not claim it. The DIP-0004
Simplified MN List **omits `scriptPayout` and `nLastPaidHeight`** (see
`mn_checkpoint.hpp`'s trust-anchor preamble), so an SML replay cannot
reconstruct the payout-bearing payee queue. Clearing the latch is therefore
NECESSARY-BUT-NOT-SUFFICIENT: `m_have_mn` still requires a non-empty payee set
AND a height-stamped snapshot, both zeroed by the desync wipe. A KAT asserts
precisely this — after a successful ladder re-arm the arm still routes to the
dashd fallback. Closing the payee axis daemonlessly is a separate problem from
closing the SML axis, and this PR closes the SML axis only.

## Tests — each one proven able to fail

19 KATs in `test/test_dash_mn_diff_ladder.cpp`, folded into the already
build.yml-allowlisted `test_dash_embedded_gbt` executable (a fresh standalone
target would register as a NOT_BUILT CTest sentinel that reports green while
never executing a line — the #137 / test_dgb_subsidy class).

Every positive test has a negative twin. Proven by mutation, not asserted:

| mutation | tests that went RED |
|---|---|
| missing `'D'` height `continue`s instead of aborting | `DeletedDiffHeightAborts`, `EndToEndMissingDiffKeepsArmLatched` |
| anchor root verify removed | `CorruptedAnchorRootRefuses` |
| per-height root verify removed | `CorruptedDiffResultRootRefusesAtThatHeight` |
| flag check removed from the re-arm | `FlagOffKeepsTodaysLatchBehaviour` |
| prune call removed | `PruneKeepsTheWindowBounded` |

The prune test was rewritten during that exercise: as first written it wrote
fewer records than the retention window, so it passed with the prune deleted —
a test that could only return one answer. It now overfills the window.

Do not merge on my behalf.

(Sizing constants `MN_LADDER_ANCHOR_INTERVAL=32` / `MN_LADDER_RETENTION_HEIGHTS=4096` are placeholders pending the review recommendations; flipping the flag default is a separate change.)

### DIFF
--- a/src/c2pool/main_dash.cpp
+++ b/src/c2pool/main_dash.cpp
@@ -216,6 +216,17 @@ const char* const kDashMnCheckpointTestnet =
 uint32_t g_mn_bridge_max_blocks =
     dash::coin::MnCheckpointLane::kDefaultMaxBridgeBlocks;
 
+// --embedded-mn-ladder-reseed: ⚠ DEFAULT OFF. Arms the MN diff-ladder RE-ARM
+// (DASH_MN_DIFF_LADDER_RECOVERY.md). The ladder itself ('D'/'A' keyspaces in
+// dash_sml_db) is written unconditionally — retaining the diffs we already
+// receive, parse and validate costs nothing and cannot change behaviour. What
+// this flag gates is whether a payee desync is allowed to attempt the local
+// root-verified replay instead of latching permanently. It is OFF because the
+// RECOVERY SEMANTICS (not the storage) are still under review; flipping the
+// default is a deliberate one-line change, not a side effect of this slice.
+// Same file-scope rationale as g_mn_bridge_max_blocks above.
+bool g_mn_ladder_reseed = false;
+
 // Report the requested sharechain peering topology at run-loop bring-up. Honest
 // about the deferred live bind: a won/seen share does NOT yet cross the wire
 // until the sharechain pool-node leaf lands.
@@ -256,6 +267,7 @@ void print_banner(const char* argv0)
         << "           [--web-port PORT] [--web-host ADDR] [--dashboard-dir PATH]\n"
         << "           [--external-ip ADDR]\n"
         << "           [--embedded-utxo] [--embedded-mainnet] [--embedded-mn-bridge-max N]\n"
+        << "           [--embedded-mn-ladder-reseed]\n"
         << "           [--embedded-oracle-shadow]\n"
         << "           [--oracle-graduation-blocks N] [--oracle-class-coverage K]\n"
         << "           [--give-author PCT] [-f|--fee PCT] [--node-owner-address ADDR]\n"
@@ -290,6 +302,13 @@ void print_banner(const char* argv0)
         << "        are transport only and NEVER move the arm. Even when armed, every\n"
         << "        per-template gate (SML fresh at tip, non-superblock, credit-pool seed\n"
         << "        height, bestCL, MN-payee cursor, DKG plan) fails closed to dashd.\n"
+        << "        --embedded-mn-ladder-reseed (DEFAULT OFF) lets a masternode\n"
+        << "        payee DESYNC attempt a LOCAL root-verified replay of the retained\n"
+        << "        mnlistdiff ladder instead of latching until an authoritative\n"
+        << "        `protx` re-seed a daemonless node cannot obtain. The ladder is\n"
+        << "        recorded either way; this flag only gates the RE-ARM. Every\n"
+        << "        replay failure (missing anchor, a gap at any height, any root\n"
+        << "        mismatch) leaves the latch SET — i.e. exactly today's behaviour.\n"
         << "        --coin-p2p-magic HEX overrides the embedded coin-P2P wire magic\n"
         << "        (default mainnet bf0c6bbd / testnet cee2caff; regtest fcc1b7dc).\n"
         << "        --regtest-force-won-block (regtest E5 harness, fail-closed) drives\n"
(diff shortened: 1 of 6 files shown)


frstrtr:
HOLD — do not merge. Converted to draft pending redesign.

The harness review of DASH_MN_DIFF_LADDER_RECOVERY.md returned work-with-changes with a finding upstream of this PR: **the ladder is built on the wrong axis.**

- The latch at `coin_state_maintainer.hpp:779` fires on payee_desync/gap_detected and is cleared only by `on_mn_list_update()` (:330-333), which requires a height-stamped `vector<pair<uint256, MNState>>` (scriptPayout, nLastPaidHeight, nConsecutivePayments, ...).
- The wire `CSimplifiedMNListEntry` (`vendor/simplifiedmns.hpp:85-119`) carries none of those fields, and `mn_checkpoint.hpp:13-16` already states in-tree that mnlistdiff omits scriptPayout/nLastPaidHeight and neither is committed in merkleRootMNList.
- Therefore retaining and replaying accepted mnlistdiffs rebuilds the **SML** axis, not the payee vector the latch needs. This PR cannot open the latch it targets.
- Second, independent: `set_on_mn_reseed` is wired only inside the `if (rpc)` branch (`main_dash.cpp:3463`); the daemonless `else` (:3467) wires no reseed at all, so even a correct payee ladder would have no caller.
- Also unsettled: the 639/639 h=0 set-gap rig may not have been the payee latch at all (arm apparently never armed before the first desync at h=2514874). A grep of the preserved artifact for `[MN-CKPT]` status lines vs `[EMB-DASH] MN payee queue DESYNC` settles which blocker actually fired — that check runs before any redesign.

The retained-diff storage layer and its KATs are not wasted — they are the right primitive for the SML axis. Redesign will split by axis (SML: per-diff DIP-4 client verification, no ladder; payee: MNState anchors via the dead MnStateDb or the 18-field checkpoint text format + runtime-armable MnCheckpointLane wired into the unset daemonless reseed seam). Held here until that split lands.

frstrtr:
**Premise refuted — holding as draft, to be closed once the replacement lands.**

This PR built recovery for the SML axis. The latch it aimed to open is on the **payee** axis, and the two do not meet: DIP-0004's Simplified MN List omits `scriptPayout` and `nLastPaidHeight`, so replaying accepted `mnlistdiff`s can never rebuild the payout-bearing queue. The SML axis was never latched in the first place — it already self-heals via `m_on_full_resync` → `getmnlistd(ZERO, tip)`.

The originating evidence was also misread. The smoke rig that motivated this **never armed**:

```
09:49:57  [MN-CKPT] bridge START: replaying h=2513001..2515025 (2025 blocks)
09:51:29  [MN-CKPT] FAIL-CLOSED at h=2514874
09:59:15  first "template sourced" request      <- EIGHT MINUTES LATER
          arm=EMBEDDED for the whole run: 0
```

There was no working payee set for a latch to wipe. The 639 empty templates are downstream of a bridge-replay divergence.

**Actual root cause**, proven against mainnet (`dash-cli`, Dash Core 23.1.7):

```
protx list valid false 2513000   2068 entries   91c8c62b...  PRESENT
protx list valid false 2514874   2059 entries   91c8c62b...  ABSENT (still REGISTERED)

chain's valid set over the window   2068 -> 2059   FELL by 9
our replay's set over the window    2067 -> 2070   CLIMBED by 3
```

PoSe bans are **not transactions**. No special transaction says "banned" — validity is consensus-computed. The bridge replays block special-transactions, so it sees every add and no PoSe remove, and projected a masternode that had been banned inside the window.

The fix is smaller than this PR: `CSimplifiedMNListEntry::isValid` is already on the wire (`vendor/simplifiedmns.hpp:104`, serialized at `:131`/`:153`) and already persisted. The forward replay must consult it and apply **removals** as well as additions.

**Salvage before closing** — three things here are worth keeping regardless of which fix lands:

1. `classify_decline()` gaining `mn-needs-reseed` — observability, wanted either way.
2. The codec constraint: `CSimplifiedMNListDiff::Unserialize` drains the remaining stream as its opaque quorum tail, so a trailing field must be read by fixed-width slicing or it is silently swallowed. A KAT pins this.
3. The mutation methodology — five mutations each proven to turn the correct test red, including a prune test that the mutation pass exposed as unable to fail and which was rewritten.

Also still open and independent of all of the above: `set_on_mn_reseed` is wired only inside the `if (rpc)` branch of `main_dash.cpp`. The daemonless `else` arms the checkpoint lane and wires no reseed callback at all, so nothing is invoked on a desync there.

**Wording note for anything downstream:** "header-anchored (SPV)" for the SML, "trusted self-anchor, progressively falsified by coinbase cross-check" for the payee. Never "consensus-anchored" — there is no independent anchor for the payout-bearing set.

frstrtr:
Closing: premise refuted, not superseded.

The SML axis this PR rebuilds was never latched — it already self-heals via `m_on_full_resync` → `getmnlistd(ZERO, tip)`. The latch is on the payee axis, which DIP-0004's Simplified MN List cannot reconstruct (it omits `scriptPayout`/`nLastPaidHeight`). Root cause is in the preceding comment: the checkpoint bridge replay applies additions and never PoSe removals, so it projected a masternode banned inside the replay window.

Branch `dash-mn-diff-ladder` @ `a2ad9ebc` is retained. The codec KAT (`CSimplifiedMNListDiff::Unserialize` drains the remaining stream, so a trailing field must be read by fixed-width slicing) and the `classify_decline()` `mn-needs-reseed` addition will be lifted from here into the PoSe-removal work rather than rewritten.

Not merged. Reopenable if the payee axis later turns out to want a retained-diff store on a properly anchored basis.

PR #1082, opened 2026-08-04
dash: hold a bounded pool of concurrent coin-P2P peers, not one

frstrtr:
## Why this is the breakeven, not an optimisation

The daemonless template arm must include DKG quorum commitments (`qfcommit`) in the blocks it builds. Those objects live **only** in each full node's in-memory `minableCommitments` map until some miner mines them. Upstream announces each one **exactly once**, at DKG finalize — `RelayInv(MSG_QUORUM_FINAL_COMMITMENT, SerializeHash(fqc))` — and serves it on getdata **by commitment hash only**. There is no request keyed by `(llmqType, quorumHash)`, and the inv hash is a digest of the whole signed object, so it cannot be derived.

**If you did not witness the announcement, you cannot obtain the object.** `MSG_CLSIG` (29) has the identical shape.

dashd has exactly the same hole. It survives it by holding 8+ persistent outbound peers for weeks, so it essentially never misses an announcement — and *that* is what makes its "I hold no commitment for this slot, therefore mine null" trustworthy.

We held **one** peer (`p2p_client.hpp:324`, `std::unique_ptr<Connection> m_peer`). At mainnet block 1520106 that produced 32 null-served rotated slots against dashd's 29 real ones and a diverged `merkleRootQuorums`, because "we didn't hear it" was inferred from a single witness. `dkg_commitments.hpp` already states the governing rule: *absence of a relayed qfcommit is NOT evidence that dashd's commitment is null — absence is not a vote.*

So this PR converts "I didn't hear it" from a guess into evidence. Miss probability falls geometrically in the peer count (~`1-(1-p)^N`). It is the precondition for ever safely porting dashd's null-commitment arm, which is the only remaining path to serving every height. **It is not a bandwidth change and should not be reviewed as one.**

## What landed

**1. A bounded pool of concurrent peers.** Default 8 (what dashd itself keeps), hard cap 16, `--coin-p2p-peers N`. Candidate *selection* is unchanged and still `CoinPeerManager`'s job — scoring, `/16` group diversity, backoff. The pool only decides how many slots are open and skips targets it already holds or is already dialing. Two supporting wiring fixes:

- `DashPeerManagerConfig::max_concurrent_connections` shipped at **3**, so the manager could never propose enough targets to fill an 8-peer pool no matter how many peers it knew. Raised to the pool target.
- The 60s dial-plan refresh now passes the peers we **actually hold** (not just the pinned ones) as the already-connected set, so the scorer stops proposing peers we are connected to *and* its `/16` diversity accounting is computed against the live pool. Without this, eight peers could legitimately end up being eight sockets into one datacentre.

**2. Per-peer state (`PeerSession`).** Liveness policy, ping nonce, handshake tracker, pre-handshake deadline, version metadata, and the unhandled-command WARNING set are all per connection. This is the part where a pool goes silently wrong: a shared `m_ping_nonce` means a pong from peer A closes peer B's outstanding ping, so a dead peer is held "alive" indefinitely by its healthy neighbours — the same class of bug the 100s-idle-timeout fix removed, but harder to see because the peer *count* still looks right. There is a KAT for exactly that.

The scheduling **grain** is shared (one repeating pool tick); the **decisions** are not. `PeerLiveness::tick()` is a pure function of that peer's own timestamps, so evaluating eight of them from one timer is equivalent to eight timers with one timer's worth of asio state. It also means no timer is ever destroyed from inside its own completion handler. Keepalive semantics (120s ping / 1200s unanswered-ping drop) are byte-identical; only punctuality changes.

**3. Inbound fan-in dedup, bounded — eviction stated and proven.** `InvDedup` admits the first `(type, hash)` and suppresses the rest, so one announcement seen by N peers yields ONE getdata rather than N.

**Eviction bound, explicitly:** two independent bounds, both asserted by tests, neither left as a comment.
- **Capacity: 4096 entries**, strict-FIFO eviction from the oldest end. `size()` can never exceed `capacity()` after `admit()`.
- **TTL: 600s.** Entries are inserted in non-decreasing time order, so expiry is a bounded `pop_front` walk.

Worst-case residency is therefore capacity entries (~36 bytes each), reached and held, never exceeded. `DashInvDedup.long_running_arrival_stream_is_bounded_not_a_leak` feeds 200,000 distinct announcements and asserts the resident set stays at 4096 and that the eviction queue and membership index never desynchronise — that gap *is* the leak, so it is checked directly rather than inferred. Both bounds are configurable (`configure_inv_dedup`), capacity floored at 1 so dedup cannot be configured off.

**Accepted trade-off, stated plainly:** suppression means an object is pulled from ONE peer, so if that peer answers with nothing we do not re-ask a different announcer. That is the **same** exposure the single-peer client already had, not a new one — dashd serves what it has just announced straight out of its own in-memory map, and the one routinely-observed `notfound` case (a superseded ChainLock) is already handled as benign and self-correcting. Re-asking a second announcer would require knowing whether the object ever arrived, and for qfcommit/clsig the inv hash is a digest of the whole signed object, so there is no key to reconcile an arrival against. Deliberately not built.

Non-actionable inv types (MSG_TX) do not consume dedup slots, so a busy mempool cannot evict the qfcommit/clsig keys we care about.

**4. Demux and routing.** `core/socket.cpp:321` already tags every inbound message — and every error — with its own socket's `m_addr`, so the delivering peer is known exactly. A message from a socket we do not hold is dropped rather than misattributed. A **PRIMARY** peer carries every request/response leg (`getheaders` / `getmnlistd` / `getqrinfo` / `govsync` / block getdata) so the `ReplyMatcher` deferral state on that Connection stays coherent and a reply is always matched to the peer that was asked. Non-primary peers are witnesses: they deliver inv-pushed traffic and are pulled from, but are never asked a stateful question. Losing the primary promotes a survivor (oldest handshaked) and re-fires `on_handshake_complete` — which is exactly the re-ask the single-peer client performed on every reconnect, so no sync leg silently loses its driver.

`getaddr` is the one deliberately broadcast leg: an addr reply carries no per-peer matching state, and address breadth is what refills the pool after a loss.

**5. Won-block relay — BROADCAST TO ALL, and this is the policy.** This is the money path, so it is decided here rather than inferred. `submit_block_p2p_raw` writes the block to **every handshaked peer** and returns the count. The two failure modes are not symmetric:

- Relaying a found block to N peers costs N block-sized sends and is protocol-wise a non-event — every node we reach forwards it anyway, and one that already has it ignores the duplicate. **Duplicate submission is success.**
- Relaying to ONE peer that happens to be slow, forked, or half-open loses the block outright. **There is no retry: the share is already spent.**

So we buy redundancy with bandwidth we do not care about, on the one message per day where it matters. No peer selection, no "best peer" heuristic, no primary-only shortcut. The return count is honest — 0 means the block reached no peer and the caller must fall back to the submitblock-RPC arm, which is what the dual-path broadcaster's NEVER-SILENT-DROP contract depends on.

**6. Graceful degradation.** Losing peers — including the primary — never stops the node. Survivors are untouched, a survivor takes over the request legs, and the pool refills on the 30s tick and on every 60s dial-plan refresh. In-flight dial slots are reclaimed after 60s with no callback, so a black-holed SYN cannot silently cap the pool below target forever.

## Measurement

A `POOL-STATUS` line every 60s, so a soak reads peer count and durability directly instead of inferring them. Captured from a live run of the real tick loop over a real `io_context`:

```
[COIN-P2P] POOL-STATUS connected=4/4 handshaked=4 distinct_addrs=4 dialing=0
  primary=192.0.2.1:19999 sessions(started/lost)=4/0
  ages=[192.0.2.1:19999=1s* 192.0.2.2:19999=0s 192.0.2.3:19999=0s 192.0.2.4:19999=0s]
  inv_dedup(size/cap)=0/4096 admitted=0 suppressed=0 evicted(cap/ttl)=0/0
[COIN-P2P] peer 192.0.2.4:19999 disconnected after 3s: peer timeout: no message
  received in 3s — pool 3/4
```

`distinct_addrs` is published separately from `connected` on purpose: "3 connections, 1 address" is precisely the failure the single-connection dial rotation produced unnoticed for 9 hours, and nothing named it. `suppressed` is the direct count of duplicate announcements the pool bought us. `qfcommit` and `chainlock` log lines now name the **delivering peer**, so over a soak the distribution of that field shows the pool doing its job rather than eight sockets shadowing one.

The dashboard topology JSON now publishes `peers` (real count), `peers_handshaked`, `peers_target`, `peers_distinct_addrs`, and takes the sync target from the pool's **monotone best** advertised height rather than one peer's — a lagging pool member must not walk the progress gauge backwards.

## Tests, and how failure on master was verified

New TU `test/test_dash_coin_p2p_pool.cpp`, folded into the **existing allowlisted** `test_dash_p2p_node` target (no new test target, no `build.yml` edit, nothing BLS-gated). `python3 tools/ci/check_test_target_allowlist.py` → *"CI drift-guard OK: 229 per-coin test target(s) across 6 coin lane(s) all present"*.

**MEASURED — `test_dash_p2p_node`: 67/67 pass on this branch** (28 new + the 39 pre-existing, which are unchanged and still green).

**MEASURED — the two load-bearing properties were built and RUN against unmodified master and FAIL there.** A throwaway `PoolBaseline` TU was compiled into master's own `test_dash_p2p_node` and executed:

```
[ RUN      ] PoolBaseline.two_peers_are_held_concurrently
  Value of: r.cl.is_handshake_complete()  Actual: false  Expected: true
  attaching a second peer destroyed the first peer's session --
  the client structurally holds ONE connection
[ RUN      ] PoolBaseline.repeated_block_inv_is_acted_on_once
  Expected equality:  fired: 3   vs   1
  the same inv was acted on once per announcement -- with N peers
  that is N getdata for one object
[  FAILED  ] 2 tests
```

That covers *"N peers connect concurrently"* and *"the same inv from 3 peers issues exactly ONE getdata"* as executed measurements, not claims. **INFERRED (structural, not run):** the remaining three — silent-peer reaping with survivors intact, dedup-set eviction, and pool refill — are not expressible against master at all: master holds one `PeerLiveness`, one `HandshakeTracker` and one ping nonce so there is no second peer whose ping could be confused; there is no seen-set to bound (no `InvDedup` type in the master TU); and the reconnect tick is guarded on `!m_peer` (`p2p_client.hpp:446`), i.e. it dials only while holding **zero** peers, so refill-to-N cannot be represented. Master has no peer-count surface of any kind to assert against.

New coverage:
- **Concurrency** — 8 peers held at once, all handshaked, 8 distinct addresses, exactly one primary; hard cap enforced; duplicate endpoints refused.
- **Per-peer isolation** — peer 2's pong carrying peer 1's nonce satisfies neither; peer 2 answering its own ping does not close peer 1's; peer 1's inbound traffic does not refresh peer 2's deadline; per-peer version metadata and unhandled-command sets do not bleed; a message from an unattached socket is dropped, not misattributed.
- **Independent reaping** — a fully silent peer is dropped on its own unanswered-ping deadline while the other two stay connected, keep matching pongs, and keep dispatching new invs.
- **Primary loss** — promotes a survivor and re-kicks the sync legs exactly once; losing the last peer leaves the node alive and unanswerable, not crashed; `error()` removes exactly one peer and double-fires/strangers are no-ops.
- **Fan-in** — 3 peers announcing one qfcommit produce exactly 1 outbound message, admitted=1/suppressed=2, and the getdata goes to the first announcer; the same for block invs firing `new_block` once; a genuinely new object is never suppressed.
- **Dedup bounds** — capacity FIFO eviction with the oldest-12-gone / newest-8-held check and re-admittability of evicted keys; TTL expiry reclaiming a whole 500-entry generation; the 200k-arrival leak run; capacity floored at 1; end-to-end through the client.
- **Refill** — initial connect opens the pool to target and starts at the plan's first entry (the pinned local dashd, which must not be reachable only after a full lap — this was a real bug found by the test); a full pool issues no dial; losing one peer refills exactly one slot with a target we do not already hold; empty cold-start plan arms without wedging and dials the moment seeds land.
- **Relay** — broadcast reaches every handshaked peer and not the mid-handshake one; returns 0 (not success) when no peer is answerable.
- **Routing** — stateful legs land only on the primary; `getaddr` fans out; addr records from a witness peer feed discovery.
- **Read-loop preservation** — a throwing subscriber costs one message on one peer, never the pool.

## Update: the AsAN+UBSan red was a wall-clock race — fixed by injecting the clock, not by widening tolerance

First push went red on `Linux x86_64 (AsAN+UBSan)` with **no sanitizer finding at all**: 2937/2939 passed and the two failures were my liveness tests measuring elapsed real milliseconds — `Expected: (elapsed) >= (3000), actual: 2999` (off by one millisecond) and a ping count of 5 vs 6. Under AddressSanitizer the binary runs 2-10x slower, so any assertion racing the wall clock eventually loses.

**Widening the tolerance would have been the wrong fix.** `>= 2990`, or a ranged ping count, moves the flake rather than removing it — and leaves a test that cannot distinguish *"the deadline logic regressed"* from *"the runner was busy"*. That is the check-that-cannot-fail pattern.

**The clock is now injected**, matching the seam already established by `CoinStateMaintainer::set_now_fn`:

- `CoinClient::now_sec()` — the ONE time source every deadline is measured against (handshake deadline, ping interval, unanswered-ping drop, dial-slot reclamation) — is now an injectable `std::function`, defaulting to the steady clock. `set_now_fn()` overrides it; `tick_for_test()` runs one pool tick synchronously.
- The liveness tests advance **simulated seconds one at a time, running the tick on each** — the exact cadence the real repeating timer produces, since a single jump would *not* be equivalent for the ping cadence — and assert on the resulting **decisions**.

Because simulated hours are free, the assertions are now made against the **shipped** constants rather than scaled stand-ins, and they are **exact** rather than ranged:

- a quiet peer answering every ping over two simulated hours sends exactly `7200/120 = 60` pings, matches 60 pongs, and is never dropped;
- a fully silent peer is **still held at T+1199 and gone at T+1200** — both halves asserted, which is precisely what a widened tolerance would have destroyed;
- in the pool, peer 3 is reaped at exactly its own T+1200 while peers 1 and 2 survive with exactly 10 pings each, all answered.

The same defect was present in two of my new pool tests (a 2.5s and a 6s `io_context` run); both are converted. **No real-timer drive remains in either TU except one deliberately isolated case**, as requested:

`DashCoinP2PClient.pool_tick_timer_is_actually_armed_on_the_io_context`. It exists because every clock-driven test calls `tick_for_test()` itself and would therefore still pass if `ensure_pool_timer()` were deleted and the tick never armed at all. It is written as a **liveness** assertion, not a timing one: it stops the moment the first ping appears, with a 20s watchdog against a 1s cadence. A slower machine makes it take longer; it can only fail if the tick is genuinely never armed.

**MEASURED evidence:**

| check | result |
|---|---|
| `test_dash_p2p_node` under `-fsanitize=address,undefined` | **68/68 pass, zero sanitizer findings** |
| Repeated 20x under 24 busy loops on 8 cores, ASan+UBSan | **20/20 pass, 0 failures** |
| `silent_peer_is_dropped_at_exactly_the_peer_timeout` | 3000ms of real waiting → **2ms** |
| `silent_peer_is_dropped_while_the_others_keep_delivering` | 6000ms → **6ms** |
| `quiet_peer_that_answers_pings_survives_indefinitely` | 6000ms → **18ms** (and now covers 2 simulated hours, not 6 real seconds) |
| the one real-time test, worst case under 3x CPU oversubscription | **2017ms against a 20s watchdog** — 10x headroom |

## Scope, honesty, and what is not proven

- **DASH lane only.** `p2p_client.hpp`, `won_block_dispatch.hpp` (comments + the relay-policy contract), `main_dash.cpp` wiring, and the test files. No shared seam changed; `core/` is untouched. No other coin's `submit_block_p2p_raw` is affected — only `main_dash.cpp` calls the DASH one, and it ignores the (new, additive) return value.
- **Reliable `qfcommit` acquisition is UNPROVEN until a soak observes it.** The KATs pin the mechanism; they cannot pin the outcome. The transport in the tests is direct `handle()` delivery, not live TCP — everything above the socket is the real code, but the wire itself is not exercised here. What *is* already proven on a live wire (soak0804a, heights 2516062/2516063) is that a held connection acquires and BLS-verifies a ChainLock within 7ms of the announcement. Acquisition works; being connected when the announcement happens was the remaining variable, and that is what this changes. **The acceptance evidence is qfcommit arrivals becoming reliable in a soak — not that bandwidth rose.**
- **Design fork, surfaced rather than decided silently:** the dedup set is **one shared set**, not one per peer. A per-peer set would defeat the purpose entirely — each peer's announcement would be admitted separately and we would be back to N getdata. The only real alternative is a shared set *plus* per-object announcer tracking so a second announcer can be re-asked when the first does not deliver; I did not build it, for the reason given in the trade-off above (no key exists to reconcile a qfcommit/clsig arrival against). Recommendation: keep the shared set; revisit only if a soak shows getdata going unanswered.
- **Landed as one PR deliberately.** It is sizeable (~1.2k lines, mostly one file plus tests) but not separable in a way that is *safer*: a pool without per-peer state isolation, or with an unbounded dedup set, would be worse than the single peer we have now — it would look like progress while introducing a cross-peer liveness bug or a slow leak on a node that runs for weeks. Splitting would mean landing exactly one of those intermediate states.


frstrtr:
**Honest restatement of what this PR buys — its stated justification does not survive measurement.**

I argued this change on acquisition coverage: `qfcommit` and `clsig` are announced once, so miss probability should fall geometrically in peer count, `P(witness) = 1 − (1−p)^N`. **That formula only pays when `p < 1`, and I never checked whether it was.**

**It is not.** dashcore v23.1.7, the *inbound peer* path — `CQuorumBlockProcessor::ProcessMessage`, `blockprocessor.cpp:159-162`, after full BLS `qc.Verify(..., checkSigs=true)`:

```cpp
if (auto inv_opt = AddMineableCommitment(qc)) {
    ret.m_inventory.emplace_back(inv_opt.value());
}
```

and the caller, `net_processing.cpp:3638-3640`:

```cpp
for (const auto& inv : result.m_inventory) { RelayInv(inv); }
```

**"Announced once at DKG finalize" describes only the originator.** Every node that receives and verifies a commitment re-announces it to all its own peers, deduped by `try_emplace` so each relays it exactly once. That is a gossip flood — structurally the same propagation blocks get. `MSG_QUORUM_FINAL_COMMITMENT` was never a narrow-window lottery, so **`p ≈ 1` per connected peer is structural** and N=8 cannot beat N=1 on coverage. No amount of soak time can change that; it is a property of the protocol.

**Measured, same binary, 15-min steady state, N=8 (`soak0804d`) vs N=1 (`soak0804e`):**

| | admitted | per hour | suppressed | announcers/obj |
|---|---|---|---|---|
| N=8 | 6 | 24.0 | 41 | 7.83 |
| N=1 | 6 | 24.0 | 0 | 1.00 |

Identical. The pool saw 41 redundant announcements of *the same six objects*. And the first concurrent `qfcommit` reached all three nodes within 0.8 s — two independent single-peer nodes, on **different** peers, both acquired it.

(An earlier "pool 61 / control 34" reading was sync-phase activity, not steady-state relay — the nodes synced at different times. Not a witness advantage.)

## What this PR actually buys

- **Resilience** — losing a peer no longer means holding zero peers. Real, and not previously true.
- **Latency** — 0.36 s earlier than N=1 on the observed commitment; first-of-N wins the race. Real, but irrelevant to whether a commitment can be mined.
- **Acquisition coverage** — **no.** That was the argument, and it does not hold.

Two wiring bugs found and fixed on the way were worth having regardless: `max_concurrent_connections` shipped at **3**, and the 60 s refresh passed only *pinned* keys as "connected", so the diversity scorer ran against an empty set. Without those, "8 peers" could have been eight sockets into one netblock.

**Nothing here needs reverting** — the change is tested, bounded, and harmless. But the precondition that actually governs acquisition is *being connected at all when the flood passes*, and that is #1074's win (holding the peer instead of dropping it every 101 s), not this PR's.

Two caveats on the record: `AddMineableCommitment` relays again when a commitment arrives with more signers, so a node can legitimately announce the same quorum twice; and this settles **witnessing**, not **serving** — `InvDedup` asks only one announcer and does not retry a second on `notfound`. Cheap under a flood, but noted rather than claimed de-risked.

PR #1089, opened 2026-08-04
dash: event-driven resume for the creditpool-stale serve-gate window (soak0804e)

frstrtr:
## What this is

soak0804e (N=1, 3 h, master 932302a88) measured `creditpool-stale` at ~3.3% of wall-clock across 9 episodes (durations [5,5,5,5,5,60,63,64,208] s), every decline exactly `value = threshold-1`. This PR answers the two questions the finding posed and fixes the resume path. **The consensus gate is untouched** — dashd demands exact `creditPoolBalance` equality (`bad-cbtx-assetlocked-amount`, specialtxman.cpp:749-755) and the refusal between tip-advance and body-parse is correct and pinned by a KAT as preserved.

## Q1 — is the credit-pool advance event-driven or poll-driven? (MEASURED, call chain)

The **ingest is fully event-driven**; no periodic tick anywhere in the chain:

1. block inv -> `new_block` fired (`src/impl/dash/coin/p2p_client.hpp` inv handler, ~:1901-1904)
2. `new_block` subscriber sends `getheaders` + `request_block(hash)` **immediately**, same ordered TCP stream, headers land first (`src/c2pool/main_dash.cpp` ~:3792-3806)
3. headers -> `HeaderChain::add_headers` -> tip-changed callback -> `on_new_tip` (tip advances; gate now correctly stale) + `invalidate_template_cache` + bump + `notify_all` (`main_dash.cpp` ~:3835-3894)
4. block body arrives -> `full_block` (`p2p_client.hpp` block handler ~:1912-1931) -> merkle-bound -> `block_connected` (`src/impl/dash/coin/live_feed.hpp` :104-136) -> `CoinStateMaintainer::on_block_connected` (`src/impl/dash/coin/block_connect_ingest.hpp` :54-58) -> `advance_credit_pool_on_block` (`coin_state_maintainer.hpp`)

So blocks are getdata'd promptly on inv (no lazy body fetch). **But the RESUME was not event-driven, twice over:**

- `on_block_connected`'s success path ended in `republish()` **without** firing the state-dirty sink — unlike every other async advance (mnlistdiff apply, ChainLock adopt, reorg wipe, all of which call `notify_state_dirty()`). The gate turning viable re-issued no work until the next unrelated signal.
- Even when a bump did land, `DASHWorkSource`'s negative template cache (`kRetryAfter` = 5 s, `work_source.cpp:68`) refused to re-source after a failed attempt regardless of events — quantizing any resume to the 5 s retry window.

## Q2 — the fix (latency-only)

- **Maintainer** (`coin_state_maintainer.hpp`): wrap the block-connect fold and fire `notify_state_dirty()` exactly on the credit-pool seed's stale->fresh transition (seed becoming current AT the tip). Historical E2b window fills advance the seed to below-tip heights and can never fire it — no bootstrap notify storm (pinned by KAT).
- **Work source** (`work_source.{hpp,cpp}`): a failed sourcing attempt negative-caches against the generation it was made at (captured before the blocking source, so a bump landing mid-source still counts); a generation bump breaks through and re-sources immediately. With no event, the down-dashd 5 s throttle is byte-identical.

Post-fix resume: body parse -> `notify_state_dirty` -> bump + `notify_all` -> session re-pull re-sources immediately -> EMBEDDED resumes. The residual window is the genuine header-arrival -> body-parse network latency — the same block-boundary physics dashd itself has.

## Evaluation-quantization report (measure-and-report, no fix needed beyond the above)

- MEASURED (code): decline/resume edges are only *observed* at sourcing attempts (`ServeGateJournal::observe` runs inside `resource_template_now` only). During a decline, sourcing cadence is floored by `kRetryAfter` = 5 s and ceilinged by whoever asks (sessions' 1 s notify retry; the soak probe at 60 s).
- The five 5 s episodes are the **retry-quantum floor**, not body latency: true condition duration <= 5 s each, likely sub-second-to-~2 s (one getdata response after the headers response on an already-open stream). INFERRED from the transport ordering.
- The 60/63/64 s episodes carry the **60 s probe-cadence signature** (resume edge only observable at the next probe). Their true durations may be the same few seconds; they cannot be distinguished from a genuinely slow body fetch without the maintainer-side timestamps, which the journal did not record per-episode. INFERRED.
- The single 208 s episode fits neither quantum and is likely real: `request_block` goes to the primary peer with no timeout/re-request, and `InvDedup` (TTL 600 s) suppresses acting on the same block inv from other peers — an unanswered getdata can stall the body until a later headers-driven re-request. INFERRED, N=1; separate lane if it recurs.
- Net: of the measured 3.3%, the majority is probe/retry-cadence artifact. True steady-state unavailability was plausibly ~0.15% excluding the 208 s tail (~2% including it). A production miner polling continuously would have seen roughly body-latency + up-to-5 s per block pre-fix (~2-4%/episode-block), and sees body-latency only (~1 s-class) post-fix.

## Tests (both behaviour KATs verified to FAIL on pre-fix master)

- `DashCoinStateMaintainer.TipBodyArrivalFiresStateDirtyWithoutTemplateRequest` — body arrival advances the pool + fires the re-issue sink with **no template request in between**; also pins the correct refusal (`creditpool-stale`) between tip-advance and body-parse. FAILS on master.
- `DashCoinStateMaintainer.HistoricalBodyFoldBelowTipDoesNotFireStateDirty` — below-tip folds still refuse and do not notify (guard; passes pre-fix too, by design).
- `DashStratumWorkSource.GenerationBumpBreaksThroughNegativeTemplateCache` — the 5 s retry quantum yields to an event; holds with no event. FAILS on master.

Both suites fully green with the fix (25/25 maintainer, 68/68 work source). Targets fold into the already-allowlisted `test_dash_coin_state_maintainer` / `test_dash_stratum_work_source`; `tools/ci/check_test_target_allowlist.py` green (230 targets).

DASH lane only. No consensus-gate change. Non-destructive.

### DIFF
--- a/src/impl/dash/coin/coin_state_maintainer.hpp
+++ b/src/impl/dash/coin/coin_state_maintainer.hpp
@@ -768,6 +768,39 @@ public:
     /// template with a phantom payee. Returns apply_block's ApplyResult.
     MnStateMachine::ApplyResult
     on_block_connected(const dash::coin::BlockType& block, uint32_t height) {
+        // EVENT-DRIVEN RESUME (soak0804e, creditpool-stale ~3.3% wall-clock):
+        // between the header tip advance (on_new_tip) and this body fold the
+        // credit-pool seed is exactly one block behind the tip and the serve
+        // gate CORRECTLY refuses (value = threshold-1; consensus demands exact
+        // creditPoolBalance equality — dashd specialtxman.cpp:749-755). The
+        // ingest was already event-driven, but the RESUME was not: a successful
+        // tip-body fold ended in republish() without firing the state-dirty
+        // sink, so no work re-issue happened until the next unrelated signal
+        // (template request / mnlistdiff / next tip, up to minutes away).
+        // Detect the stale->fresh transition on the credit-pool axis — the
+        // seed becoming current AT the tip — and fire the same re-issue path
+        // every other async advance uses (H-6). Historical window fills advance
+        // the seed to heights BELOW the tip and can never fire this, so the
+        // E2b bootstrap causes no notify storm. The refusal between
+        // tip-advance and body-parse is untouched: this is latency work, the
+        // gate itself is not weakened.
+        const bool cp_was_current_at_tip =
+            m_have_tip
+            && m_state.credit_pool_height() == static_cast<int32_t>(m_prev_height);
+        auto r = on_block_connected_impl(block, height);
+        const bool cp_now_current_at_tip =
+            m_have_tip
+            && m_state.credit_pool_height() == static_cast<int32_t>(m_prev_height);
+        if (!cp_was_current_at_tip && cp_now_current_at_tip)
+            notify_state_dirty();
+        return r;
+    }
+
+private:
+    /// The body of on_block_connected. Reached ONLY through the public wrapper
+    /// above (which detects the credit-pool stale->fresh transition around it).
+    MnStateMachine::ApplyResult
+    on_block_connected_impl(const dash::coin::BlockType& block, uint32_t height) {
         // E2 finding A (reward-critical, defence-in-depth): this component EXTRACTS
         // the reward-critical creditPoolBalance from the block's coinbase, so it
         // validates its own input at the trust boundary. The body↔header binding is
@@ -935,6 +968,7 @@ public:
         return r;
     }
 
+public:
     /// Reorg / MN-list gap / mempool flush: invalidate the live bundle so the
     /// next get_work falls back to dashd until a fresh tip rebuilds it. The
     /// stashed tip params are dropped -- a reorg means the old prev_hash is no
--- a/src/impl/dash/stratum/work_source.cpp
+++ b/src/impl/dash/stratum/work_source.cpp
@@ -314,6 +314,13 @@ void DASHWorkSource::resource_template_now() const
     // OUTSIDE the lock (the fallback arm is a blocking dashd RPC); the cache is
     // then updated under template_mutex_.
 
+    // Captured BEFORE sourcing: a failed attempt negative-caches against THIS
+    // generation, so an event (work-generation bump) that lands while the
+    // blocking source runs still breaks through the retry throttle (soak0804e
+    // resume-quantization fix — see cached_work()).
+    const uint64_t gen_at_source =
+        work_generation_.load(std::memory_order_relaxed);
+
     // ── Mainnet embedded gate (v0.2.4 trigger) ──────────────────────────────
     // The SML/QuorumManager wiring emits a real DIP-0004 type-5 CCbTx, and its
     // byte-parity against a real dashd getblocktemplate is PROVEN (from the raw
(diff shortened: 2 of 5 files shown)


frstrtr:
## Operator direction folded in: serve-tip source measured, body-first scoped as follow-up

Operator direction (via coordinator) reframes the target from "make the advance faster" to "make the window not exist" — LTC-style body-first state advance. Addressed as follows.

### What the serve tip is TODAY (MEASURED, file:line)

**Header-first.** The serving height `m_prev_height` — the template height and the threshold of every freshness gate (`creditpool-stale`, `payee-stale`, `dmn-stale`, `bestcl-*`) — is written only by `NodeCoinState::set_tip` (`src/impl/dash/coin/node_coin_state.hpp:302-306`), reached only via `CoinStateMaintainer::republish()` from `on_new_tip` (`src/impl/dash/coin/coin_state_maintainer.hpp:744-758`), which is fired by the header chain's tip-changed callback through `tip_advance_from_chain` (`src/impl/dash/coin/live_feed.hpp:60-78`; wiring `src/c2pool/main_dash.cpp` ~:3835-3860, `new_tip.happened`). Block bodies play no role in serve-tip advance.

So the operator's diagnosis is confirmed: we declare serve-tip = H off the header, then correctly refuse until body H's coinbase is parsed — the `creditpool-stale` window is self-inflicted by the tip-advance ordering, and dashd never exhibits it because dashd's tip only advances after full-block processing.

### Why body-first is a follow-up PR, not folded in here

By the direction's own criterion ("if body-first turns the PR from latency-tuning into architecture, say so and split"): it does.

1. **`m_prev_height` fan-out.** It feeds template build, all gate thresholds, the serve-gate journal heights, the dashboard topology card, won-block height-race detection, the oracle shadow, and the reorg wiring. Moving its advance from header-time to body-time needs a deferred-TipAdvance latch in `main_dash`, ordering guarantees among the `block_connected` subscribers (maintainer fold must complete before the tip fires), and skip-tip handling when a newer header lands before an older body.
2. **A policy flip that needs its own review.** Header-first stale-work invalidation (kept, per constraint 1) plus body-first serving means: on header H we invalidate jobs, and the replacement job is built on serve-tip H-1 — deliberately serving prev=H-1 work while knowing H exists. That is dashd-equivalent propagation-race physics and the right steady-state behavior, but it is the *opposite* of today's refuse-while-stale policy, and in the lost-body tail (the 208 s episode here) it would serve doomed-tip work for minutes unless paired with a body-fetch watchdog/re-request. That trade deserves its own PR with the invariant test the direction names — "serve height never exceeds the height whose body has been parsed" — which fails on today's header-first master by design and therefore cannot land inside this PR without leaving CI red.
3. **Bounded full-block buffer + undo.** Reorg undo for creditpool/MN-root needs the bodies of disconnected blocks. Proposed bound for the follow-up: retain full blocks for heights ABOVE the last ChainLocked height plus a 2-block margin, floor 6, hard cap 24 (~1 h of DASH blocks). Rationale: ChainLock depth is the principled reachability cap — nothing at or below a ChainLocked block can be reorged — and DASH ChainLocks normally land within seconds of the block, so the buffer is typically 1-3 blocks; the 24 cap bounds memory through a ChainLock outage of ~1 h, beyond which the existing wipe + cold-resync path (`on_sml_reorg`, `coin_state_maintainer.hpp`) remains the correct fallback rather than unbounded retention.

### Why this PR still lands now

This change is the substrate body-first needs, not a detour: body-first serve-tip advance fires exactly at the body-connect fold, which is the event this PR makes notify (`notify_state_dirty` on the stale->fresh transition) and makes the work source honor immediately (generation break-through of the retry throttle). With this PR the window shrinks to genuine arrival+parse latency and the resume is event-driven; the follow-up then removes the window as a class by deferring the serve-tip advance to that same, now-notifying, fold — and adds the atomicity invariant test as its fails-on-master proof.

PR #1096, opened 2026-08-04
dash: wire post-broadcast found-block confirm/orphan lane

frstrtr:
## What & why

DASH found blocks stayed **\"pending\" on the dashboard forever**, and orphans (e.g. hotel block 2508008) were discovered by humans, not the board. Root cause: `main_dash.cpp` wired **none** of the post-broadcast confirmation seam — `grep set_block_verify_fn|add_chain_verify_fn|schedule_block_verification main_dash.cpp` returns 0 — while LTC wires both (`main_ltc.cpp:2105`/`:3013`/`:4258`/`:6315`). This PR arms that lane for DASH.

Premise re-verified on base `edcc2762`: DASH had zero calls to the three functions; the two found-block record paths are `main_dash.cpp` peer-found feed (the `make_on_block_found` handler) and the local solo win (`set_on_found_block_fn`).

## Scope: telemetry only

Runs **strictly after submission**; touches no submit / mint / target / payout path. The pre-broadcast payee guard (`work_source.cpp`) remains the sole consensus gate. A wrong verdict here can only mislabel a dashboard row, never a broadcast.

## Change

- **New header-only resolver** `dash::coin::block_confirm::resolve_status` (`src/impl/dash/coin/block_confirm.hpp`) — the daemonless confirm/orphan verdict. Given \"which hash won height h\", the tip height, and our recorded block, it returns `>0` confirmations (confirmed), `<0` orphaned, `0` pending.
- **Daemonless-first wiring** in `main_dash.cpp`: `set_block_verify_fn` resolves against the embedded X11+DGW `HeaderChain` (`get_header_by_height()` reflects the current best branch, rebuilt from the tip via `prev_hash` on every reorg), so **orphan detection needs no dashd** and does not depend on our own orphaned header ever being relayed to us — it compares the *winner* at our recorded height against our hash. `dashd getblockheader` (confirmations `== -1` ⇒ orphaned) is used **only** on the fallback arm when there is no embedded chain. The block height comes from our own found-block record — authoritative even for an orphan.
- `schedule_block_verification` called on **both** DASH found-block record paths so the poller arms at the standard cadence, exactly as LTC does.
- `MiningInterface::run_block_verification_now` — a small, chain-neutral synchronous test/introspection seam driving the real `verify_found_block` transition (the async `schedule_block_verification` remains the live driver).

## Daemonless resolution — confirmed possible, no new protocol work

The embedded header chain already answers \"is this hash on the best chain at depth N\" via `get_header_by_height`, so no `dash-orphan-flag-daemonless-todo` is needed. The dashd arm is a pure fallback.

## Tests

`test/test_dash_block_confirm_lane.cpp` (folded into the allowlisted `test_dash_node` target; `tools/ci/check_test_target_allowlist.py` passes):

- `resolve_status` confirmed / orphaned / pending, depth gating, and a reorg that restores our block to confirmed.
- MiningInterface flips **pending→confirmed** and **pending→orphaned** through the real transition code (logs emit the actual `BLOCK CONFIRMED` / `BLOCK ORPHANED` lines).
- `NoVerifierLeavesPending` pins the **master baseline**: with no verifier armed a DASH found block stays pending forever (the defect).

Locally: `test_dash_node` builds and all 9 `DashBlockConfirmLane.*` pass; the `c2pool-dash` binary (main_dash.cpp) builds clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01XxfnL8QaD6DHtwD5UgXHwn

### DIFF
--- a/src/c2pool/main_dash.cpp
+++ b/src/c2pool/main_dash.cpp
@@ -72,6 +72,7 @@
 #include <impl/dash/coin/quorum_member_source.hpp>  // E1 Phase-L: daemonless member-set sourcing (the provider)
 #include <impl/dash/coin/utxo_lane.hpp>    // dash::coin::UtxoLane — embedded UTXO/fee lane (E2b, #738)
 #include <impl/dash/coin/header_chain.hpp>       // dash::coin::HeaderChain — SPV header/tip authority (E2a)
+#include <impl/dash/coin/block_confirm.hpp>      // dash::coin::block_confirm — post-broadcast confirm/orphan verdict
 #include <impl/dash/coin/chain_rpc.hpp>          // dash::coin::chain_rpc — daemonless getbestblockhash/getblockhash/getblockchaininfo
 #include <impl/dash/coin/coin_state_maintainer.hpp>  // dash::coin::CoinStateMaintainer — populate ordering gate (E2a)
 #include <impl/dash/coin/sml_quorum_db.hpp>      // dash::coin::SMLDb / QuorumDb — SML+quorum persistence (incremental restart)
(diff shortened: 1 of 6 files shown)


frstrtr:
Superseded by #1098 — identical tree, re-committed clean (author frstrtr, no AI-attribution footer) to satisfy the project's no-attribution law and the attribution-gate. Branch left intact per non-destructive policy.

PR #1105, opened 2026-08-05
ci: reap-guard — host-keyed heavy-leg concurrency cap + #1059 detector auto-rerun backstop

frstrtr:
## Reap-guard: host-keyed heavy-leg cap (fix) + detector auto-rerun (net)

Cycle-2 verdict accepted. Ordering per integrator: concurrency cap leads, auto-rerun is the backstop.

### 1. Heavy-leg concurrency cap
The two RAM-heavy legs — `linux-asan` (ASan/UBSan build+test, `build.yml`) and CodeQL `analyze` (`codeql-analysis.yml`) — co-resident on one physical host is the steps=0 reap root cause (the 08:19 / 08:39 reaps; #1027/#1031 class). Both jobs now carry the **same cross-workflow** job-level group:

```yaml
concurrency:
  group: c2pool-heavy-leg
  cancel-in-progress: false
```

**Why this denies a second heavy leg on the same host:** a concurrency group name is repo-wide (spans workflows), so `build.yml` `linux-asan` and `codeql-analysis.yml` `analyze` mutually exclude. With `cancel-in-progress: false`, when one heavy leg holds the group the second is **queued (pending), not started** — so it can never become co-resident on any host, 198 included. This admits **at most one heavy leg fleet-wide**, a strict superset of the "≤1 heavy leg per physical host" invariant.

**Honest caveat (so the evidence stays honest):** a *per-host* group — one heavy leg per host, letting different hosts each run one in parallel — is **not expressible in pure workflow YAML**. Job-level `concurrency:` is evaluated at **queue time, before runner assignment**, and the only host discriminator — the `145/198/905` token — lives in the runner **name**, which is not in the concurrency-expression context. A true per-host group therefore needs per-host runner **labels** (e.g. `c2pool-host-198`) plus pinning each heavy leg to a host, a separate runner-admin step that also couples each required check to one host's uptime. This PR takes the pure-workflow superset; say the word and I add the labels as a follow-up for cross-host heavy parallelism.

### 2. Detector auto-rerun backstop (#1059)
`runner-drop-detector.yml` now, for each `steps=0` drop with `conclusion==failure` **still on its first live attempt**, calls `gh run rerun <run_id> --failed` once and emits a one-line marker:

    ::notice title=reap-rerun::run <id> (<job> on <runner>) was host-reaped [steps=0]; auto-reran --failed. A green on attempt 2 is a REAP-RERUN, NOT a first-pass green.

Guards: (a) only `failure` conclusion — `cancelled` steps=0 is the concurrency-cancel class, not a reap, so it is never reran; (b) rerun is gated on the **live** `.run_attempt==1` (re-queried per scan), so an already-reran leg is not reran again when its stale attempt-1 record reappears in a later 120m window — no rerun-storm. Permission bumped `actions: read -> write` for `gh run rerun`. Detector still `exit 1` on any drop: the red stays; rerun is the net, not the fix.

### Acceptance evidence
- Concurrency-group expression: cited above; identical `c2pool-heavy-leg` group on both heavy legs with `cancel-in-progress: false` => the second heavy leg is held pending, never co-resident.
- Detector firing on a pinned 198 record: linked in the PR thread.

### Fast-follow (after this PR, per integrator item 3)
Widen the scan and pin 145-1's 23:22-23:33Z leg. Caveat noted honestly: 198-3's rerun landed on 198-2, so 905/145 co-residency is inferred from the shared steps=0 / ~10m server-constant shape, not a direct re-test on 905.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -262,6 +262,15 @@ jobs:
   # Will flip to required once known UAFs are fixed (Phase 7).
   # ════════════════════════════════════════════════════════════════════════════
   linux-asan:
+    # Reap-guard (#1027/#1031 class): host co-residency of two RAM-heavy legs
+    # (this ASan/UBSan build + CodeQL) is what produced the steps=0 reaps on
+    # 198 and 145. A single cross-workflow group with cancel-in-progress:false
+    # admits at most ONE heavy leg fleet-wide, so a second is QUEUED, never
+    # co-resident. (Per-host granularity would need per-host runner labels;
+    # this pure-workflow group is a strict superset of that invariant.)
+    concurrency:
+      group: c2pool-heavy-leg
+      cancel-in-progress: false
     name: Linux x86_64 (AsAN+UBSan)
     # dash bring-up slices may opt out of sanitizers via dash-skip-sanitizers; KEEP for consensus/payout slices (default = on)
     if: ${{ !(github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'dash-skip-sanitizers')) }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -263,20 +263,58 @@ jobs:
   # ════════════════════════════════════════════════════════════════════════════
   linux-asan:
     # Reap-guard (#1027/#1031 class): host co-residency of two RAM-heavy legs
-    # (this ASan/UBSan build + CodeQL) is what produced the steps=0 reaps on
-    # 198 and 145. A single cross-workflow group with cancel-in-progress:false
-    # admits at most ONE heavy leg fleet-wide, so a second is QUEUED, never
-    # co-resident. (Per-host granularity would need per-host runner labels;
-    # this pure-workflow group is a strict superset of that invariant.)
-    concurrency:
-      group: c2pool-heavy-leg
-      cancel-in-progress: false
+    # (this ASan/UBSan build + CodeQL c-cpp analyze) is what produced the
+    # steps=0 reaps on 198 and 145. Exclusion is enforced HOST-SIDE via an
+    # exclusive flock on /tmp (the "Acquire per-host heavy-leg lock" step
+    # below): runner processes on one physical host (905/905-2..905-8,
+    # 198-2/198-3) share /tmp, so the flock is exactly per-host mutual
+    # exclusion -- a second heavy leg on the same host WAITS (nothing is
+    # ever cancelled) while heavy legs on different hosts run in parallel.
+    # A workflow-level concurrency group is deliberately NOT used: GitHub
+    # keeps at most 1 running + 1 pending run per group and CANCELS older
+    # pending entries, so a shared repo-global group self-cancels sibling
+    # jobs of the same run and serializes/starves all heavy legs repo-wide.
     name: Linux x86_64 (AsAN+UBSan)
     # dash bring-up slices may opt out of sanitizers via dash-skip-sanitizers; KEEP for consensus/payout slices (default = on)
     if: ${{ !(github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'dash-skip-sanitizers')) }}
     runs-on: ${{ (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository) && fromJSON('["self-hosted","Linux","X64","c2pool-build"]') || 'ubuntu-24.04' }}
     continue-on-error: true
     steps:
+      # Reap-guard: host-scoped exclusivity vs the other RAM-heavy leg
+      # (CodeQL c-cpp analyze). The lock is held by a disowned holder
+      # process so it spans every step of this job (the CodeQL sibling's
+      # heavy phase is an action step that cannot be wrapped in a shell
+      # flock, hence the cross-step holder shape on both sides). Released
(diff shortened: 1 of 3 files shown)


frstrtr:
Red is structural, not a flake — the new global `concurrency: group: c2pool-heavy-leg` causes it.

GitHub keeps at most **one pending run per concurrency group**. `cancel-in-progress: false` protects the *running* job only — when a third run enters the group, the previously *queued* run is CANCELLED as superseded. This morning's traffic (three PRs' CI in parallel) did exactly that: this branch's Analyze (c-cpp) got cancelled, and its replacement has sat QUEUED for ~4h because every heavy leg repo-wide now serializes through one global slot.

So as written the guard converts busy mornings into (a) red PRs from superseded-cancels and (b) multi-hour queue starvation.

A repo-global workflow concurrency group is the wrong primitive for host exclusivity. "CodeQL and ASan never co-resident on one physical host" needs runner-level exclusivity instead, e.g.:
- give the heavy legs a dedicated runner label pair so the scheduler can't co-place them, or
- a host-side lock (flock in a pre-step wrapper on the runner host) so exclusion happens at the machine, not across the whole repo's workflow queue.

Recommend reworking before merge — as-is this would make every busy day red. Leaving the PR unmerged (it's red anyway, so the sweep won't touch it).

frstrtr:
Precision on the mechanism (sharper than my previous comment): the cancel needs NO cross-PR traffic at all. This PR itself puts THREE jobs into the one `c2pool-heavy-leg` group per run — Analyze (c-cpp), Analyze (python), and the ASan leg. GitHub allows at most 1 running + 1 pending per group, so the third arrival cancels the oldest pending — every single run of this PR self-cancels one of its own jobs. That's why Analyze (c-cpp) is CANCELLED while its siblings proceeded. Deterministic, will reproduce on every push. The runner-label / host-side-lock recommendation stands.

frstrtr:
### Detector-firing evidence (live)

The exact 198 records are in the Cycle-2 thread; the recent detector issues surface a 145-1 reap of the identical class (`steps=0` / `failure` / attempt 1), so I demonstrate on that concrete record — run 30930832865, `dash-g3a`:

Detector `select` filter against **attempt 1**:

    c2pool-linux-145-1   failure   run=30930832865   attempt=1   0 steps   dash-g3a / DASH gate G3a

→ the detector flags it (runner_name present, completed, non-skipped, steps==0). This is the reap record.

New rerun guard, live: that run is now `run_attempt=2, conclusion=success` — it was reaped on attempt 1 and reran to green. My guard re-queries the **live** attempt (`=2`), requires `==1`, and therefore **SKIPs** re-rerun — exactly the no-storm behavior. Under this PR the attempt-2 green would have carried the `reap-rerun` marker so it's not mistaken for a first-pass green.

Honest note (per item 3): I'll pin the exact 198 run — and 145-1's 23:22–23:33Z leg — in the widen-the-scan fast-follow after this lands.

frstrtr:
Reworked per the two review comments above — the repo-global `concurrency: group: c2pool-heavy-leg` is gone.

**What replaced it: host-side exclusivity via flock.** Runner processes on one physical host (905/905-2..905-8, 198-2/198-3) share `/tmp`, so an exclusive `flock` on `/tmp/c2pool-heavy-leg.lock` is exactly per-host mutual exclusion — the actual invariant this guard wants (ASan and CodeQL c-cpp never co-resident on one box) with none of the workflow-queue semantics:

- a second heavy leg on the same host **waits** — nothing is ever cancelled, queued means waiting, not killed;
- heavy legs on **different hosts run in parallel** — no repo-wide serialization, no starvation;
- the three-jobs-one-group self-cancel shape is structurally gone: only the two genuinely RAM-heavy legs (build.yml linux-asan, CodeQL c-cpp) take the lock; CodeQL python (`build-mode: none`) doesn't.

**Why not runner labels:** checked the fleet — all Linux runners carry the identical label set, and hosts run multiple runner slots (905 has 8). Disjoint labels for the two legs would still let two heavy jobs land on sibling slots of the same host, so label routing can't express the per-host invariant at all. Per-ref concurrency was also rejected: it doesn't give host exclusivity across refs.

**Mechanics:** the lock is held across the whole job by a disowned holder process (the CodeQL heavy phase is `codeql-action/analyze`, an action step that can't be wrapped in a shell `flock` command), released in an `always()` step. Two anti-wedge nets: a 4 h holder TTL so a host-reaped job can never leave the lock stuck, plus the runner's end-of-job orphan cleanup. Acquire waits up to 60 min; CodeQL job cap goes 90 → 150 min to budget the worst-case wait behind one full ASan leg, while the 60-min **step** cap on the analyze phase keeps the wedged-tracer fast-red property unchanged.

The #1059 detector auto-rerun backstop is kept as-is (its header comment now points at the flock as the fix). Lock semantics (acquire, second-contender block, release-on-kill) verified with a local flock harness before push.

frstrtr:
Post-merge field report: under today's CI storm (~8 concurrent PR runs) the heavy-leg flock behaved correctly but its 60-min acquire timeout tripped on a PR run (#1110's AsAN leg: "per-host heavy-leg lock not acquired within 60 min") — heavy legs serialized on one host longer than the timeout. Suggested tuning for the ci lane: (a) raise the acquire timeout to ~120 min OR (b) spread CodeQL/ASan onto disjoint runner hosts via labels so the lock rarely contends, and (c) consider making the acquire-step failure auto-retry once. Not urgent — storm-load only; single-PR traffic is fine.

PR #1118, opened 2026-08-05
dash(replay): W3 standalone UTXO fold — LevelDB Tier-B chainstate, 100-block undo window, resumable; dashd-exact hash_serialized_2 gate

frstrtr:
Full-history replay **W3** (spec: `DASH_FULL_HISTORY_REPLAY_MODE.md` WP table + `DASH_REPLAY_SIZING_REPORT.md` §Tier-B, frstrtr/the): the **standalone UTXO fold** — the Tier-B chainstate half, feature-flagged, serve-path-inert.

## What it is

`src/impl/dash/coin/replay_utxo_fold.hpp` — a strict left fold of every block's txs into a LevelDB-backed UTXO set:

- **Consumer shape mirrors the replay seam** (`on_replay_block(height, hash, block)` + a plain `std::function` form) **without including the W1/W2 headers** — the sibling PRs (#1115 is W2) are unmerged, and a shared type would couple their landing order. W5 wires fold↔transport with a one-line adapter.
- **dashd ConnectBlock/UpdateCoins equivalence:** coinbase maturity metadata carried (`fCoinBase`+height; `COINBASE_MATURITY=100` predicate exposed); **zero-value spendable outputs KEPT** (deliberate divergence from the mempool lane's `UTXOViewCache`, which drops them as a fee shortcut — wrong for set parity); `OP_RETURN`/oversize excluded per `IsUnspendable`; genesis short-circuited (its coinbase is not in dashd's chainstate); empty-vin special txs (asset-unlock type 9) fold outputs only; **no historical script verification** — ChainLock-as-assumevalid per spec §4.2.
- **Fail-closed, named causes:** missing input (block's changes fully rolled back), height gap, cold start above h=1 (a UTXO fold must start at genesis — see honesty note), BIP30 duplicate, versioned-format mismatch (v1, fail-loud), store errors.
- **Resumable:** write-back overlay + ONE atomic WriteBatch per flush window (default 2000 blocks) carrying coins+undo+cursor; crash resumes at `resume_height()`; redelivery of already-folded heights is acknowledged idempotently.
- **Undo window: last 100 blocks** (Tier-B sizing: ~2 MB), pruned in-batch; `disconnect_tip` restores the byte-identical set (KAT: connect→disconnect hash equality).

## THE gate

`hash_serialized_2()` is a **byte-exact transcription of dashd v23.1.0 `kernel/coinstats.cpp`** (`ComputeUTXOStats`/`ApplyHash`/`PrepareHash`/`FinalizeHash`) — **including the v22-era ternary-precedence quirk**: `VARINT(nHeight * 2 + fCoinBase ? 1u : 0u)` binds as `(code) ? 1 : 0`, so every group's code varint is the literal `0x01`, never the height code. dashd byte-parity IS the spec; a KAT pins the quirk positively **and** negatively so a well-meaning "fix" trips loudly.

Full-chain gate (documented in the module header; not runnable in CI — needs the genesis→tip body stream from W2):

```
c2pool-dash --replay-utxo-db PATH --replay-utxo-hash \
  --replay-utxo-expect 3d14913768a9d492bfa7a42fe9b111cff625b80e35bb4133e1d60cf3991c2319
```

must exit 0 (byte-exact match vs live dashd `gettxoutsetinfo` at h=2,516,758; ~4.52M txouts / ~244 MiB). Until that gate is green the fold state feeds **no** serving decision.

**Honesty note vs the sizing report:** Tier-B says "built from the same ~29 GB replay" (DIP3→tip) — but a correct UTXO set needs bodies from **genesis**, not DIP3 (pre-DIP3 coins are still spendable). Pre-DIP3 bodies are small; the genesis header walk is already W2's join-check path.

## Support changes

- `src/core/leveldb_store.{hpp,cpp}`: additive `for_each_prefix()` ordered streaming scan (`list_keys` can't paginate; the coin scan is ~4.5M rows).
- `src/c2pool/main_dash.cpp`: `--replay-utxo-db` / `--replay-utxo-hash` / `--replay-utxo-expect` — a status/hash **utility that runs and exits** before any node construction. Absent, nothing constructs the module: the serve path is byte-identical.

## KATs (16, all green)

Folded into the allowlisted `test_dash_mempool` target (**same target, no new allowlist entry**; drift-guard run locally: `CI drift-guard OK`, 232 targets):

- dashd `WriteVarInt` pinned vectors + independent-encoder agreement + decode round-trip; `WriteCompactSize` pinned vectors.
- Synthetic-chain `hash_serialized_2` vs an **independently hand-assembled preimage** (own mini-encoders in the test TU); zero-value inclusion, OP_RETURN exclusion, txid-group ordering, group terminator, hashBlock prefix. Empty-set hash additionally cross-checked against an independent Python SHA256d.
- The ternary quirk pinned (literal `0x01`; the "fixed" encoding must NOT match).
- Refusals: missing-input rollback, gap, cold-start, BIP30, version fail-loud.
- Resume across close/reopen (byte-identical hash), undo retention/prune (default window asserted = 100), disconnect/reconnect set identity, maturity metadata, empty-vin special-tx shape, plain-callback consumer form.

## Status / merge posture

**do-not-merge**: replay lane lands in order (W1 fold engine, W2 transport #1115, then this); the full-chain gate KAT is the graduation criterion. No serve-path changes anywhere in this diff.


frstrtr:
**Un-conflicted against current master** (merge of `origin/master` INTO the branch — **no force-push**), now MERGEABLE.

**Conflict resolution**
- `src/c2pool/main_dash.cpp` (the only real conflict): include block — **KEPT BOTH** sibling includes (W2 `replay_bulk_fetch.hpp` + W3 `replay_utxo_fold.hpp`). All replay argv rows verified present post-merge: `--replay-bulk`, `--replay-bulk-capture`, `--replay-utxo-db/-hash/-expect`.
- `test/CMakeLists.txt`: auto-merged cleanly; verified **KEEP-ALL** — `test_dash_mempool` keeps `test_dash_replay_utxo_fold.cpp` (W3), `test_dash_p2p_node` keeps `test_dash_replay_bulk_fetch.cpp` (W2), W1/W4 replay TUs intact. No sibling TU dropped.

**Seam check vs the merged siblings — NO adapter drift**
W2's real `IReplayBlockConsumer::on_replay_block(uint32_t, const uint256&, const BlockType&)` is **signature-identical** to the shape this module mirrored. Proven, not asserted: the test TU now includes the real `replay_bulk_fetch.hpp` and adds `DrivesW2BulkLaneConsumerInterface` — a 3-line adapter deriving W2's actual interface, driving the fold through an interface pointer including the fail-closed false-return contract. Any future drift in that interface breaks this build loudly instead of at integration time.

W1's `DmlFoldEngine` is a deliberately different shape (`fold_block(block, height) -> FoldResult`, no `IReplayBlockConsumer` inheritance) — the UTXO fold has no dependency on it; the two consume the same block stream independently via W2.

**Verified on the merged tree**
- 17 W3 KATs green (was 16 + the new adapter KAT); `test_dash_mempool` 43/43.
- CI test-allowlist drift-guard: `OK, 232 per-coin test targets`.
- `c2pool-dash` links; `--replay-utxo-*` utility and W2's `--replay-bulk` both still function (gate-mismatch path exits non-zero with the named cause, as designed).

`do-not-merge` removed — no problems found; ready for rollup verification.

PR #1125, opened 2026-08-05
dash(replay): wire W4's derived quorum membership into W1's fold — live daemonless replay reaches tip, 4672/4672 byte-exact roots [do-not-merge]

frstrtr:
## The final seam of the DASH FULL-HISTORY REPLAY mode

`the/docs/DASH_FULL_HISTORY_REPLAY_MODE.md` §3 ("Rotated-quorum member sets are
self-derivable"), §4.5 (Phase 1), §6 W4/W5.

**W1** (#1117) folds the deterministic masternode list and self-checks it, per
block, against that block's own committed cbTx `merkleRootMNList`. Its one
unresolved input is `MembersFn`: dashd's punish loop
(`specialtxman.cpp:159-174` → `HandleQuorumCommitment`) needs the ORDERED
member list of a quorum to attribute `PoSePunish` by `validMembers` bitset
index, and W1 fails closed without a resolver, by design.

**W4** (#1119) derives exactly that list from replayed chain bytes — including
the DIP-0024 quarter-rotation port and the `CQuorumSnapshot` PRODUCER that
makes a qrinfo port unnecessary. Its one unresolved input is `MnListAtFn`: the
masternode list at a quorum's WORK block (cycle base − 8).

The two holes are each other's answer. This PR closes the loop.

```
        DmlFoldEngine ──(MN list at work blocks)──▶ QuorumReplayEngine
              ▲                                             │
              └────────────(ordered member sets)────────────┘
```

### The 2513130 seam, and how it closed

The first live daemonless replay (Tier-A run, 2026-08-05, LAN archival peer
VM210 `192.168.86.165:9999`) folded **129** consecutive byte-exact roots from
anchor 2513000 and then **failed closed, loudly, at h=2513130** — the first
mainnet block in that window whose mined qfcommits actually mark members
invalid (four `llmq_60_75` commitments, quorumIndex 2/7/10/13, one invalid
member each). The stop height had been predicted by a static scan before the
run; the run stopped exactly there. `main_dash` had no member resolver.

The same offline harness, fed the 125 needed member sets from the anchor
(dashd `quorum info` at 2513000 — a Phase-1 anchor-class input), proved
**3851/3851** byte-exact roots 2513000→2516851 with zero divergence. That run
proved the DML fold and the root chain; it explicitly did **not** prove
membership *derivation*.

### Live proof of this PR — 4672/4672, membership DERIVED

Same binary, same bulk lane → fold path, same LAN archival peer, `nice`+`ionice`
on an idle host:

```
[run] REPLAY QUORUM SEAM ARMED: W4 member derivation feeds the W1 MembersFn
      (pre-anchor header hashes seeded=1792, pre-anchor cycle snapshots seeded=3,
       pre-anchor work lists seeded=3; rotated membership reads NOTHING seeded
       from cycle base h=2513088 onward)
[run] W2 REPLAY BULK lane ARMED: start_h=2512201 consumer=W1-DML-FOLD(root-checked)
      members=W4-DERIVED

[REPLAY-FOLD SUMMARY] folded=4672 roots_matched=4672 range=2512201..2516872
      mns=2971 wall=300s reg=18 upd=47 rev=2 revive=40 punish=208 ban=58
      spent=12 DIVERGED=none
[REPLAY-SEAM] member_cycles_derived=236 skipped=0 members_answered=162
      members_missing=0 lists_retained=89
[BULK]  delivered=2516872/2516872 total=4672blk/63MB corrupt=0 notfound=0
```

- **4672 / 4672 consecutive byte-exact `merkleRootMNList` checks, anchor
  h=2512200 → tip h=2516872, zero divergence.** It ran straight through
  h=2513130 and every later punishing commitment.
- **208 PoSe punishes and 58 bans folded through DERIVED membership.** A ban is
  an `isValid` flip, which is *under* the committed root — a wrong member set
  would punish the wrong masternode and diverge the very next block. 58 of them
  landed byte-exact.
- **0 resolver misses, 0 skipped member cycles.**
- This **supersedes** the anchor-fed 3851/3851: it covers that whole range plus
  799 blocks earlier and 21 later, with member sets *computed* rather than
  supplied.
- Final MN count 2971 — the same value dashd reports for
  `protx list registered true 2516851`, and the same the Tier-A run ended on.

### What is still seeded, stated plainly

A rotated cycle at base B needs (snapshots at B−C, B−2C, B−3C) + (MN lists at
B−8, B−C−8, B−2C−8, B−3C−8). Snapshots the replay produces itself from B
onward; the lists come from the replayed DML. But with an anchor at X, the
FIRST cycle a replay is asked for is the first batch mined after X, whose base
satisfies B0 − X < C — so deriving it reaches ~3 cycles *before* the anchor. No
choice of X escapes that; moving X back moves B0 back with it.

So a Phase-1 rotated lane seeds, once: **3 pre-anchor `CQuorumSnapshot`s** (a
skip-list mode + an active-member bitset — **not** a member set) and **the MN
list at 3 pre-anchor work heights** (the same trust class as the anchor
prestate, and strictly an *input* to the quarter rotation rather than its
output). Both are guarded:

- `seed_snapshots()` **refuses** any cycle at or after `anchor + 3·dkgInterval`
  — the cycles the replay owes. With anchor 2512200 that line is exactly
  **h=2513088**, the cycle h=2513130's punishing commitments belong to. From
  there on the rotated lane reads nothing seeded, and the run above states that
  in its ARMED line.
- `seed_work_lists()` **refuses** any height at or after the anchor.
- Non-rotated types are refused a snapshot outright — their membership is a
  function of the replayed list alone.

Phase 2 (genesis replay) retires even those. Nothing here is an anchor-supplied
*member set*, which is what the seam existed to remove.

### KATs (allowlisted DASH target `test_dash_mn_state`, 12 tests, all green)

`test/test_dash_replay_quorum_seam.cpp`, against real mainnet bytes captured
read-only from a Dash Core v23.1.7 archival node:

| KAT | What it pins |
|---|---|
| 1 | the pre-anchor snapshot seed parses a real `quorum rotationinfo` capture; the guard refuses producible cycles, non-rotated types, malformed input |
| 2 | the pre-anchor work-list seed parses 4 real work blocks; refuses heights the replay folds itself |
| 3 | **DERIVATION**: the 32 member lists of cycle base 2513088, computed by the quarter-rotation port, matched proTxHash-for-proTxHash **in order** against dashd's own `quorum info 5 <hash>` for quorumIndex 2 (quorum base block 2513090) |
| 4 | **h=2513130**: (a) with no resolver, fails closed with the live run's exact blocking condition, *at `tx[3]`* — proving it folded the all-valid commitments in `tx[1]`/`tx[2]` with no member set; (b) with the KAT-3 **derived** lists it folds, punishes exactly 4, and reproduces the block's own committed root `2315e6df…120b` byte-exact |
| 5 | the member-set size check is a **BOUND**, not an equality (a list shorter than the bitset still folds to the same committed root); an overrunning list is still refused |
| 6 | the bridge installs **both** directions: the fold's refusal changes from "no resolver installed" to a named derivation miss, and the list handed to the quorum lane carries the collateral outpoint that makes the upstream score tiebreak decidable |

Fails on master: KAT-4a's `tx[3]` assertion and KAT-5 both contradict master's
`fold_qfcommit` (which demands a resolver for every non-null commitment and
requires `members.size() == validMembers.size()` exactly); the rest need the
bridge, which master does not have.

### Carried over, with attribution to the finding

Three fixes surfaced only at integration (proving-ground branch
`dash/replay-integration` @293b9d24 — deliberately not a PR); each was checked
against master and none had landed:

1. **`fold_qfcommit`'s `members.size() == validMembers.size()` relaxed to a
   BOUND.** dashd iterates `members.size()` and indexes `validMembers[i]`;
   `GetAllQuorumMembers` may return fewer than `params.size` on a thin list.
   This was the W4 "INTERFACE RECONCILIATION POINT", W1's to relax.
2. **A member resolver is demanded ONLY when a commitment actually marks
   someone invalid** — dashd's punish loop is a provable no-op otherwise. 470 of
   the 595 non-null commitments in 2513001..2516851 mark nobody invalid; this is
   the difference between a run that stalls at h=2513003 and one that needs
   membership only where membership changes state.
3. **The fold's cursor is authoritative over the bulk lane's persistent
   cursor.** A restart resumed the lane at h=2513098 while the fold sat at the
   anchor, and the forward-contiguous fold refused the first delivery. A
   mid-range resume needs a snapshot v3 at the cursor height; until that path
   exists this refuses to half-resume rather than silently skipping blocks.

### Also carried: W5 integration (no PR existed)

`--replay-fold-prestate` (bulk lane DRIVES the fold + per-block root check),
`replay_prestate.hpp`, `replay_fold_consumer.hpp`, the data-driven
`test_dash_replay_run.cpp` harness. Without it there is no `MembersFn` in
`main_dash` to wire. The W3 UTXO lane is deliberately **not** carried — that is
#1118's.

### Scope

DASH lane only. Everything is behind the existing replay flags: with
`--replay-fold-prestate` absent nothing is constructed, and no serve, gate,
template or stratum path is touched. `--replay-fold-quorums` is a further
explicit opt-in. Test-allowlist drift-guard passes (no new target).

`do-not-merge`: this carries an unmerged W5 and a 3.3 MB fixture set; it is up
for review of the seam, not for landing as-is.


frstrtr:
### Un-conflicted against current master + re-verified

`origin/master` (e4fbf2a38, i.e. #1118 W3 + #1120 spork + #1123 web-relay on top
of W1/W2/W4/P0) merged **into** this branch — merge commit b0299a5b, no
force-push. `MERGEABLE` again.

**Conflict shape and resolution.** One conflict, in `src/c2pool/main_dash.cpp`:
both sides appended a replay include after `replay_bulk_fetch.hpp`. Resolved
**KEEP ALL** — master's `replay_utxo_fold.hpp` (W3) plus this branch's
`replay_prestate.hpp` / `replay_fold_consumer.hpp` / `replay_quorum_bridge.hpp`.
`test/CMakeLists.txt` auto-merged and was checked by hand: both sibling TUs are
present and neither was dropped —
`test_dash_mempool … test_dash_replay_utxo_fold.cpp` (W3's) and
`test_dash_mn_state … test_dash_replay_fold.cpp test_dash_replay_run.cpp
test_dash_replay_quorum_seam.cpp` (W1's + W5's + this seam's).

**W5 collision check: none.** Master still has no `replay_prestate.hpp`,
`replay_fold_consumer.hpp` or `--replay-fold-prestate` — #1118 carried only the
W3 UTXO lane. So the W5 integration carried here is still the only copy, not a
duplicate, and this branch's diff vs master is exactly the genuinely new
pieces: `replay_quorum_bridge.hpp`, the fold-consumer pre/post block hooks, the
two seed formats + their guards, `seed_work_block_cl()` on W4's engine, the
carried W5 files, the KAT TU + fixtures, and the two `gen_replay_kat.py`
generators. Nothing of W3/spork/web-relay is re-introduced or re-touched.

**Re-verified on the merged tree** (rebuilt from scratch sources, targets
`c2pool-dash`, `test_dash_mn_state`, `test_dash_mempool`):

```
test_dash_mn_state   81 tests / 6 suites — 80 PASSED, 1 SKIPPED
                     (DashReplayRun.AnchorToTipRootChain: data-driven, inert
                      without C2POOL_REPLAY_PRESTATE/CORPUS — by design)
  DashReplayQuorumSeam.*                                        12 / 12 PASSED
  DashReplayQuorumSeam.Height2513130FailsClosedWithoutAResolver     PASSED
  DashReplayQuorumSeam.Height2513130FoldsThroughDerivedMembership   PASSED
test_dash_mempool    43 tests / 6 suites — 43 PASSED   (W3's TU, unaffected)
```

The h=2513130 pin still behaves exactly as specified against the new master:
with no resolver it fails closed at `tx[3]` with the live run's own blocking
condition (having folded the all-valid commitments in `tx[1]`/`tx[2]` with no
member set), and with the derived cycle-2513088 member lists it folds,
punishes exactly 4, and reproduces the block's committed
`2315e6df…120b` byte-exact.

frstrtr:
### CI-parity system-toolchain build: **clean, no blocker**

The Tier-A/seam runs were built on .211 with a userspace conda-forge gcc-13
(that host has no system toolchain and no passwordless sudo), which is fine for
proving and **not** a hotel artifact. So this branch was rebuilt from scratch on
the actual CI configuration, on dev-host VM111:

```
Ubuntu 24.04.4 LTS · system /usr/bin/c++ = g++ 13.3.0 (Ubuntu 13.3.0-6ubuntu2~24.04.1)
conan install . -pr:a=ci/conan/linux-gcc13.profile --build=missing --output-folder=build_ci
cmake -S . -B build_ci -DCMAKE_TOOLCHAIN_FILE=build_ci/conan_toolchain.cmake -DCMAKE_BUILD_TYPE=Release
cmake --build build_ci --target c2pool-dash test_dash_mn_state
```

- **0 errors**, both targets built and linked; `c2pool-dash --version` reports
  `b0299a5b` (this branch's merge head).
- `test_dash_mn_state` on that binary: **80 PASSED, 1 SKIPPED** (the
  data-driven `DashReplayRun.AnchorToTipRootChain`), including all 12
  `DashReplayQuorumSeam.*`.

Nothing in this branch is toolchain-sensitive — the new code is plain C++20
(`std::function`/`std::map`/`std::optional`/`std::array`) over existing in-tree
headers, and the conda gcc-13 and system gcc-13 are the same major/minor.

**One deploy note, pre-existing and not from this branch:** a default configure
produces a **BLS-STUB** binary (`[init] DASH BLS backend: STUB (fail-closed …)`)
on both toolchains. A hotel artifact needs `-DC2POOL_DASH_BLS=ON
-DDASHBLS_ROOT=…` (what the "DASH + real BLS" CI job builds) plus the known
`LD_LIBRARY_PATH` lib bundle — a lone binary swap crash-loops. Neither is
introduced here; flagging now rather than at deploy time.

Issue #1134, opened 2026-08-05
dash(stratum): DASHWorkSource::resource_template_now reads unguarded NodeCoinState on the rpc_pool worker — safe today only by !coin_p2p arm exclusivity (work_source.cpp:335-383)


frstrtr:
## The same unguarded-NodeCoinState-read shape as the 2026-08-05 crash, on the SERVE path, one config flag away from firing

PR #1135 fixes the oracle-shadow worker reading `NodeCoinState` off the io thread. This issue records the **sibling** the audit turned up, which that PR deliberately does not fix (it would change the serve path).

### The reads

`src/impl/dash/stratum/work_source.cpp`, all inside `resource_template_now()`:

| line | accessor |
|---|---|
| `:335` | `coin_state_.populated()` |
| `:346` | `coin_state_.populated()` |
| `:353` | `coin_state_.select_work(dashd_fallback_)` |
| `:371` | `coin_state_.describe_decline()` |
| `:383` | `coin_state_.embedded_template_emit_ok(sel.work, &emit_why)` |

`NodeCoinState` has **zero mutexes**, and `select_work()` / `describe_decline()` hand raw pointers into live containers out of `make_embedded_work_inputs()` (`node_coin_state.hpp:1066` `&m_mnstates`, `:1087` `&m_sml`), which `build_embedded_workdata()` then dereferences for the whole template assembly.

### The thread

`resource_template_now()` runs on `refresh_executor_` when one is installed (`work_source.cpp:656-659`, inside `cached_work()`), and that executor is a **1-thread `boost::asio::thread_pool`**:

```cpp
// main_dash.cpp:5054-5066
std::shared_ptr<boost::asio::thread_pool> rpc_pool;
if (!coin_p2p && rpc && stratum_server) {
    rpc_pool = std::make_shared<boost::asio::thread_pool>(1);
    work_source->set_refresh_executor(
        [rpc_pool](std::function<void()> job) { boost::asio::post(*rpc_pool, std::move(job)); });
```

Meanwhile `work_source.cpp:625` (`embedded_template_emit_ok` in `cached_work()`) reads the same coin state on the **io/stratum** thread. Two threads, same unguarded object.

### Why it does not corrupt anything today

Pure arm exclusivity, nothing more:

- the executor is installed only under `!coin_p2p` (`main_dash.cpp:5055`);
[...]

Issue #1133, opened 2026-08-05
dash: teardown-order faults — m_tracker_mutex destroyed before m_think_pool joins (node.hpp:175 vs :190), and ~Node runs after web_server/oracle_shadow (main_dash.cpp:733 vs :826)


frstrtr:
## Shutdown-only: worker pools are joined after the objects their tasks touch are already destroyed

Two independent ordering faults, both on the shutdown path. **Neither can fire while `ioc.run()` is running** — they are strictly teardown bugs, which is why they have never been seen live. They will bite the first time a shutdown happens with work in flight (SIGTERM during a think cycle, or a systemd stop mid-sync).

### 1. `m_tracker_mutex` is destroyed before `m_think_pool` is joined

`src/impl/dash/node.hpp`, NodeImpl member declaration order:

```
:165   boost::asio::thread_pool m_verify_pool{4};
:175   boost::asio::thread_pool m_think_pool{1};
:176   std::atomic<bool> m_think_running{false};
:177   std::atomic<bool> m_clean_running{false};
:189   std::atomic<bool> m_rethink_pending{false};
:190   mutable std::shared_mutex m_tracker_mutex;      <-- declared AFTER the pools
```

Members are destroyed in **reverse** declaration order, so `m_tracker_mutex` (`:190`) is destroyed **before** `~thread_pool` joins `m_think_pool` (`:175`). Every think/clean task takes `std::unique_lock(m_tracker_mutex)` and the IO-side helpers take `shared_lock(..., std::try_to_lock)`. Destroying a `std::shared_mutex` that another thread may lock is UB. Same for the three `std::atomic<bool>` flags at `:176`, `:177`, `:189` that the pool tasks read and write.

Note the *earlier* members are fine: `m_tracker` (`:109`) and `m_storage` (`:110`) are declared **before** the pools, so they outlive the join. The declaration block is right up to `:175` and wrong after it.

**Fix:** declare both pools **last** in the class (or add an explicit `~NodeImpl` that calls `m_verify_pool.join(); m_think_pool.join();` first). The pools must be the first thing torn down and the last thing declared.

### 2. `p2p_node` outlives — and is joined after — every object declared after it in `run()`

`src/c2pool/main_dash.cpp`:

```
:733   dash::Node p2p_node(&ioc, &config);          <-- declared FIRST => destroyed LAST
...
:824   std::shared_ptr<dash::coin::EmbeddedOracleShadow> oracle_shadow;
[...]

Issue #1132, opened 2026-08-05
dash(node): add_verified_shares leaks every skipped share — the two continue paths bypass the only owner transfer (node.cpp:130-136)


frstrtr:
## Shares skipped by `add_verified_shares()` are never freed

`src/impl/dash/node.cpp:127-137`:

```cpp
for (auto& share : data.m_items)
{
    // Skip shares that failed phase-1 verification (hash still null).
    if (share.hash().IsNull())
        continue;                    // <-- node.cpp:130-131  LEAK

    if (m_chain->contains(share.hash()))
    {
        ++dup_count;
        continue;                    // <-- node.cpp:136      LEAK
    }

    m_tracker.add(share);            // <-- the ONLY thing that takes ownership
    ...
}
```

`data.m_items` is `std::vector<ShareType>` (`node.hpp:68`), and `ShareType` is `chain::ShareVariants<DashFormatter, DashShare>` = `std::variant<Args*...>` (`src/sharechain/share.hpp:67`) — a **raw owning pointer**, heap-allocated on the reception path. `HandleSharesData` (`node.hpp:66-80`) has no destructor and `~std::vector<ShareType>` destroys the variants, not the pointees.

`m_tracker.add(share)` is what transfers the pointer into the chain's ownership. Both `continue` paths bypass it, and nothing else ever calls `destroy()` on those shares. They leak, permanently, for the process lifetime.

### Volume

Not a corner case. The `dup_count` branch is the **normal** outcome for any share the node already has — every overlapping batch from every peer during sync leaks one allocation per duplicate. `dup_count` is printed on the very next log line (`node.cpp:157`), so the leak rate is already observable in production logs.

### Fix

`share.destroy()` before each `continue` (or restructure so ownership transfer is unconditional and the tracker rejects). Check the sibling lanes — this is a shared reception-path shape.

Found during the root-cause of the 2026-08-05 hotel heap corruption (PR #1135); the crash itself was unrelated (see PR). Siblings: #1129, #1130, #1131, #1133.

Issue #1131, opened 2026-08-05
dash,btc,ltc: latent DOUBLE FREE in think-P1 bad-share removal — verified.remove(bad) omits owns_data=false (share_tracker.hpp:889/853/875)


frstrtr:
## `verified.remove(bad)` defaults to `owns_data=true` — it would `delete` a share `chain.remove()` then deletes again

`src/impl/dash/share_tracker.hpp:886-892` (think-P1 bad-share removal):

```cpp
try {
    invalidate_weight_caches(bad);
    if (verified.contains(bad))
        verified.remove(bad);        // <-- share_tracker.hpp:889, owns_data defaults to TRUE
    if (chain.remove(bad))           // <-- also owns_data=true
        ++removed_count;
} catch (...) {}
```

`chain` and `verified` are two `ShareChain` instances (`share_tracker.hpp:329-330`) that hold **the same raw share pointers** — `ShareType` is `chain::ShareVariants<...>`, a `std::variant<Args*...>` (`src/sharechain/share.hpp:67`). `ShareChain::remove(hash, owns_data = true)` calls `it->second.share.destroy()` (`src/sharechain/sharechain.hpp:668`, `:846`), i.e. it `delete`s the pointed-to share.

So `verified.remove(bad)` followed by `chain.remove(bad)` destroys the same object twice.

**The rest of the codebase already knows this.** Every other `verified.remove` call site passes the flag explicitly:

```
src/impl/dash/node.cpp:1045   m_tracker.verified.remove(h,         /*owns_data=*/false);
src/impl/dash/node.cpp:1094   m_tracker.verified.remove(aftertail, /*owns_data=*/false);
src/impl/btc/node.cpp:1678    m_tracker.verified.remove(to_remove, /*owns_data=*/false);
src/impl/btc/node.cpp:2318    m_tracker.verified.remove(h,         /*owns_data=*/false);
src/impl/btc/node.cpp:2392    m_tracker.verified.remove(aftertail, /*owns_data=*/false);
src/impl/ltc/node.cpp:1634    m_tracker.verified.remove(to_remove, /*owns_data=*/false);
src/impl/ltc/node.cpp:2259    m_tracker.verified.remove(h,         /*owns_data=*/false);
src/impl/ltc/node.cpp:2333    m_tracker.verified.remove(aftertail, /*owns_data=*/false);
```

Only the three `share_tracker.hpp` copies omit it:

| file:line |
|---|
[...]

PR #1135, opened 2026-08-05
dash(oracle-shadow): resolve the embedded arm on the io thread and hand the worker a deep copy — root-cause fix for the 2026-08-05 hotel heap corruption [do-not-merge]

frstrtr:
## What broke

Hotel primary `c2pool-dash` died **SIGABRT at 21:52:38 MSK on 2026-08-05**, 25 min after deploying `cdd908ce` (`0.2.4-366-gb0299a5b`):

```
malloc(): unaligned tcache chunk detected
double free or corruption (!prev)
```

systemd restarted it 16 s later. **No damage, no shares lost, no block affected.**

## Mechanism

`NodeCoinState` (`src/impl/dash/coin/node_coin_state.hpp`) has **zero mutexes** (`grep -c -i mutex` -> 0), and `select_work()` hands **raw pointers into its live containers** out to the caller:

```cpp
node_coin_state.hpp:1066   e.mnstates = &m_mnstates;   // MnStateMachine map
node_coin_state.hpp:1087   e.sml      = &m_sml;        // vendor::CSimplifiedMNList
```

`build_embedded_workdata()` then dereferences all four (`coin/work_source.hpp:214`, `:225`) for the **entire duration** of the template assembly — mempool selection plus both merkle roots over the full MN list.

`EmbeddedOracleShadow::process_tip()` called `select_work()` on its **own worker thread** (`embedded_oracle_shadow.hpp:664`, worker started `:628`, holding a `const NodeCoinState&` at `:947`) while the io thread mutated the very same containers through `CoinStateMaintainer`:

| writer | what it does |
|---|---|
| `coin_state_maintainer.hpp:489` | `apply_diff` -> `CSimplifiedMNList` vector realloc |
| `coin_state_maintainer.hpp:482` | `mnList.clear()` (null baseBlockHash resync) |
| `coin_state_maintainer.hpp:716` | `mnList.clear()` (`on_sml_reorg`) |
| `coin_state_maintainer.hpp:1092` | `apply_block` -> `MnStateMachine` RB-tree rebalance |

`MNState` (`mn_state_db.hpp:60`) owns two heap vectors, so the overlap corrupts **allocator metadata** rather than merely returning stale values — which is exactly why glibc reported tcache/double-free instead of a wrong template.

### Trigger chain (`main_dash.cpp:4165-4211`)

1. `coin_state.new_tip.happened()` wakes the shadow worker (subscription at `:2941-2943`).
2. ~17 us later the io thread runs `invalidate_template_cache()` -> `bump_work_generation()` -> `notify_all()`.
3. 26 sessions hit a null cache, and `resource_template_now()` runs **inline on the io thread** — the coin-P2P arm has no `refresh_executor_` (`main_dash.cpp:5055`, and `:4197-4204` says so explicitly).
4. => a **second concurrent `select_work()`** while the worker is still inside its own.

The worker exists **only** under `--embedded-oracle-shadow` (gate `main_dash.cpp:2854`, ctor `:2933`). `--embedded-shadow-compare` widens the window by adding a second worker contending for `NodeRPC::m_rpc_mutex` (`rpc.cpp:211`, 12 s socket deadline), stretching the oracle worker's residency inside `select_work()`.

### Confidence — stated plainly

**This was NOT reproduced under a sanitizer.** The diagnosis is convergent **static + timing** evidence: the unguarded-read shape is unambiguous in the source, the writer set is enumerated above, and the trigger interleaving matches the crash timestamp. It is not a captured ASAN trace, and this PR does not claim one.

## The fix

Resolve the embedded arm **inside `on_new_tip()`, on the tip/io thread that exclusively owns `NodeCoinState`**, and enqueue a by-value `TipJob` carrying a deep-copied `DashWorkData`. `process_tip()` and `worker_loop()` no longer name `coin_state_` at all.

This is not a new idea — `EmbeddedShadowCompare::on_serve` already does exactly this (`embedded_shadow_compare.hpp:381-387`: io thread resolves, queue carries a value, worker owns what it reads). This PR copies that established shape rather than inventing a third form.

`classify_decline()` re-runs `make_embedded_work_inputs()` over the **same** unguarded containers, so it is the same exposure and is now sampled on the tip thread too, beside the selection it explains.

### Why deep copy and not a mutex on `NodeCoinState`

I evaluated the lock and rejected it:

1. **The lock would sit on the hot serve path.** `resource_template_now()` reads `NodeCoinState` inline on the io thread for **26 sessions per work-generation bump**. Every one of those would now take a lock that the shadow worker also wants.
2. **The worst case is strictly worse than the bug.** `select_work()`'s fallback arm is a dashd `getblocktemplate` RPC behind `NodeRPC::m_rpc_mutex` (12 s socket deadline). A lock held across `select_work()` is a lock held across that RPC — converting a rare, restart-recoverable corruption into a **routine multi-second serve stall**. That is a bad trade on a mining node.
3. **A partial lock gives false confidence.** `NodeCoinState` exposes many accessors that return references into its internals (`mnstates()`, `sml()`, `mempool()`, `qmgr()`). Locking `select_work()` alone would leave those unprotected while making the class *look* thread-safe.
4. **The copy is not new work.** `WorkSelection` is already **all by value** — `DashWorkData` (`rpc_data.hpp:77`) contains no pointers. We are not adding a copy; we are moving *where* an existing one is produced. It happens once per tip (~2.6 min on mainnet) on a thread that already runs the identical build many times per tip.

If the shadow lane were on the money path this trade would deserve more argument. It is observe-only, so ownership is the right axis.

### Behaviour deltas — the complete list

This is a **lifetime/ownership** change. **No serve gate, arm decision, coinbase, payee, payout or consensus path is touched**, and the shadow stays OBSERVE-only. Two observable deltas, both in the shadow's own diagnostics:

1. **`select_work()`'s fallback arm here is now a no-op lambda.** It must be: the fallback is a dashd RPC and this now runs on the io thread. The shadow never read the fallback template anyway — `process_tip()` consumes `sel.work` **only** on the `is_embedded` branch; the fall-closed branch uses a literal `{}` (`note_fall_closed(next_height, classify(next_height, {}, ...))`). So this is byte-equivalent for every consumer, and one wasted per-tip `getblocktemplate` disappears with it.
2. **A dashd blip on the fall-closed path no longer drops the sample.** Previously a throwing `dashd_gbt_()` inside `select_work()` propagated out of `process_tip()` into `worker_loop()`'s catch, and the tip was silently lost (no `note_fall_closed`, no `remember_tip`). Now the fall-closed sample is always recorded. This makes the graduation ledger **more** conservative (fall-closed counts against the clean streak), never less.

## Other off-io-thread readers of `NodeCoinState` — full enumeration

I traced every reader rather than fixing one and leaving a sibling. Ground truth for "the io thread": `main_dash.cpp:5454` is the **single** `ioc.run()` call, no `std::thread` wraps it, and there is no `make_strand` anywhere in `src/impl/dash/` — so io thread == main thread, and every `subscribe` / timer / socket handler is that one thread.

**Off-io readers: exactly two.**

1. **`EmbeddedOracleShadow` worker** — fixed here.
2. **`DASHWorkSource::resource_template_now()`** on the `rpc_pool` worker (`main_dash.cpp:5056`): reads `coin_state_.populated()` (`work_source.cpp:335`, `:346`), `select_work()` (`:353`), `describe_decline()` (`:371`), `embedded_template_emit_ok()` (`:383`) — while `cached_work()` reads the same state on the io thread at `:625`. **Not fixed here** — it is on the serve path and wants its own PR and soak. It is safe *today* only by **arm exclusivity**: `set_refresh_executor` is installed under `!coin_p2p` (`main_dash.cpp:5055`) while every `NodeCoinState` mutator lives inside `if (coin_p2p)`. Filed as **#1134**, and this PR adds a tripwire test (below) that goes red the day that guard is relaxed.

**Cleared** (off-io but provably no `NodeCoinState` path): `EmbeddedShadowCompare` worker (`embedded_shadow_compare.hpp:364` — holds no coin-state reference at all); ZMQ subscriber (`zmq_tip_notify.cpp:68` — posts to `ioc`); DNS-seed detached threads (`coin_peer_manager.hpp:1137`, `:1162`); `NodeImpl::m_verify_pool` / `m_think_pool` (`node.hpp:165`/`:175` — `NodeImpl::select_work()` and `coin_state()` have **zero callers**, dead seam; `main_dash` uses a standalone `node_coin_state` at `:1690`); WebServer HTTP thread and its precompute/submit threads (dashboard seams go through `get_stratum_workers()` / `peek_template()`, both mutex-guarded snapshots at `work_source.cpp:292`/`:301`); `run_mine_block` (`main_dash.cpp:5589`, stack-local empty bundle); `dash::stratum::get_work()` (`get_work.hpp:61` — no production callers, tests only).

## Test — real red/green, both runs executed

`test/test_dash_oracle_shadow_ownership.cpp`, 8 tests.

- Rides the **existing allowlisted** `test_dash_node_coin_state` target (a new `add_executable` would be a NOT_BUILT sentinel, #143).
- **Not `#ifdef`-guarded** (#895). The `#error` on a missing compile definition is a hard build failure, not a skip — same as the house pattern.
- A true UAF race is nondeterministic and was not reproduced under a sanitizer, so this is a **source-structural ownership guard** — the deterministic red/green this invariant admits. House pattern: `src/impl/dgb/test/race913_lock_guard_test.cpp` (PR #1114).

It strips comments before asserting, so the doc blocks that *quote* the pre-fix expressions cannot satisfy or break a check.

### RED — same test binary, sources from current `master` (unfixed)

```
[==========] Running 8 tests from 1 test suite.
[ RUN      ] DashOracleShadowOwnership.ProcessTipNeverTouchesCoinState
test_dash_oracle_shadow_ownership.cpp:144: Failure
  body.find("coin_state_")
    Which is: 566
  std::string::npos
process_tip() runs on the WORKER thread and must never read NodeCoinState (no mutex;
select_work() hands out raw pointers into live containers -> the 2026-08-05 heap
corruption). Resolve on the tip thread in on_new_tip() and enqueue a deep copy instead.
[  FAILED  ] DashOracleShadowOwnership.ProcessTipNeverTouchesCoinState
[ RUN      ] DashOracleShadowOwnership.WorkerLoopNeverTouchesCoinState
[       OK ] DashOracleShadowOwnership.WorkerLoopNeverTouchesCoinState
[ RUN      ] DashOracleShadowOwnership.OnNewTipResolvesTheArmOnTheTipThread
test_dash_oracle_shadow_ownership.cpp:167: Failure
Expected: (body.find("coin_state_.select_work(")) != (std::string::npos)
on_new_tip() runs on the tip/io thread that owns NodeCoinState and is where the
embedded arm must be resolved.
[  FAILED  ] DashOracleShadowOwnership.OnNewTipResolvesTheArmOnTheTipThread
[ RUN      ] DashOracleShadowOwnership.SelectWorkCalledOnlyFromOnNewTip
test_dash_oracle_shadow_ownership.cpp:184: Failure
Value of: at > on_tip.begin && at < on_tip.end
  Actual: false
select_work() is called from OUTSIDE on_new_tip() — only the tip/io thread may read
NodeCoinState.
[  FAILED  ] DashOracleShadowOwnership.SelectWorkCalledOnlyFromOnNewTip
[ RUN      ] DashOracleShadowOwnership.ClassifyDeclineSampledOnTheTipThread
test_dash_oracle_shadow_ownership.cpp:200: Failure
Value of: at > on_tip.begin && at < on_tip.end
  Actual: false
classify_decline() re-reads m_mnstates/m_sml — it must be sampled on the tip thread
beside select_work(), never on the worker.
[  FAILED  ] DashOracleShadowOwnership.ClassifyDeclineSampledOnTheTipThread
[ RUN      ] DashOracleShadowOwnership.QueuedJobCarriesTheTemplateByValue
test_dash_oracle_shadow_ownership.cpp:212: Failure
TipJob (the by-value queue element) not found
[  FAILED  ] DashOracleShadowOwnership.QueuedJobCarriesTheTemplateByValue
[ RUN      ] DashOracleShadowOwnership.PendingSlotHoldsTheJobByValue
test_dash_oracle_shadow_ownership.cpp:235: Failure
Expected: (src.find("std::optional<TipJob>")) != (std::string::npos)
the coalescing queue slot must be std::optional<TipJob> (by value)
test_dash_oracle_shadow_ownership.cpp:237: Failure
  src.find("std::optional<std::pair<uint32_t, uint256>> pending_")
    Which is: 53334
  std::string::npos
the pre-fix queue element is back: it carried only (height, hash), which forced the
worker to re-read NodeCoinState for the template.
[  FAILED  ] DashOracleShadowOwnership.PendingSlotHoldsTheJobByValue
[ RUN      ] DashOracleShadowOwnership.RefreshExecutorStaysOffTheCoinP2PArm
[       OK ] DashOracleShadowOwnership.RefreshExecutorStaysOffTheCoinP2PArm

[  PASSED  ] 2 tests.
[  FAILED  ] 6 tests
```

### GREEN — same binary, this branch

```
[==========] Running 8 tests from 1 test suite.
[ RUN      ] DashOracleShadowOwnership.ProcessTipNeverTouchesCoinState
[       OK ] DashOracleShadowOwnership.ProcessTipNeverTouchesCoinState (0 ms)
[ RUN      ] DashOracleShadowOwnership.WorkerLoopNeverTouchesCoinState
[       OK ] DashOracleShadowOwnership.WorkerLoopNeverTouchesCoinState (0 ms)
[ RUN      ] DashOracleShadowOwnership.OnNewTipResolvesTheArmOnTheTipThread
[       OK ] DashOracleShadowOwnership.OnNewTipResolvesTheArmOnTheTipThread (0 ms)
[ RUN      ] DashOracleShadowOwnership.SelectWorkCalledOnlyFromOnNewTip
[       OK ] DashOracleShadowOwnership.SelectWorkCalledOnlyFromOnNewTip (0 ms)
[ RUN      ] DashOracleShadowOwnership.ClassifyDeclineSampledOnTheTipThread
[       OK ] DashOracleShadowOwnership.ClassifyDeclineSampledOnTheTipThread (0 ms)
[ RUN      ] DashOracleShadowOwnership.QueuedJobCarriesTheTemplateByValue
[       OK ] DashOracleShadowOwnership.QueuedJobCarriesTheTemplateByValue (0 ms)
[ RUN      ] DashOracleShadowOwnership.PendingSlotHoldsTheJobByValue
[       OK ] DashOracleShadowOwnership.PendingSlotHoldsTheJobByValue (0 ms)
[ RUN      ] DashOracleShadowOwnership.RefreshExecutorStaysOffTheCoinP2PArm
[       OK ] DashOracleShadowOwnership.RefreshExecutorStaysOffTheCoinP2PArm (6 ms)
[----------] 8 tests from DashOracleShadowOwnership (12 ms total)
[  PASSED  ] 8 tests.
```

The two that pass in both are correct: `WorkerLoopNeverTouchesCoinState` (pre-fix `worker_loop` reached the coin state *through* `process_tip`, so it never named it textually — the guard exists to stop a future direct read) and `RefreshExecutorStaysOffTheCoinP2PArm` (a forward-looking tripwire for #1134, not part of this fix).

Reproduce:

```
g++ -std=c++17 -DDASH_ORACLE_SHADOW_SRC='"<tree>/src/impl/dash/coin/embedded_oracle_shadow.hpp"' \
               -DDASH_MAIN_SRC='"<tree>/src/c2pool/main_dash.cpp"' \
    test/test_dash_oracle_shadow_ownership.cpp -lgtest -lgtest_main -pthread -o t && ./t
```
Point the two definitions at a `master` checkout for RED, at this branch for GREEN.

## Companion issues filed from the same audit

| # | severity | what |
|---|---|---|
| #1129 | security-relevant | `size_t` underflow in `ShareMessage::unpack()` on peer-supplied bytes, all 5 lanes. **Honest caveat in the issue:** I could not construct a live path today — the sole caller maintains `offset <= len` — so it is latent hardening, not a live remote crash. Filed with the one-token fix. |
| #1130 | severe | `handle_get_share` returns borrowed `ShareType` handles that outlive the tracker lock (`node.cpp:205-251`). **Measured NOT to be this crash** — zero `[Pool] Sending`, `[CLEAN]` and `handle_get_share` lines in the crash window. Recorded in the issue so nobody re-litigates it. |
| #1131 | latent double free | `verified.remove(bad)` omits `owns_data=false` (`share_tracker.hpp:889` + btc `:853` + ltc `:875`); every other call site in the tree passes it. Unreachable only because of a `continue` 20 lines earlier. |
| #1132 | leak | `add_verified_shares` leaks every skipped share — both `continue` paths bypass the only owner transfer (`node.cpp:130-136`). The `dup_count` branch is the *normal* path. |
| #1133 | shutdown only | `m_tracker_mutex` (`node.hpp:190`) destroyed before `m_think_pool` (`:175`) joins; `~Node` (`main_dash.cpp:733`) runs after `web_server`/`oracle_shadow` (`:826`/`:824`). |
| #1134 | latent, serve path | The sibling reader enumerated above. |

## Verification done

- `g++ -fsyntax-only` on a TU including the modified header — clean.
- Standalone gtest compile + both runs of the new guard, output above.
- Read-only inspection on the hotel nodes; **no restarts, no deploys, no config changes** were made anywhere.

## Deploy note

`do-not-merge` is on this PR pending operator review. The fix is header-only plus a test; the hotel deploy needs the **binary + lib bundle**, not a lone binary swap.

### DIFF
--- a/src/impl/dash/coin/embedded_oracle_shadow.hpp
+++ b/src/impl/dash/coin/embedded_oracle_shadow.hpp
@@ -637,16 +637,76 @@ public:
     EmbeddedOracleShadow(const EmbeddedOracleShadow&) = delete;
     EmbeddedOracleShadow& operator=(const EmbeddedOracleShadow&) = delete;
 
+    /// One queued tip sample. It is a VALUE — everything the worker needs is
+    /// COPIED into it on the tip thread. It deliberately carries no pointer,
+    /// reference or handle into NodeCoinState; see on_new_tip().
+    struct TipJob {
+        uint32_t     next_height{0};
+        uint256      next_prev_hash;
+        /// Which arm the embedded selector chose FOR THIS TIP, resolved on the
+        /// tip thread (see on_new_tip). false => fall-closed to dashd.
+        bool         is_embedded{false};
+        /// The embedded template, DEEP-COPIED out of the selector. Meaningful
+        /// only when is_embedded; DashWorkData is all-by-value (rpc_data.hpp),
+        /// so once it is here it owns every byte the worker reads.
+        DashWorkData emb;
+        /// Log-only decline diagnostic, sampled beside is_embedded. Empty when
+        /// is_embedded.
+        std::string  decline_reason;
+    };
+
     /// Tip-thread entry: ENQUEUE ONLY (never blocks the caller). The worker
     /// thread runs the compare so a hung dashd cannot perturb tip processing.
     /// Coalescing: if a tip is already pending unprocessed, it is replaced by
     /// the newest (a shadow samples tips; a skipped intermediate is a coverage
     /// gap, never a wrong count — and on the ~2.6 min interval the worker keeps
     /// up with room to spare).
+    ///
+    /// OWNERSHIP (heap-corruption fix, hotel SIGABRT 2026-08-05 21:52:38 MSK):
+    /// the embedded arm is resolved HERE, on the tip/io thread that exclusively
+    /// owns NodeCoinState, and the worker is handed a DEEP COPY. It used to be
+    /// resolved on the WORKER thread (process_tip called coin_state_.
+    /// select_work()), which dereferenced `&m_mnstates` / `&m_sml` — raw
+    /// pointers select_work() hands into live containers (node_coin_state.hpp
+    /// :1066/:1087) — while the io thread was concurrently rebalancing the
+    /// MnStateMachine RB-tree (coin_state_maintainer.hpp:1092), reallocating
+    /// the CSimplifiedMNList vector (:489) and clear()ing it (:482/:716).
+    /// NodeCoinState has ZERO mutexes, and MNState owns two heap vectors
+    /// (mn_state_db.hpp:60), so the overlap corrupted ALLOCATOR METADATA rather
+    /// than merely reading stale values — glibc "malloc(): unaligned tcache
+    /// chunk detected" / "double free or corruption (!prev)".
+    ///
+    /// This is the same shape EmbeddedShadowCompare::on_serve already uses
+    /// (embedded_shadow_compare.hpp:381-387): the io thread resolves, the queue
+    /// carries a value, the worker owns what it reads. It is a lifetime/
+    /// ownership change ONLY — no serve gate, no arm decision, no consensus
+    /// path is touched, and the shadow remains OBSERVE-only.
     void on_new_tip(uint32_t next_height, const uint256& next_prev_hash) {
+        TipJob job;
+        job.next_height    = next_height;
+        job.next_prev_hash = next_prev_hash;
+        // The fallback arm is a NO-OP HERE, deliberately. select_work()'s
+        // fallback is a dashd getblocktemplate RPC (NodeRPC::m_rpc_mutex,
+        // rpc.cpp:211, 12 s socket deadline) and this runs on the io thread,
+        // which must never block on an RPC. The shadow never read the fallback
+        // template anyway — process_tip() consumes sel.work ONLY on the
+        // is_embedded branch — so an empty DashWorkData is byte-equivalent for
+        // every consumer, and one wasted per-tip RPC disappears with it.
+        WorkSelection sel = coin_state_.select_work(
+            []() { return DashWorkData{}; });
+        job.is_embedded = (sel.source == WorkSource::Embedded);
+        if (job.is_embedded) {
+            job.emb = std::move(sel.work);
+        } else {
+            // decline-reason DIAGNOSTIC (log-only): captured on the SAME
+            // sample as is_embedded so it matches the selection decision
+            // exactly — and now on the same thread and in the same instant,
+            // instead of across a worker-side RPC round trip.
+            job.decline_reason = coin_state_.classify_decline();
+        }
         {
             std::lock_guard<std::mutex> lk(q_mu_);
-            pending_ = std::make_pair(next_height, next_prev_hash);
+            pending_ = std::move(job);
         }
         q_cv_.notify_one();
     }
@@ -654,19 +714,22 @@ public:
     /// The actual per-block shadow-compare (worker thread). OBSERVE-ONLY. All
     /// heavy RPCs run here BEFORE mu_ is taken, so /embedded_oracle (stats_json)
     /// never blocks on a dashd RPC.
-    void process_tip(uint32_t next_height, const uint256& next_prev_hash) {
+    ///
+    /// THREAD CONTRACT: this function must NEVER touch coin_state_. Everything
+    /// it needs about the embedded arm arrives in `job`, already copied by
+    /// on_new_tip() on the owning thread. The structural guard in
+    /// test/test_dash_oracle_shadow_ownership.cpp pins that.
+    void process_tip(const TipJob& job) {
+        const uint32_t  next_height    = job.next_height;
+        const uint256&  next_prev_hash = job.next_prev_hash;
         if (!dashd_gbt_) { LOG_WARNING << "[EMB-ORACLE] no dashd oracle bound"; return; }
 
         // ── Phase 1: ALL heavy RPCs, NO lock ────────────────────────────────
         // Runs on the worker thread; taking mu_ here would make /embedded_oracle
         // (stats_json) block on a dashd RPC. So gather everything unlocked, then
         // take mu_ only for the fast analysis + ledger phase.
-        WorkSelection sel = coin_state_.select_work(dashd_gbt_);
-        const bool is_embedded = (sel.source == WorkSource::Embedded);
-        // decline-reason DIAGNOSTIC (log-only): captured on the SAME
-        // sample as is_embedded so it matches the selection decision exactly.
-        const std::string decline_reason =
-            is_embedded ? std::string() : coin_state_.classify_decline();
+        const bool is_embedded = job.is_embedded;
+        const std::string& decline_reason = job.decline_reason;
         DashWorkData emb, dashd;
         bool dashd_rpc_ok = false, aligned = false;
         std::optional<int64_t> base_cp;   // creditPool(N-1) from the CONNECTED block
@@ -674,7 +737,7 @@ public:
         ProposalResult pr;
 
         if (is_embedded) {
-            emb = std::move(sel.work);
+            emb = job.emb;   // deep copy already owned by this job
             try { dashd = dashd_gbt_(); dashd_rpc_ok = true; }
             catch (const std::exception& ex) {
                 LOG_WARNING << "[EMB-ORACLE] dashd oracle threw: " << ex.what() << " — void";
@@ -840,15 +903,17 @@ private:
     // RPCs + file writes) never runs on the tip-dispatch thread.
     void worker_loop() {
         for (;;) {
-            std::pair<uint32_t, uint256> job;
+            TipJob job;
             {
                 std::unique_lock<std::mutex> lk(q_mu_);
                 q_cv_.wait(lk, [this] { return stop_ || pending_.has_value(); });
                 if (stop_ && !pending_.has_value()) return;
-                job = *pending_;
+                job = std::move(*pending_);
                 pending_.reset();
             }
-            try { process_tip(job.first, job.second); }
+            // `job` is a self-contained VALUE: from here on the worker reads
+            // nothing that the io thread can concurrently mutate.
+            try { process_tip(job); }
             catch (const std::exception& e) {
                 LOG_WARNING << "[EMB-ORACLE] worker exception: " << e.what();
             }
@@ -944,6 +1009,11 @@ private:
         }
     }
 
+    // TIP-THREAD ONLY. NodeCoinState carries no mutex and hands raw pointers
+    // into its live containers out of select_work(); it may be read ONLY from
+    // the thread that mutates it (the io/tip thread). The single legal use of
+    // this reference is inside on_new_tip(). The worker thread must never
+    // reach it — that was the 2026-08-05 hotel heap corruption.
     const NodeCoinState&          coin_state_;
     std::function<DashWorkData()> dashd_gbt_;
     ProposalFn                    proposal_fn_;
@@ -958,7 +1028,7 @@ private:
     std::thread             worker_;
     std::mutex              q_mu_;
     std::condition_variable q_cv_;
-    std::optional<std::pair<uint32_t, uint256>> pending_;
+    std::optional<TipJob>   pending_;
     bool                    stop_{false};
 
     // Trackers below are WORKER-THREAD-EXCLUSIVE (process_tip only) — no lock.
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -899,7 +899,15 @@ if (BUILD_TESTING AND GTest_FOUND)
     # NodeCoinState: node-held embedded coin-state holder over the
     # work_source selector link set (mn_state_machine + mempool header-only);
     # no ltc/pool SCC dependency. Mirrors the test_dash_work_source link set.
-    add_executable(test_dash_node_coin_state test_dash_node_coin_state.cpp)
+    # D-DASH.ORACLE-OWNERSHIP rides this EXISTING allowlisted target rather than
+    # a new add_executable — a fresh target absent from the build.yml --target
+    # allowlist becomes the silently-passing NOT_BUILT sentinel (#143). It is a
+    # source-structural guard over embedded_oracle_shadow.hpp + main_dash.cpp,
+    # so it needs no link deps of its own; the compile definitions hand it the
+    # paths to the SHIPPED sources.
+    add_executable(test_dash_node_coin_state
+        test_dash_node_coin_state.cpp
+        test_dash_oracle_shadow_ownership.cpp)
     target_link_libraries(test_dash_node_coin_state PRIVATE
         GTest::gtest_main GTest::gtest
         dash_x11 core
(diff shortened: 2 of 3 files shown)


frstrtr:
## Live exposure confirmed on both hotel nodes (read-only check, 2026-08-06 00:06 MSK)

No restarts, deploys or config changes were made — `ps` / `/proc/<pid>/cmdline` only.

**Primary — 109.161.57.3**

```
PID 1437924   started  Ср авг  5 21:52:54 2026
/home/fanfan/c2pool-dash-cdd908ce-STAGED --run --coin-rpc 127.0.0.1:9998 ... --stratum 7903
  ... --coin-p2p-discover --embedded-oracle-shadow --embedded-mainnet --embedded-shadow-compare
```

The start time is **21:52:54** — exactly **16 s** after the 21:52:38 SIGABRT. That is the systemd auto-restart, and it independently corroborates the crash timeline. Up 2 h 14 m since, no repeat.

**Reserve — 109.161.52.148**

```
PID 2064211   started  Ср авг  5 21:28:39 2026
/home/fanfan/c2pool-dash-cdd908ce-STAGED --run ... --coin-p2p-discover
  --embedded-oracle-shadow --embedded-mainnet --embedded-shadow-compare
```

Up 2 h 38 m, has not crashed.

### Why this matters for the review

Both nodes carry **exactly** the flag combination that arms this defect, on the exact binary that crashed:

- `--embedded-oracle-shadow` — creates the worker thread that calls `select_work()` off the io thread. Without it the racing reader does not exist at all (gate `main_dash.cpp:2854`, ctor `:2933`).
- `--coin-p2p-discover` — puts the node on the coin-P2P arm, where `resource_template_now()` runs **inline** on the io thread (no `refresh_executor_`, `main_dash.cpp:5055`), supplying the second concurrent `select_work()`.
- `--embedded-shadow-compare` — adds the second worker contending for `NodeRPC::m_rpc_mutex` (`rpc.cpp:211`, 12 s socket deadline), which stretches the oracle worker's residency inside `select_work()` and widens the window.

So the exposure is live on both nodes right now, and the 2 h 38 m of quiet since is what a low-probability interleaving looks like, not evidence that it is gone.

**Nothing here needs an urgent tap.** The failure mode is a SIGABRT that systemd recovers from in 16 s with no share or block loss — the cost of waiting for a reviewed deploy is bounded and small. This note is so the morning reviewer knows the affected nodes are the ones still running, not so anyone reaches for a hot patch.

### CI note

`Analyze (c-cpp)` is red at the **`Initialize CodeQL`** step (`Set up job` -> `Checkout` -> `Stage CodeQL bundle` all green, every step after it *skipped*). That is the known infra failure mode, not a CodeQL finding on this diff.

frstrtr:
## CI: full rollup green, and the new tests are proven to have actually RUN

Everything passes except `Analyze (c-cpp)`, which is red at the **`Initialize CodeQL`** step (`Set up job` -> `Acquire heavy-leg lock` -> `Checkout` -> `Stage CodeQL bundle` all green, every step from `Set Conan home` onward *skipped*). Known infra failure mode, not a finding on this diff.

Green: `Linux x86_64` (11m10s), `Linux x86_64 (AsAN+UBSan)` (10m48s), `DASH + real BLS` (8m54s), `coin-matrix / dash smoke` (6m49s), `dash-g1` byte-parity KAT, `dash-g3a` populated-regtest block-production, `COIN_BCH` + its ASAN leg, `COIN_DGB +AUX_DOGE`, all six coin-matrix smokes, `dgb-phase-b`, `LTC gate G3a`, `DOGE AuxPoW smoke`, **`Test allowlist drift-guard`**, `attribution-gate`, `PR path filter`, `PR platform coverage status`, `Analyze (python)`, `aggregate`, `evaluate`.

### Not a NOT_BUILT sentinel (#143) and not hollow-green

From the `Linux x86_64` job log — the new source compiles into the existing allowlisted target, and ctest registers and runs all 8 cases:

```
[100%] Building CXX object test/CMakeFiles/test_dash_node_coin_state.dir/test_dash_oracle_shadow_ownership.cpp.o
[100%] Building CXX object test/CMakeFiles/test_dash_node_coin_state.dir/test_dash_node_coin_state.cpp.o
[100%] Linking CXX executable test_dash_node_coin_state
[100%] Built target test_dash_node_coin_state
...
          Start 2274: DashOracleShadowOwnership.ProcessTipNeverTouchesCoinState
          Start 2275: DashOracleShadowOwnership.WorkerLoopNeverTouchesCoinState
          Start 2276: DashOracleShadowOwnership.OnNewTipResolvesTheArmOnTheTipThread
          Start 2277: DashOracleShadowOwnership.SelectWorkCalledOnlyFromOnNewTip
          Start 2278: DashOracleShadowOwnership.ClassifyDeclineSampledOnTheTipThread
          Start 2279: DashOracleShadowOwnership.QueuedJobCarriesTheTemplateByValue
          Start 2280: DashOracleShadowOwnership.PendingSlotHoldsTheJobByValue
          Start 2281: DashOracleShadowOwnership.RefreshExecutorStaysOffTheCoinP2PArm
2273/3214 Test #2274: DashOracleShadowOwnership.ProcessTipNeverTouchesCoinState .......  Passed  0.05 sec
2274/3214 Test #2275: DashOracleShadowOwnership.WorkerLoopNeverTouchesCoinState .......  Passed  0.02 sec
2275/3214 Test #2276: DashOracleShadowOwnership.OnNewTipResolvesTheArmOnTheTipThread ..  Passed  0.02 sec
2276/3214 Test #2277: DashOracleShadowOwnership.SelectWorkCalledOnlyFromOnNewTip ......  Passed  0.02 sec
2277/3214 Test #2278: DashOracleShadowOwnership.ClassifyDeclineSampledOnTheTipThread ..  Passed  0.02 sec
```

`test_dash_node_coin_state` was already in the `build.yml --target` allowlist, so no allowlist edit was needed and the drift-guard confirms it.

PR #1150, opened 2026-08-06
dash(stratum): resolve the coin-state arm on the owning thread and hand the background re-source a VALUE — closes the SERVE-path sibling of the hotel heap corruption [do-not-merge]

### LINKED ISSUE #1134: dash(stratum): DASHWorkSource::resource_template_now reads unguarded NodeCoinState on the rpc_pool worker — safe today only by !coin_p2p arm exclusivity (work_source.cpp:335-383)

frstrtr:
Closes #1134. Step 1 of cutting dashd: this must land before anyone removes `--coin-rpc`.

## The exposure

`DASHWorkSource::resource_template_now()` read `NodeCoinState` directly — `populated()`, `select_work()`, `describe_decline()`, `embedded_template_emit_ok()` — and it runs on the **rpc_pool worker thread** whenever `refresh_executor_` is wired (`main_dash.cpp:5511`). Meanwhile `cached_work()` reads the same object on the io/stratum thread. Two threads, one object with **zero mutexes** (`grep -c -i mutex node_coin_state.hpp` -> 0).

`select_work()` hands **raw pointers into live containers** out to the caller — `e.mnstates = &m_mnstates` (an `MnStateMachine` map) and `e.sml = &m_sml` (a `CSimplifiedMNList` vector) — and `build_embedded_workdata()` dereferences them for the **entire** template assembly. `MNState` owns two heap vectors, so an overlapping maintainer write corrupts **allocator metadata**, not merely values. That is exactly the shape that killed the hotel primary with glibc `unaligned tcache chunk` / `double free` on 2026-08-05 and that #1135 fixed for `EmbeddedOracleShadow` — except this sibling is on the path that builds the templates our miners work on. A corrupted template here is a lost block, not a lost diagnostic sample.

It never fired because of **arm exclusivity and nothing else**: the executor is installed under `!coin_p2p` while every `NodeCoinState` mutator lives inside `if (coin_p2p)`, so on the arm that has the worker the coin state happened to be write-quiescent after startup. No assert, no comment, no invariant — a configuration coincidence. **Removing `--coin-rpc` changes exactly those flags.**

## The fix — lifetime/ownership only

Split the re-source at the thread boundary, mirroring #1135 and `EmbeddedShadowCompare::on_serve` rather than inventing a third form:

| | thread | does |
|---|---|---|
| `resolve_coin_state_arm()` | **owning** | every `NodeCoinState` read; returns a self-contained by-value `CoinStateArm` |
| `resource_template_now(CoinStateArm)` | **any** | every dashd RPC, the GBT-xcheck, the arm log, the serve-gate journal, the shadow-compare enqueue, the cache update. Names no coin state at all. |

`CoinStateArm` carries `gen_at_source`, `try_embedded`, `is_embedded`, a **deep-copied** `DashWorkData`, the `DeclineReport` and `declined_height` — no pointer, reference or handle into the coin state. `cached_work()` resolves **before** posting, outside `template_mutex_`, and the posted lambda carries the value.

`select_work()`'s fallback arm is bound to a **no-op** inside the resolution: the real fallback is a dashd `getblocktemplate` behind `NodeRPC::m_rpc_mutex` (12 s socket deadline) and the resolution may run on the io thread. `select_dash_work()` only invokes the fallback on the non-viable branch, so `resource_template_now()` makes exactly the same RPC, exactly once — only on the other side of the boundary. **Zero RPCs added, zero removed.**

### Why not a mutex on `NodeCoinState`

I evaluated it. The four reasons #1135 gave apply here with **more** force, because this is the serve path:

1. **It lands on the hot path.** `resource_template_now()` sources for 26 sessions per work-generation bump.
2. **The worst case is strictly worse than the bug.** A lock held across `select_work()` is a lock held across a 12 s dashd RPC deadline — converting a rare, restart-recoverable corruption into a **routine multi-second serve stall on a mining node**.
3. **A partial lock gives false confidence.** `mnstates()`, `sml()`, `mempool()`, `qmgr()` all return references into the internals; locking `select_work()` alone would leave them unprotected while making the class *look* thread-safe.
4. **The copy is not new work.** `WorkSelection` is already all-by-value and `DashWorkData` contains no pointers. Only the *location* of an existing copy moved.

## Behaviour deltas — the complete list

**No serve gate, arm decision, coinbase, payee, payout or consensus path is touched.** The arm that gets selected, the template bytes served, the `[EMBED-GATE]`/`[DASH-STRATUM-GBT]` text and the `no_work_reason` surface are all unchanged. Three deltas, all on the executor arm only, all in the conservative direction:

1. **`gen_at_source` is now sampled at resolve time instead of at rpc_pool run time** — i.e. it additionally covers the queueing delay. Both the negative cache (`template_last_fail_gen_`) and the stored `template_cache_gen_` key off it, and an *earlier* sample can only make the cache **more** willing to re-source, never less. Same discipline the soak0804e fix established, applied one step earlier.
2. **The local CPU-only embedded build moves onto the owning thread.** The dashd RPC — the only reason the decouple exists — stays on the pool. On the arm that has an executor today the embedded arm is not even armed, and on the `coin_p2p` arm this build already ran inline on io, so this is a no-op in both shipping configurations and the correct split for the combined one.
3. **The resolution catches `...` as well as `std::exception`,** failing closed into the same empty-template/negative-cache path. It runs between `template_refresh_inflight_` being claimed and the job being posted; an escape there would strand the single-flight flag and wedge every future re-source.

Also corrected: the class threading contract claimed `coin_state_` "has its own internal locking". **It has none.** That sentence is a large part of why a background re-source over it read as obviously fine.

## Off-io reader audit — re-confirmed on current master

Ground truth: `main_dash.cpp:6061` is the **single** `ioc.run()`, no `std::thread` wraps it, there is no `make_strand` anywhere in `src/impl/dash/`, and `core::StratumServer` (which is constructed over that same `ioc`) creates **no thread, strand or pool of its own** — so "the io thread" is the main thread, and every stratum handler, timer and socket callback is that one thread.

**Nothing new landed since #1135.** `git diff 41281884..f5588fd6 -- src/` adds **zero** `std::thread` / `thread_pool` / `make_strand` / `detach()`. The complete thread inventory of the DASH lane is byte-for-byte the one #1135 enumerated. #1126's new `[EMBED-STATUS]` diagnostic reads `node_coin_state` heavily but runs on an `io::steady_timer(ioc)` — the io thread. #1128's and #1126's `work_source.cpp` additions (`[BLOCK-LEDGER]`) read only `wd` and `serve_gate_mutex_`-guarded state.

**Off-io readers of `NodeCoinState`: now ZERO.**

1. `EmbeddedOracleShadow` worker — fixed in #1135.
2. `DASHWorkSource::resource_template_now()` on the rpc_pool worker — **fixed here**.

Cleared (off-io but provably no `NodeCoinState` path), re-verified: `EmbeddedShadowCompare` worker (holds no coin-state reference); ZMQ subscriber (`zmq_tip_notify.cpp:68` — posts to `ioc`); DNS-seed detached threads (`coin_peer_manager.hpp:1137`/`:1162`); `NodeImpl::m_verify_pool` / `m_think_pool` (`NodeImpl::select_work()`/`coin_state()` still have zero callers — dead seam); WebServer HTTP thread (the one lambda that captures the work source calls only `embedded_arm_status_json()`, `serve_gate_mutex_`-guarded; the dashboard seams go through `get_stratum_workers()` / `peek_template()`, both mutex-guarded snapshots); `run_mine_block`; `dash::stratum::get_work()` (still tests-only).

## Is the arm-exclusivity dependency GONE or narrowed?

**Gone, for this shape.** Every remaining `NodeCoinState` read in `DASHWorkSource` is in `resolve_coin_state_arm()`, `cached_work()` or `get_work()` (tests-only), and reaching any of them means going through `core::StratumServer`, which by construction runs on `ioc`. The residual condition is the ordinary "this object's `IWorkSource` surface is called from the stratum io thread" — enforced by the type of the constructor argument, not by a flag combination. `set_refresh_executor` can be installed on either arm without reopening this hole, which is precisely what the dashd cut needs.

Two honest caveats, neither arm-dependent:
- The `NodeCoinState` **class** is still lock-free, so a *future* off-io reader is still a hazard. The new guard enumerates every reader by name and fails on a new one, which is the strongest structural answer available short of locking the class.
- `cached_work()`'s serve-time embedded re-check still reads the coin state on the calling thread. That is *correct by definition* (it consults the CURRENT state) and it is the io thread — but it is a contract, now written down in the header and pinned by the allowlist test, rather than a proof.

## Tests — real red/green, both runs executed

`test/test_dash_work_source_ownership.cpp`, 6 tests. Rides the **existing allowlisted** `test_dash_node_coin_state` target (a new `add_executable` would be a NOT_BUILT sentinel, #143). **Not `#ifdef`-guarded** (#895) — the `#error` on a missing compile definition is a hard build failure, not a skip. A true UAF race is nondeterministic and was **not** reproduced under a sanitizer, so this is a source-structural ownership guard, the house pattern for exactly this situation (`src/impl/dgb/test/race913_lock_guard_test.cpp`, #1114; #1135's own suite). Comments are stripped before asserting, so the doc blocks that *quote* the pre-fix expressions cannot satisfy or break a check.

`EveryCoinStateReadSitsInAnOwningThreadMember` enumerates **every** `coin_state_` access in the file and names the enclosing member, so a newly added reader fails **by name** instead of slipping past a stale signature list.

### RED — same test source, sources from `master` (f5588fd6, unfixed)

```
[==========] Running 6 tests from 1 test suite.
[ RUN      ] DashWorkSourceOwnership.SourcingFunctionNeverTouchesCoinState
test_dash_work_source_ownership.cpp:178: Failure
Value of: body.empty()
  Actual: true
Expected: false
resource_template_now(CoinStateArm arm) not found in .../master_tree/src/impl/dash/stratum/work_source.cpp — the ownership split (#1134) is absent, so the function the refresh_executor_ job runs still resolves the arm itself.
[  FAILED  ] DashWorkSourceOwnership.SourcingFunctionNeverTouchesCoinState (1 ms)
[ RUN      ] DashWorkSourceOwnership.EveryCoinStateReadSitsInAnOwningThreadMember
test_dash_work_source_ownership.cpp:214: Failure
Value of: ok
  Actual: false
Expected: true
NodeCoinState is read inside DASHWorkSource::resource_template_now(), which is not on the coin-state owning-thread allowlist. NodeCoinState has ZERO mutexes and select_work() hands raw pointers into m_mnstates/m_sml; reading it off the io thread is the 2026-08-05 heap-corruption shape. Resolve on the owning thread (resolve_coin_state_arm) and pass a by-value CoinStateArm.
  ... (x5 — one per read: populated(), populated(), select_work(), describe_decline(), embedded_template_emit_ok())
[  FAILED  ] DashWorkSourceOwnership.EveryCoinStateReadSitsInAnOwningThreadMember (1 ms)
[ RUN      ] DashWorkSourceOwnership.OwningThreadResolutionTakesNoDashdRpc
test_dash_work_source_ownership.cpp:236: Failure
  src.find("coin_state_.select_work(dashd_fallback_)")
    Which is: 20387
  std::string::npos
select_work() is being handed the REAL dashd fallback on the coin-state owning (io) thread. That puts a 12 s getblocktemplate deadline on the thread that serves 26 stratum sessions. The fallback RPC belongs in resource_template_now(), which runs on the rpc_pool when one is wired.
test_dash_work_source_ownership.cpp:244: Failure
Expected: (resolve.begin) != (std::string::npos)
resolve_coin_state_arm() not found — the #1134 split is absent.
[  FAILED  ] DashWorkSourceOwnership.OwningThreadResolutionTakesNoDashdRpc (1 ms)
[ RUN      ] DashWorkSourceOwnership.CachedWorkResolvesTheArmBeforePostingTheJob
test_dash_work_source_ownership.cpp:264: Failure
Expected: (resolved) != (std::string::npos)
cached_work() does not resolve the arm at all — the background job is still reading NodeCoinState on the rpc_pool thread.
[  FAILED  ] DashWorkSourceOwnership.CachedWorkResolvesTheArmBeforePostingTheJob (1 ms)
[ RUN      ] DashWorkSourceOwnership.CoinStateArmIsAllByValue
test_dash_work_source_ownership.cpp:289: Failure
Expected: (s) != (std::string::npos)
struct CoinStateArm (the by-value ownership handoff) not found in .../master_tree/src/impl/dash/stratum/work_source.hpp
[  FAILED  ] DashWorkSourceOwnership.CoinStateArmIsAllByValue (0 ms)
[ RUN      ] DashWorkSourceOwnership.HeaderDoesNotClaimCoinStateLocksItself
test_dash_work_source_ownership.cpp:336: Failure
  hdr.find("`coin_state_` has its own internal locking")
    Which is: 2287
  std::string::npos
the threading contract still claims NodeCoinState locks itself. It contains ZERO mutexes. Say so, or the next reader repeats #1134.
test_dash_work_source_ownership.cpp:339: Failure
Expected: (hdr.find("NO internal locking")) != (std::string::npos)
the threading contract must state that coin_state_ has no internal locking and name the threads allowed to read it.
[  FAILED  ] DashWorkSourceOwnership.HeaderDoesNotClaimCoinStateLocksItself (0 ms)

[  PASSED  ] 0 tests.
[  FAILED  ] 6 tests
```

The superseded #1135 tripwire, same RED run:

```
[ RUN      ] DashOracleShadowOwnership.RefreshExecutorArmGuardSupersededByCoinStateOwnership
test_dash_oracle_shadow_ownership.cpp:284: Failure
Expected: (sig) != (std::string::npos)
resource_template_now(CoinStateArm) is gone — the #1134 ownership split has been reverted, so the rpc_pool job resolves the arm itself again and the only thing between this node and the 2026-08-05 heap corruption is the !coin_p2p install guard in main_dash.cpp. Restore the split before relaxing that guard.
[  FAILED  ] DashOracleShadowOwnership.RefreshExecutorArmGuardSupersededByCoinStateOwnership (7 ms)
```

### GREEN — same test source, this branch

```
[==========] Running 6 tests from 1 test suite.
[ RUN      ] DashWorkSourceOwnership.SourcingFunctionNeverTouchesCoinState
[       OK ] DashWorkSourceOwnership.SourcingFunctionNeverTouchesCoinState (1 ms)
[ RUN      ] DashWorkSourceOwnership.EveryCoinStateReadSitsInAnOwningThreadMember
[       OK ] DashWorkSourceOwnership.EveryCoinStateReadSitsInAnOwningThreadMember (1 ms)
[ RUN      ] DashWorkSourceOwnership.OwningThreadResolutionTakesNoDashdRpc
[       OK ] DashWorkSourceOwnership.OwningThreadResolutionTakesNoDashdRpc (1 ms)
[ RUN      ] DashWorkSourceOwnership.CachedWorkResolvesTheArmBeforePostingTheJob
[       OK ] DashWorkSourceOwnership.CachedWorkResolvesTheArmBeforePostingTheJob (1 ms)
[ RUN      ] DashWorkSourceOwnership.CoinStateArmIsAllByValue
[       OK ] DashWorkSourceOwnership.CoinStateArmIsAllByValue (0 ms)
[ RUN      ] DashWorkSourceOwnership.HeaderDoesNotClaimCoinStateLocksItself
[       OK ] DashWorkSourceOwnership.HeaderDoesNotClaimCoinStateLocksItself (0 ms)
[----------] 6 tests from DashWorkSourceOwnership (5 ms total)
[  PASSED  ] 6 tests.
```

Reproduce standalone:

```
g++ -std=c++17 -DDASH_WORK_SOURCE_SRC='"<tree>/src/impl/dash/stratum/work_source.cpp"' \
               -DDASH_WORK_SOURCE_HDR='"<tree>/src/impl/dash/stratum/work_source.hpp"' \
    test/test_dash_work_source_ownership.cpp -lgtest -lgtest_main -pthread -o t && ./t
```

Point the definitions at a `master` checkout for RED, at this branch for GREEN.

### In-tree, on the real allowlisted targets

```
$ ./test/test_dash_node_coin_state
[==========] 84 tests from 8 test suites ran. (5 ms total)
[  PASSED  ] 84 tests.

$ ./test/test_dash_stratum_work_source
[==========] 70 tests from 8 test suites ran. (257 ms total)
[  PASSED  ] 70 tests.
```

`test_dash_stratum_work_source` is the behavioural proof that this is ownership-only: it includes `IoThreadDecoupleServesCachedAndRefreshesOffThread` (asserts the fallback is **never** called on the caller thread, single-flight holds, and the refresh runs off-thread) and both `DashStratumCoinP2pTipInvalidate` tests, which drive `cached_work()` through a deferred executor exactly as the rpc_pool does. All still green, unmodified.

## The tripwire

#1135's `DashOracleShadowOwnership.RefreshExecutorStaysOffTheCoinP2PArm` asserted the `!coin_p2p` install guard. This change makes that guard **non-load-bearing**, and the dashd cut intends to relax it — so continuing to assert it would pin the exact configuration detail we are about to change. It is **superseded, not deleted**: renamed to `RefreshExecutorArmGuardSupersededByCoinStateOwnership`, kept in the same suite with the history written into the comment, and it now asserts the ownership that replaced the guard (the `CoinStateArm` overload exists, its body never names `coin_state_`, and `resolve_coin_state_arm()` is where the reads went). Revert the split -> RED, with the arm-guard warning restated in the failure message.

## Merge sequencing

**This PR does not touch `main_dash.cpp`** — no overlap with the PPLNS / sharechain-explorer work (#1144, #1145) or with the honest-RPC-refusal work. The only shared file with anything in flight is `test/CMakeLists.txt` (two lines) and `test/test_dash_oracle_shadow_ownership.cpp` (#1135's own file, no other PR should be in it).

## Deploy note

`do-not-merge` pending operator review. Header + one .cpp + tests; a hotel deploy needs the **binary + lib bundle**, not a lone binary swap. Read-only inspection only — no restarts, no deploys, no config changes on any node.

### DIFF
--- a/src/impl/dash/stratum/work_source.cpp
+++ b/src/impl/dash/stratum/work_source.cpp
@@ -315,20 +315,52 @@ int64_t DASHWorkSource::peek_template_age_sec() const
 // IWorkSource: work generation -- Stage 4c (the template trio).
 // ─────────────────────────────────────────────────────────────────────────────
 
-void DASHWorkSource::resource_template_now() const
-{
-    // The blocking select_work()/dashd-GBT re-source + cache update. Runs either
-    // inline on the caller (legacy blocking path) OR on the background rpc_pool
-    // thread (io-thread-decouple, via refresh_executor_). select_work() is done
-    // OUTSIDE the lock (the fallback arm is a blocking dashd RPC); the cache is
-    // then updated under template_mutex_.
+// ─────────────────────────────────────────────────────────────────────────────
+// #1134 — OWNERSHIP SPLIT ON THE SERVE PATH.
+//
+// resource_template_now() used to read NodeCoinState itself, and it runs on the
+// background rpc_pool thread whenever refresh_executor_ is wired. NodeCoinState
+// has ZERO mutexes and select_work() hands RAW POINTERS into its live containers
+// (`&m_mnstates`, `&m_sml`) which build_embedded_workdata() then dereferences
+// for the WHOLE template assembly — the identical shape that took the hotel
+// primary down with glibc heap corruption on 2026-08-05 (PR #1135), except this
+// one is on the path that serves miners.
+//
+// It did not fire only because of ARM EXCLUSIVITY: the executor is installed
+// under `!coin_p2p` while every NodeCoinState mutator lives inside
+// `if (coin_p2p)`, so on the arm that has the worker the coin state happened to
+// be write-quiescent. Nothing in the code enforced that, and removing --coin-rpc
+// steers the process straight into the combination it depends on being
+// impossible.
+//
+// The split below is a LIFETIME/OWNERSHIP change only:
+//
+//   resolve_coin_state_arm()      OWNING THREAD. Every NodeCoinState read.
+//                                 Returns a self-contained value.
+//   resource_template_now(arm)    ANY THREAD. Every dashd RPC, the arm log,
+//                                 the journal, the cache update. Names no
+//                                 coin state at all.
+//
+// Same shape as EmbeddedOracleShadow::on_new_tip/process_tip (#1135) and
+// EmbeddedShadowCompare::on_serve. NOT a mutex on NodeCoinState — see the PR
+// body: this is the hot serve path (26 sessions per work-generation bump), a
+// lock across select_work() is a lock across a 12 s dashd RPC deadline, the
+// class exposes many reference-returning accessors so a partial lock only LOOKS
+// safe, and WorkSelection is already all-by-value so the copy is not new work.
+// ─────────────────────────────────────────────────────────────────────────────
 
-    // Captured BEFORE sourcing: a failed attempt negative-caches against THIS
-    // generation, so an event (work-generation bump) that lands while the
-    // blocking source runs still breaks through the retry throttle (soak0804e
-    // resume-quantization fix — see cached_work()).
-    const uint64_t gen_at_source =
-        work_generation_.load(std::memory_order_relaxed);
+DASHWorkSource::CoinStateArm DASHWorkSource::resolve_coin_state_arm() const
+{
+    CoinStateArm arm;
+
+    // Captured BEFORE sourcing — and the resolution IS the first step of
+    // sourcing. A failed attempt negative-caches against THIS generation, so an
+    // event (work-generation bump) that lands while the blocking source runs
+    // still breaks through the retry throttle (soak0804e resume-quantization
+    // fix — see cached_work()). Sampling it here rather than on the rpc_pool
+    // thread additionally covers the queueing delay, which can only ever make
+    // the cache MORE willing to re-source, never less.
+    arm.gen_at_source = work_generation_.load(std::memory_order_relaxed);
 
     // ── Mainnet embedded gate (v0.2.4 trigger) ──────────────────────────────
     // The SML/QuorumManager wiring emits a real DIP-0004 type-5 CCbTx, and its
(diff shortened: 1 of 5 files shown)


frstrtr:
Rebased onto current master `011b104d` (#1144 landed while this was open). Clean rebase, no conflicts — #1144's `test/CMakeLists.txt` edit is in a different block than this PR's two-line addition to the existing `test_dash_node_coin_state` target.

Re-verified against the rebased tree: `DashWorkSourceOwnership` 6/6 and `DashOracleShadowOwnership` 8/8 green, and `main_dash.cpp` still has exactly one `set_refresh_executor(` install site (the superseded tripwire's precondition).

**Sequencing note for the in-flight work:** this PR touches **no** `main_dash.cpp`. #1145's `files` listing shows `src/impl/dash/stratum/work_source.cpp`, but that is stale-base noise — its actual diff (`gh pr diff 1145`) touches only `main_dash.cpp`, `dashboard_pplns.hpp`, `test/CMakeLists.txt`, `test_dash_dashboard_pplns.cpp` and two web-static pages. The only real shared file between this PR and #1145 is `test/CMakeLists.txt`, in different blocks. Either order merges cleanly.

frstrtr:
CI note: `coin-matrix / dash smoke (Linux x86_64)` came back **failure with ZERO steps** on runner `c2pool-linux-145-1`, annotation "The self-hosted runner lost communication with the server". That is the known starvation signature of the dual-role .145 host (it is both a CI runner and the mining testbed), not a result from this diff — no step ever executed, so nothing in the change was compiled or run on that leg. The `aggregate` job failed only as its rollup.

Everything that actually exercises this change is green: **DASH + real BLS 3m10s pass** (builds and runs `test_dash_node_coin_state` incl. the new `DashWorkSourceOwnership` suite and the amended tripwire, plus the 70-test `test_dash_stratum_work_source`), `dash-g1` byte-parity KAT pass, `dash-g3a` populated-regtest block-production pass, `Linux x86_64 (AsAN+UBSan)` pass, `attribution-gate` pass, CodeQL pass.

I have re-run the failed legs. Flagging the .145 starvation as a harness issue rather than chasing it here.

PR #1153, opened 2026-08-06
dash(sml): the tip getmnlistd had no retry and no log line — 586 of 940 fallback seconds. Instrument it, and land the bounded re-request policy [do-not-merge]

frstrtr:
The single highest-value item in the serving lane by measured time — **586 of the 940 seconds** the embedded arm spent on the dashd fallback. This PR lands the **instrument** and the **tested policy**; the wiring waits on one more measurement, for a reason given below.

## What is broken

The SML advances on exactly **one** trigger: a tip change fires `send_getmnlistd(sml_base, new_tip)` and the reply moves `sml_current_hash` forward. **There is no retry anywhere** — the only other `getmnlistd` call sites are the handshake, the RPC seed, and the historical work-block snapshot requests. A single request that is never sent, or sent and never answered, strands the embedded arm until the *next block* fires the trigger again.

Measured on the hotel node (`109.161.52.148`, `a73da8ec`), 5h33m window: the arm spent **943 s** on the fallback, and **586 s of it — 62% of the total and 99% of all `dmn-stale` time — sat in five episodes of 275/156/103/51/1 s.** Every one was closed by `tip advanced`, i.e. by the chain moving on, never by the SML catching up at the tip it was already on:

```
07:34:37.559  [EMBED-GATE] DECLINED cause=dmn-stale
   ... 156 s of dashd-fallback serving, no [SML] line at all ...
07:37:12.924  [EMB-DASH] tip advanced h=2517141      <-- the NEXT BLOCK
07:37:13.471  [COIN-P2P] mnlistdiff: base=...1f tip=...11 +1mn
07:37:13.491  [EMBED-GATE] RESUMED
```

The two competing explanations were **eliminated, not assumed**:
- **Not the base-continuity guard** — zero `[SML] REJECT` lines in the window.
- **Not the historical demux stealing the tip reply** — `QuorumMemberSource::on_mnlistdiff` matches **strictly** on `baseBlockHash.IsNull() && blockHash ∈ await set`, so a tip *incremental* can never be claimed by it.

## Why the retry is not wired yet

Because one question is still open, and the two answers need **different fixes**.

`send_getmnlistd()` returned **silently** when there was no primary peer, and logged nothing on success either. So from the log it is impossible to tell whether the stalled request was **sent and unanswered** (a retry fixes it) or **never sent at all** (a retry sends nothing either, and the peer-availability path is the bug). Wiring the retry now would be tuning against an unmeasured mechanism — the same mistake this lane has spent the week correcting. The instrument ships first; the next stall window will say which case it is, and the wiring follows it.

## What lands

1. **`send_getmnlistd()` observability** — one INFO line per request with full base/target hashes, and a WARNING on the previously silent no-primary drop. Full hashes deliberately: at mainnet difficulty the leading ~14 nibbles are difficulty padding and carry no information, and this fires ~once per tip so it can afford the bytes.

2. **`SmlResyncWatchdog`** — the re-request policy as a testable unit, not yet on a timer. Fail-quiet and strictly bounded, because the failure mode of a retry is hammering a peer that is already struggling:
   - only while the SML is behind the tip;
   - only after `min_quiet` (20 s — well above the measured ~54 ms healthy round trip, well below a block interval);
   - geometric backoff between attempts;
   - at most `max_attempts` (3) per stuck `(tip, sml)` pair, after which it goes **silent** and the next block does exactly what it does today;
   - any SML progress or tip change resets everything;
   - a ZERO/cold SML is never its business (that is a full-snapshot re-seed, a much heavier request with its own drivers).

## Nothing is relaxed

No gate, threshold or predicate is touched. The watchdog emits a **request** and nothing else, and that request is **byte-identical to the one the tip-change path already sends** — so it can only make the state the gate reads arrive *sooner*. A gate satisfied earlier because its input arrived is categorically different from a gate satisfied because it was lowered, and nothing here can serve a template that would not have been served one block later anyway. `RequestIsIdenticalToTheOneTheTipChangeWouldSend` pins that equivalence.

## Tests

`test/test_dash_sml_resync_watchdog.cpp`, 8 tests, on the **existing allowlisted** `test_dash_node_coin_state` target (a new `add_executable` would be a NOT_BUILT sentinel, #143); not `#ifdef`-guarded (#895). They pin the **bounds**, which is where the risk is.

```
[       OK ] DashSmlResyncWatchdog.StuckSmlIsReRequestedAfterTheQuietPeriod (0 ms)
[       OK ] DashSmlResyncWatchdog.OrdinaryPerTipWindowNeverRetries (0 ms)
[       OK ] DashSmlResyncWatchdog.SmlProgressResetsAttempts (0 ms)
[       OK ] DashSmlResyncWatchdog.ReachingTheTipClearsEverything (0 ms)
[       OK ] DashSmlResyncWatchdog.BacksOffGeometricallyThenStopsAtTheCap (0 ms)
[       OK ] DashSmlResyncWatchdog.NewTipRestartsTheBudget (0 ms)
[       OK ] DashSmlResyncWatchdog.ColdOrWipedSmlIsNeverReRequestedHere (0 ms)
[       OK ] DashSmlResyncWatchdog.RequestIsIdenticalToTheOneTheTipChangeWouldSend (0 ms)
[  PASSED  ] 8 tests.
```

Whole target green (86/86) and the full `c2pool-dash` binary builds clean with the `p2p_client.hpp` change.

**Honest note on red/green:** these are green-only. `SmlResyncWatchdog` is a new class, so there is no unfixed version to fail against — a RED here would be manufactured and I am not dressing one up. The defect it addresses is evidenced by production measurement (above) rather than by a failing assertion, and `OrdinaryPerTipWindowNeverRetries` is the test that would catch the way this change could plausibly go wrong.

## When it is wired — the one design trap found

The watchdog must be fed the **same** tip hash the gate compares against (`NodeCoinState::m_prev_hash`), **not** the header-chain tip. They differ during `tip-body-pending`, and feeding the header tip would let the SML run *ahead* of the serve tip — producing `dmn-stale` from the other direction. That needs a `prev_hash()` accessor and belongs with the wiring, so it is not in this PR.

## Sequencing

Touches `src/impl/dash/coin/p2p_client.hpp`, one new header, one new test, and `test/CMakeLists.txt`. **No `main_dash.cpp`**, **no `work_source.cpp`**, **no `node_coin_state.hpp`** — so no collision with #1150, #1152, #1145 or #1148. #1148 also touches `p2p_client.hpp`; my change there is confined to the body of `send_getmnlistd()`.

`do-not-merge`. Read-only on the node throughout; builds on VM905.

### DIFF
--- a/src/impl/dash/coin/p2p_client.hpp
+++ b/src/impl/dash/coin/p2p_client.hpp
@@ -1323,9 +1323,29 @@ public:
     uint64_t body_rerequests_exhausted() const { return m_body_rerequests_exhausted; }
 
     /// Send a getmnlistd (SML diff request) — E2/E3 masternode-list sync seam.
+    ///
+    /// OBSERVABILITY (2026-08-06). This used to return SILENTLY when there was
+    /// no primary peer, and logged nothing on the way out either. That made a
+    /// whole class of stall undiagnosable: when the hotel node's SML sat behind
+    /// the tip for 156 s there was no way to tell from the log whether the
+    /// request had been sent and gone unanswered, or had never been sent at
+    /// all. Those two have completely different fixes, and the absence of this
+    /// one line is what made the question unanswerable.
+    ///
+    /// Full hashes, not a prefix: at mainnet difficulty the leading ~14 nibbles
+    /// of a block hash are difficulty padding and carry no information, and
+    /// this line fires about once per tip so it can afford the bytes.
     void send_getmnlistd(const uint256& base_block_hash, const uint256& block_hash)
     {
-        if (!m_primary) return;
+        if (!m_primary) {
+            LOG_WARNING << "[COIN-P2P] getmnlistd DROPPED (no primary peer):"
+                        << " base=" << base_block_hash.GetHex()
+                        << " target=" << block_hash.GetHex()
+                        << " — the SML cannot advance until a peer is up";
+            return;
+        }
+        LOG_INFO << "[COIN-P2P] getmnlistd -> base=" << base_block_hash.GetHex()
+                 << " target=" << block_hash.GetHex();
         auto msg = message_getmnlistd::make_raw(base_block_hash, block_hash);
         m_primary->write(msg);
     }
--- a/src/impl/dash/coin/sml_resync_watchdog.hpp
+++ b/src/impl/dash/coin/sml_resync_watchdog.hpp
@@ -1,0 +1,165 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+#pragma once
+
+// dash::coin::SmlResyncWatchdog — re-ask for the tip mnlistdiff when the one we
+// asked for never came back.
+//
+// ── THE GAP THIS CLOSES (measured, hotel node 109.161.52.148, 2026-08-06) ────
+//
+// The SML advances on exactly one trigger: a tip change fires
+// `send_getmnlistd(sml_base, new_tip)` (main_dash.cpp) and the reply moves
+// `sml_current_hash` forward. There is NO retry anywhere — the only other
+// getmnlistd call sites are the handshake, the RPC seed, and the historical
+// work-block snapshot requests. So a single request that is never sent or never
+// answered leaves the SML behind the tip until the NEXT block fires the trigger
+// again, roughly one block interval (~2.6 min on mainnet) of serving from the
+// dashd fallback instead of the embedded arm.
+//
+// That is not a hypothesis. Over a 5h33m window the embedded arm spent 943 s on
+// the fallback, and 586 s of it — 62% of the total, and 99% of all `dmn-stale`
+// time — sat in just FIVE episodes of 275/156/103/51/1 s. Every one was closed
+// by `[EMB-DASH] tip advanced`, i.e. by the chain moving, never by the SML
+// catching up at the tip it was already on. Traced at h=2517141:
+//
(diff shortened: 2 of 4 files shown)


frstrtr:
Rebased onto `65f04fe8`. The conflict was in `test/CMakeLists.txt` and was entirely self-inflicted: #1150 added `test_dash_work_source_ownership.cpp` to the `test_dash_node_coin_state` target and this PR adds `test_dash_sml_resync_watchdog.cpp` to the same target. Resolution keeps both.

Re-verified on the new merge base: `test_dash_node_coin_state` **92/92** — #1150's ownership suite, the oracle-shadow suite and the new watchdog suite all coexisting on that target.

frstrtr:
CI: `COIN_BCH Linux x86_64 (AsAN+UBSan)` failed with **zero steps** on runner `c2pool-linux-145-1`, annotation "The self-hosted runner lost communication with the server". Not this diff — no step executed, and a BCH ASAN leg cannot be affected by a change confined to `src/impl/dash/coin/p2p_client.hpp` plus a new DASH-only header and test. Re-run triggered.

**This is the second .145 starvation on my PRs in this session** (the first took out `coin-matrix / dash smoke` on #1150). Same host, same zero-step signature both times. .145 is dual-role — CI runner *and* mining testbed — so CI legs land on a box that is already loaded. Flagging as a harness pattern rather than chasing it per-PR.

PR #1202, opened 2026-08-09
dash(serve-gate): per-cause episode duration and cumulative time roll-up

frstrtr:
## What

`creditpool-stale` had 107 episodes and zero attributed duration; serving loss could not be attributed to a cause. `ServeGateJournal` already computed the per-cause segment clock (`cause_sec`/`prev_cause_sec`) but the consumer only ever printed the whole-episode `dur=`, and there was no cumulative summary — so every serving-percentage claim on this programme rested on episode COUNTS, which are not commensurable with seconds. Closes the #119 follow-up ("add per-cause TIME to ServeGateJournal").

Measured on soak-dl-master, daemonless, 36,854 s window (2026-08-09 03:26:23 → 13:40:37 UTC): only `qc-plan-underivable` carried a duration (5,526 s / 10 episodes); `creditpool-stale` (107 episodes), `emit-bestcl-null-committed` (2), `mn-needs-reseed` (1), `bestcl-stale` (1), `not-populated` (1) were count-only.

## Change (instrument only — no behaviour change)

- `ServeGateJournal::rollup(now)` returns per-cause attributed seconds (closed segments + the currently-open segment's partial, summed from the per-cause segment clock — never `episode_sec`) plus the wall-clock DENOMINATOR observed since the first observation.
- `seg_terminal_name()` names every closing segment in the qc classifier's shape (`resumed` / `cause-change`).
- `work_source` emits `[EMBED-GATE-SEG] cause=<name> dur=<n>s terminal=<...>` on every segment close — a SEPARATE line from the episode line, so per-cause seconds and whole-episode seconds are never mixed and `grep 'cause=... dur='` sums to the correct per-cause histogram. Adds `cause_dur=` to the decline line, and a periodic `[EMBED-GATE-ROLLUP]` that prints the denominator alongside the numerators (a percentage with an implicit denominator is how the best-share bug shipped).

## Tests

New KATs in `test_dash_serve_gate_journal`: per-cause time attributed by TIME not count (the brief's high-count/low-time shape), open-segment partial, observed-wall-clock denominator, terminal naming. The six existing segment-model KATs are unchanged and still pass. Header-only suite verified locally (11/11); `work_source.cpp` syntax-checks clean against the project include set.

## Not a gate

PR-1 and PR-2 touch different files and can merge in any order; what waits on this is the CLAIM about what they bought, not their merge. HEAD asserted against origin/master (2a051f5f); no literal commit pins.

### DIFF
--- a/src/impl/dash/coin/serve_gate_journal.hpp
+++ b/src/impl/dash/coin/serve_gate_journal.hpp
@@ -36,8 +36,12 @@
 /// seconds), no coin state. Header-only so it folds into the already
 /// allowlisted dash test targets.
 
+#include <algorithm>
 #include <cstdint>
+#include <map>
 #include <string>
+#include <utility>
+#include <vector>
 
 namespace dash {
 namespace coin {
@@ -143,6 +147,9 @@ public:
     Decision observe(Served served, const std::string& cause, int64_t now_sec) {
         Decision d;
         d.previous_cause = m_last_cause;
+        // The roll-up denominator starts at the very first observation, serving
+        // or not — it is wall clock OBSERVED, not wall clock declined.
+        if (m_first_observed_sec < 0) m_first_observed_sec = now_sec;
 
         if (served == Served::Embedded) {
             const bool was_declining = (m_arm == Arm::Declining);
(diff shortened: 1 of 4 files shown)


frstrtr:
Context for why this instrument is the first of the series, from the measurement that motivated it.

`qc-plan-underivable` is the ONLY refusal cause that currently carries a duration, and that is exactly why it is the only one whose story we can tell. On `soak-dl-master` (master, daemonless, no `--coin-rpc`), window 2026-08-09 03:26:23 → 13:40:37 UTC = 36,854 s, it cost **5,526 s = 15.0 % of wall clock** — a number that exists only because someone wired `dur=` and `terminal=` into that one path.

Against it, `creditpool-stale` produced **107 episodes and zero measured seconds**. It is the dominant cause by count and completely dark by time. Same for `emit-bestcl-null-committed`, `mn-needs-reseed`, `bestcl-stale`, `not-populated`.

That asymmetry is the whole reason this PR is priority zero: until it lands, no one can attribute serving loss to a cause, and any percentage claimed for a fix is resting on episode counts.

The qc path is also the model to copy, not just the exception to fix. Its `terminal=real-arrived` label is what made #1203 provable — every single episode ends because the real commitment arrived, which is how we learned the refusal is propagation wait rather than a missing mechanism. A named terminal state turned a suspected defect into a measured cost. Please keep that shape for every cause you instrument.

Refs #1203.

PR #1207, opened 2026-08-09
dash(mempool): gate serving on tx validity instead of set membership

frstrtr:
## The defect

The condition blocking `--embedded-serve-mempool-txs` was a **set-membership test**: `ours_only == 0` in the `[SHADOW-TXSET]` comparison against dashd's template. It is wrong in two directions at once.

**1. UNREACHABLE.** `ours_only` counts transactions in *our* selection that are absent from *dashd's template*. Two independently-connected nodes never hold identical mempools — relay is gossip, admission is per-node policy over a per-node UTXO view, and the two templates are snapshots taken at different instants.

Measured on the production hotel, **6056 samples**:

| | |
|---|---|
| `ours_only == 0` | 91.2% |
| `ours_only > 0` | 8.8% |
| mean when non-zero | 65.3 |
| max | 287 |
| zero rate over the **last 200** samples | **37%** |

A gate whose passing condition is a coincidence is not a gate. And the "it's usually zero, so it's nearly satisfied" reading dies on that last row.

**2. BLIND to the case that costs money.** `ours_only == 0` says our set is a *subset* of dashd's. It says nothing about whether the members are **acceptable**. A strict subset can still contain a transaction dashd would refuse — a double-spend of a coin our partial UTXO view still shows unspent, a non-final tx, a policy-rejected script — and mining it costs the whole block. The membership test **passes** and the block is **still rejected**. The "strict subset, therefore safe" shortcut does not even hold empirically on the numbers above.

## The condition implemented

For every transaction we hold **and would serve**, ask dashd's `testmempoolaccept` and apply three-valued logic:

```
allowed == true                                         -> VALID
allowed == false && reason == "txn-already-in-mempool"   -> ALSO VALID
allowed == false && any other reason                     -> INVALID, and a defect
```

"Already in mempool" is dashd telling us it *already holds* this transaction — the strongest possible statement that it is acceptable to it. It is exempted **by exact name**, and nothing else is: no reason-prefix matching, no "looks benign" family. `txn-already-known` is a *different* dashd reason (transaction in a block / recently rejected) and is **INVALID** — a tx we would serve that dashd has already mined is a staleness defect worth stopping on. There is a KAT for that, because widening the exemption set is widening the gate.

## The hard gate, and the window

**ZERO INVALID over 576 consecutive EVIDENCE-BEARING heights.** One INVALID anywhere in the run resets it to zero.

Why 576, stated rather than assumed:

* **Duration.** DASH mainnet targets 2.625 min/block, so 576 **distinct** heights is **at least ~25.2 hours** — one full day plus margin, and longer whenever heights go by without evidence. That is the shortest window containing every *daily* period in mempool composition: the fee/traffic diurnal cycle, at least one full DKG mining window per active LLMQ type, at least one ChainLock-less stretch. A shorter window can be passed by a quiet night. **This bound holds only because the counter advances per distinct height** — see "One height counts once" below.
* **Statistical power, stated at the unit the window actually counts.** Zero INVALID heights in 576 evidence-bearing heights bounds the per-**height** rate at 3/576 (rule of three, 95%) = **5.2e-3**: at most a **≤0.52% chance that any given template we build carries a rejectable transaction — roughly one at-risk block per 192 mined.** That is the **residual**, and it is not zero. It is why the number is 576 and not 100.

  The per-**transaction** form of this claim is deliberately *not* made. The window does accumulate ~1.7e4 probes (~30 selectable tx per evidence-bearing height on the hotel), but those are not 1.7e4 *independent* trials: a transaction that sits in the mempool for several blocks is re-probed at each of them. Only the heights are separated by a fresh tip, a fresh selection and a fresh mempool, so only the height-level bound is claimed.
* **Reachability.** Unlike `ours_only == 0`, a correct node can actually satisfy this: it asks whether *our* transactions are acceptable, not whether two mempools coincide.

The window is stated in **blocks** because the evidence is per template height; hours are the derived reading.

### One height counts once

`apply()` is fed **one sample per probe run, and a probe run is not a block.** `probe_validity()` runs from `process()`, `process()` runs from `on_serve`, and `on_serve` fires on *every template re-source*: every work-generation bump plus a 30 s staleness re-poll (`work_source.cpp` `kStaleAfter`). At 157.5 s/block that is roughly **five to six samples on one height**.

Counting samples would therefore have made 576 counter ticks mean **~110 real heights — about 4.8 hours, not 25.2** — and a node parked on a frozen tip could tick the counter to 576 without ever advancing a block. That is precisely the "a shorter window can be passed by a quiet night" failure the threshold exists to exclude. Measured at its true unit, the sample-counting window bounded the per-template rate at 3/110 = **2.7%, one at-risk block per 37 — 5.2x worse than the number the threshold was argued from.**

So the window counts **distinct heights**, over a **monotone cursor**, and the rule is deliberately **asymmetric**:

* A sample at a height that does not advance the cursor (a repeat of the last counted height, or a rollback below it) contributes **no new evidence**: it cannot advance `consecutive_clean`, cannot raise `heights_with_evidence`, and its transactions are kept out of the independent-trial totals.
* That same sample **can still kill the run**. An INVALID is an *observed defect* whenever it is observed — the mempool at one height is not constant, and a conflicting spend arriving between two probes of the same height is exactly the event that costs a block. Ignoring it because "we already counted this height" would trade over-counting clean evidence for under-counting defects: one blindness for another.

A repeat can therefore only ever **hurt** the run, never help it. Nothing is erased: `samples_seen`, `samples_repeat_height`, `repeat_txs_probed` and a per-line `sample=COUNTED | REPEAT-HEIGHT(no-advance) | NO-EVIDENCE(no-advance)` disposition keep every probe on the record — dropped from the window arithmetic, never dropped from the log.

A height whose *first* probe found nothing is **not** burned: the cursor only moves on evidence, so a later probe of that same height that does find transactions is the first evidence there and counts.

**Evidence-bearing** is load-bearing. A height where the probe set was empty — or where every entry came back with no answer — observed *nothing*. Counting it clean would let 576 empty heights open the gate on zero evidence: the same vacuous pass the old condition had, sign-flipped. Such heights neither advance nor reset the run; they are counted and logged separately.

## ours_only is now INFORMATIONAL

Still measured, still printed on every `[SHADOW-TXSET]` line, as a coverage statistic. It blocks nothing. The line now says so, in place of the old "do NOT enable --embedded-serve-mempool-txs until this is zero".

## Refusals name themselves

Per the #1038/#1039 serve-gate discipline, a decline that does not name its cause and bound is a silent decline:

```
[MEMPOOL-VALIDITY] h=2518789 REFUSED cause=tx-invalid txid=<64 hex>
  reason=<dashd's string, verbatim> threshold=576-consecutive-clean-heights
  clean_run=0/576 — dashd's testmempoolaccept says this transaction is NOT
  acceptable; serving it can cost a whole block. --embedded-serve-mempool-txs
  stays OFF and the clean run RESETS to 0
```

Plus a per-height summary line so a soak can read the series, and a `mempool-validity-gate` object in the shadow-compare stats JSON.

## The flag stays DEFAULT-OFF

This PR changes the **condition** that decides when `--embedded-serve-mempool-txs` may be turned on. It does not turn it on, and nothing in it can change, delay, or re-order a served template. The probe runs on `EmbeddedShadowCompare`'s existing worker thread — off the miner-facing path, exactly like the oracle fetch. With no dashd RPC arm bound the gate is unevaluated and therefore **CLOSED**, which is the correct reading of "we could not ask".

## Mechanics worth reviewing

* **`mempool_validity_gate.hpp`** — pure classify / aggregate / window logic + probe-set assembly + log lines. Header-only so it is KAT-pinnable without a node or a daemon.
* **`NodeRPC::test_mempool_accept`** — **one transaction per call**, because that is the only shape Dash Core's `testmempoolaccept` takes ("Array must contain exactly one raw transaction"). Never throws; a transport error, a down daemon, or a malformed answer all return a **null** answer, which classifies **UNPROBED**. Returning anything readable as `allowed` would turn "we could not ask" into "dashd said yes" — the exact failure this gate exists to remove.
* **What gets probed.** Only the **mempool-sourced** transactions. A served body is `[type-6 quorum commitments][pinned local txs][mempool txs]`; the builder records that last range as a *range* (`m_mempool_tx_first_index` / `_count`) rather than duplicating the block body for a diagnostic. Commitments and pins are refused by relay policy **by design** (a zero-fee pin especially) — probing them would manufacture INVALID verdicts out of correct behaviour.
* **Evaluable while the flag is OFF.** With serving disabled the served set is empty by construction, so the candidate set — what selection *would* have chosen — now carries its wire hex. Still never served: no consensus path reads it, and the coinbase value and merkle are byte-identical to a build with that line deleted.
* **The package artefact, handled honestly — and narrowly.** A child probed alone answers `missing-inputs` when its parent rides the same template but is not yet in dashd's mempool. That is an artefact of the one-tx probe shape — a block carrying parent *and* child is perfectly valid, and selection is topological (mempool.hpp G1) so the parent *is* in the template. Those are classified UNPROBED, but **only** when the builder recorded the dependency from the real vins. The same reason string with no recorded dependency is a genuine missing input and is **INVALID**.

  **`bad-txns-inputs-missingorspent` is not in the exemption.** `depends_on_in_set_parent` is a fact about the *transaction*, not about the *input that failed*, and `testmempoolaccept` never names which outpoint it refused — so a recorded dependency does not imply the failing input was that one. A transaction spending one in-set parent output **and** one genuinely double-spent output carries the flag and is a real defect: the exact loss this gate exists to prevent, classified UNPROBED and waved through.

  A per-**input** fix is not available from the data dashd returns: even with the full per-input in-set map on our side, the answer is a bare reason string with no outpoint, so nothing can attribute the failure to a specific input. The exemption is therefore **narrowed** to the one reason whose meaning is unambiguous — `missing-inputs` = "this input is not known to me". `bad-txns-inputs-missingorspent` conflates unknown with **spent** in its own name and is now **INVALID even with the flag set**. The cost is stated rather than hidden: an in-set-parent child that dashd answers with the conflating reason will reset the clean run. That fails toward a **closed** gate on ambiguous evidence, which is the direction a gate is allowed to be wrong in.

## Testing

`test_dash_embedded_gbt`: **270/270 green**, including

* **24 gate KATs** (`test_dash_mempool_validity_gate.cpp`): the three-valued logic; the exact-name exemption and three near-miss strings that must *not* be exempted; no-answer and short-answer-vector both UNPROBED; the in-set-parent excuse and its negative cases; the refusal line's txid/reason/threshold; the full clean run and the one-short case; one INVALID killing the run with the high-water mark surviving; empty heights neither advancing nor resetting; an all-UNPROBED height not counting as evidence; probe-set assembly from both template shapes and the mis-populated-WorkData case that reports nothing rather than guessing.
* **6 of those 24 are new, and were written to fail first.** The pre-existing suite could not have caught the sample-vs-height defect: its `feed_clean()` helper always advances `from_height + i`, so it never presented the input that breaks the counter. The new KATs feed **repeats**: five probes of one height advancing the window once; a whole window (`required + 50`) of samples at **one** height leaving the gate CLOSED; an INVALID on an already-counted height still killing the run while a clean re-probe does not resurrect it; a rollback/alternation not counting a height twice; an empty first probe not burning its height; and `bad-txns-inputs-missingorspent` staying INVALID with an in-set parent while the narrow `missing-inputs` stays excused.

  **Verified red before the fix, not asserted red.** Built against the unfixed header, 5 of the 6 fail — the headline being `AWholeWindowOfSamplesAtOneHeightCannotOpenTheGate`, where **626 counter ticks at a single height opened the gate**:

  ```
  [ RUN      ] DashMempoolValidityGate.AWholeWindowOfSamplesAtOneHeightCannotOpenTheGate
  Value of: g.open()
    Actual: true
  Expected: false
  one height, however often probed, is one height
    g.consecutive_clean
      Which is: 626
    1u
    g.heights_with_evidence
      Which is: 626
    1u
  [  FAILED  ] DashMempoolValidityGate.AWholeWindowOfSamplesAtOneHeightCannotOpenTheGate
  ...
   5 FAILED TESTS
  ```

  (The sixth, `AnEmptyProbeAtAHeightDoesNotBurnThatHeight`, passes both before and after: it is a guard against the fix over-correcting, not a defect witness.) With the fix applied the same six are green.
* **Both directions the old condition got wrong, as explicit KATs**: 287 ours-only transactions that dashd accepts is a **clean** height; a strict subset (`ours_only == 0`, old gate passes) containing one refused transaction **closes** the gate.
* **3 new builder KATs** in `test_dash_embedded_gbt.cpp`: candidate wire hex present when suppressed; served range naming the mempool txs without duplicating them; the range skipping a mandatory type-6 commitment prefix.
* The two shadow-compare tests that asserted the old blocking text now assert the informational reframe.

Also built and linked clean: `c2pool-dash`.

## What this does NOT do

It does not arm the flag, and it does not claim the gate is *satisfied*. The gate has not yet run for a single window; it starts CLOSED and stays CLOSED until 576 consecutive evidence-bearing **distinct** heights say otherwise. Arming remains an operator decision on top of that.

### DIFF
--- a/src/c2pool/main_dash.cpp
+++ b/src/c2pool/main_dash.cpp
@@ -84,6 +84,7 @@
 #include <impl/dash/coin/tip_ingest.hpp>         // wire_tip_ingest (leg 2)
 #include <impl/dash/coin/embedded_oracle_shadow.hpp> // dash::coin::EmbeddedOracleShadow — per-block dashd cross-check (OBSERVE-only)
 #include <impl/dash/coin/embedded_shadow_compare.hpp> // dash::coin::EmbeddedShadowCompare — serve-vs-dashd template diff (OBSERVE-only, NOT a gate)
+#include <impl/dash/coin/mempool_validity_gate.hpp>   // dash::coin::MempoolValidityGate — THE --embedded-serve-mempool-txs arming condition
 #include <impl/dash/coin/block_connect_ingest.hpp>   // wire_block_connect_ingest (leg 3)
 #include <impl/dash/coin/mn_list_ingest.hpp>     // wire_mn_list_ingest (leg 4)
 #include <impl/dash/coin/govsync_ingest.hpp>     // wire_govobject_ingest / wire_govvote_ingest (E-SUPERBLOCK)
@@ -788,7 +789,9 @@ int run_node(bool testnet, const std::string& rpc_endpoint,
              // on changes what this node asks its peers for. It does NOT by
              // itself put a single transaction into a served template — that
              // remains --embedded-serve-mempool-txs' decision, gated on the
-             // [SHADOW-TXSET] coverage series.
+             // [MEMPOOL-VALIDITY] testmempoolaccept series
+             // (mempool_validity_gate.hpp) -- NOT on the [SHADOW-TXSET]
+             // ours_only coverage number, which is informational.
              bool embedded_mempool_ingest = false,
              // --pin-local-tx-hex <file>: PINNED LOCAL TX (donation-dust
              // consolidation). File holds the hex of an externally-signed,
@@ -2085,6 +2088,14 @@ int run_node(bool testnet, const std::string& rpc_endpoint,
     // implicit consequence of UTXO maturity. Audit:
     // DASH_CONNECTBLOCK_REJECT_SURFACE_AUDIT.md §1. The dashd fallback arm is
     // unaffected (its GBT bodies come from dashd itself).
+    //
+    // THE CONDITION for that opt-in is the MEMPOOL VALIDITY GATE
+    // (mempool_validity_gate.hpp): zero transactions refused by dashd's
+    // testmempoolaccept over a sustained window of consecutive
+    // evidence-bearing heights. It is NOT `ours_only == 0` -- that
+    // set-membership test is unreachable (two independently-connected mempools
+    // never coincide) and blind to a strict SUBSET that contains an invalid
+    // transaction, which is exactly the case that costs a whole block.
     node_coin_state.set_serve_mempool_txs(embedded_serve_mempool_txs);
     std::cout << "[run] embedded mempool-tx serving: "
               << (embedded_serve_mempool_txs
@@ -2611,11 +2622,28 @@ int run_node(bool testnet, const std::string& rpc_endpoint,
                 [rp = rpc.get()]() -> std::optional<dash::coin::DashWorkData> {
                     try { return rp->getwork(); }
                     catch (...) { return std::nullopt; }
+                },
+                // AcceptFn: dashd testmempoolaccept for ONE transaction we hold
+                // and would serve. THE condition that decides when
+                // --embedded-serve-mempool-txs may be armed, replacing the
+                // unreachable-and-blind `ours_only == 0` set-membership test.
+                // Runs on the SAME worker thread as the oracle fetch, so it is
+                // off the miner-facing path; a null answer counts UNPROBED,
+                // never as a pass. It arms NOTHING -- the flag stays OFF.
+                [rp = rpc.get()](const std::string& raw_tx_hex) -> nlohmann::json {
+                    return rp->test_mempool_accept(raw_tx_hex);
                 });
             work_source->set_shadow_compare(std::move(shadow));
             std::cout << "[run] --embedded-shadow-compare ARMED: OBSERVE-only "
                          "serve-vs-dashd template diff (worker-thread dashd fetch; "
-                         "NOT a serve gate; [SHADOW] log lines + counters)\n";
+                         "NOT a serve gate; [SHADOW] log lines + counters)\n"
+                      << "      [MEMPOOL-VALIDITY] gate ARMED with it: every tx we "
+                         "would serve goes to dashd testmempoolaccept. Condition "
+                         "for arming --embedded-serve-mempool-txs = ZERO INVALID "
+                         "over "
+                      << dash::coin::MempoolValidityGate::kCleanHeightsRequired
+                      << " consecutive evidence-bearing heights. ours_only is "
+                         "INFORMATIONAL and gates nothing.\n";
         } else {
             std::cout << "[run] --embedded-shadow-compare given but no dashd RPC "
                          "arm is armed (pure-daemonless) -- no oracle to compare "
@@ -4982,8 +5010,10 @@ int run_node(bool testnet, const std::string& rpc_endpoint,
                          " The mempool will now FILL from the DASH network.\n"
                       << "      This does NOT put transactions into served"
                          " templates: that stays --embedded-serve-mempool-txs"
-                         " (default OFF), gated on the [SHADOW-TXSET]"
-                         " coverage series.\n";
+                         " (default OFF), gated on the [MEMPOOL-VALIDITY]"
+                         " testmempoolaccept series -- zero INVALID over "
+                      << dash::coin::MempoolValidityGate::kCleanHeightsRequired
+                      << " consecutive evidence-bearing heights.\n";
         }
 
         // Kick the initial sync once the version/verack handshake completes:
@@ -7303,7 +7333,9 @@ int main(int argc, char** argv)
     // Default OFF = coinbase-only serving (values exact, fees forgone); the
     // mempool-tx body path with its G1-G4 guards (mempool.hpp; audit
     // DASH_CONNECTBLOCK_REJECT_SURFACE_AUDIT.md) arms only on explicit
-    // operator decision.
+    // operator decision, and only once the MEMPOOL VALIDITY GATE
+    // (mempool_validity_gate.hpp) reports zero transactions refused by dashd's
+    // testmempoolaccept over its sustained window.
     bool embedded_serve_mempool_txs = false;
     // --embedded-creditpool-publish-at-serve-tip: publish the derived credit
     // pool AT THE SERVE TIP rather than at the folded body height (PR-5,
@@ -7318,7 +7350,10 @@ int main(int argc, char** argv)
     // SEPARATE from --embedded-serve-mempool-txs on purpose: this only makes
     // the mempool FILL. Whether its contents ever reach a served template is
     // still the other flag's decision, and that one stays default-OFF until
-    // the [SHADOW-TXSET] coverage series says it is safe.
+    // the [MEMPOOL-VALIDITY] series says it is safe: ZERO transactions that
+    // dashd's testmempoolaccept refuses, over a sustained window
+    // (mempool_validity_gate.hpp). NOT the [SHADOW-TXSET] ours_only number --
+    // that is a coverage statistic and gates nothing.
     bool embedded_mempool_ingest = false;
     std::string bestcl_policy = "freshness";   // --bestcl-policy: freshness (default, conservative proxy) | consensus-exact (dashcore's actual CheckCbTxBestChainlock rule)
     bool embedded_oracle_shadow = false;       // --embedded-oracle-shadow: per-block dashd cross-check (OBSERVE-only)
--- a/src/c2pool/main_dash.cpp
+++ b/src/c2pool/main_dash.cpp
@@ -2642,7 +2642,9 @@ int run_node(bool testnet, const std::string& rpc_endpoint,
                          "for arming --embedded-serve-mempool-txs = ZERO INVALID "
                          "over "
                       << dash::coin::MempoolValidityGate::kCleanHeightsRequired
-                      << " consecutive evidence-bearing heights. ours_only is "
+                      << " consecutive evidence-bearing DISTINCT heights (the "
+                         "probe fires ~5-6x per block; repeats do not advance "
+                         "the window). ours_only is "
                          "INFORMATIONAL and gates nothing.\n";
         } else {
             std::cout << "[run] --embedded-shadow-compare given but no dashd RPC "
@@ -5013,7 +5015,7 @@ int run_node(bool testnet, const std::string& rpc_endpoint,
                          " (default OFF), gated on the [MEMPOOL-VALIDITY]"
                          " testmempoolaccept series -- zero INVALID over "
                       << dash::coin::MempoolValidityGate::kCleanHeightsRequired
-                      << " consecutive evidence-bearing heights.\n";
+                      << " consecutive evidence-bearing DISTINCT heights.\n";
         }
 
         // Kick the initial sync once the version/verack handshake completes:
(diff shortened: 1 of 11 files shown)


frstrtr:
Adversarial review — **blocking**. Two real defects, and the first is the same class this PR exists to fix.

## 1. BLOCKING — the 576-window counts SAMPLES, not distinct heights

`MempoolValidityGate::apply()` (`mempool_validity_gate.hpp:289`) does `++heights_seen` / `++consecutive_clean` per sample, with **no comparison against the previous `s.height`**.

`probe_validity()` is called from `process()` on every `on_serve`, and `on_serve` fires on every template re-source (`work_source.cpp:1095`) — cadence `kStaleAfter = 30 s` (`work_source.cpp:68`) plus every work-generation bump. Machine-checked against this PR's own header: five samples at one height yield

```
distinct_heights=1  consecutive_clean=5  heights_with_evidence=5  txs_probed=5
```

**Consequence 1 — the duration argument does not hold.** DASH targets 157.5 s/block, so ~5-6 samples land per real height. 576 counter ticks is **~110 real blocks ≈ 4.8 hours**, not the ~25.2 h the header argues for. The header's own stated reason for choosing 576 — *"the shortest window containing every daily period … a shorter window can be passed by a quiet night"* — is precisely what the implementation permits.

**Consequence 2 — the statistical-power argument does not hold.** "Zero events in n trials bounds the per-tx invalid rate at 3/n (rule of three)" requires **independent** trials. Roughly 5/6 of the claimed ~1.7e4 probes are the same transaction re-probed at the same height (and mempool txs persist across heights besides). The quoted 1.7e-4 bound, and the "≤0.5 % per template / one at-risk block per 190 mined" residual, are inflated ~5x and correlated.

**The 18 new KATs cannot catch this**, because `feed_clean()` always advances `from_height + i` — the tests never present the input that breaks it.

Fix is a one-line height dedup in `apply()` plus a KAT that feeds repeated samples at one height.

This is worth stating plainly: a gate written to replace a wrong unit (set membership) shipped with a window measured in a wrong unit (samples per height). Same family, one layer down.

## 2. SECONDARY — the in-set-parent exemption is per-TRANSACTION, not per-INPUT

`mempool_validity_gate.hpp:179` excuses **any** missing-inputs-family reason (including `bad-txns-inputs-missingorspent`, `is_missing_inputs_reason` at `:100`) whenever `depends_on_in_set_parent` is set.

A transaction spending one in-set parent output **and** one genuinely double-spent output classifies `UNPROBED` instead of `INVALID`. A recorded dependency does not imply the **failing** input was that one. Narrow, but it is a hole exactly where the PR body argues one was closed.

## 3. MINOR — unconditional per-build packing

`embedded_gbt.hpp:636` packs and hex-encodes the entire candidate set on every template build; `build_embedded_workdata` has no knowledge of whether `--embedded-shadow-compare` is armed, so a node that never probes still pays it on the template-build path. Negligible at ~30 candidate tx, flagged because the 2026-08-07/08 tip-freeze came from this class of unconditional per-build work.

## What is clean, verified rather than assumed

- **Attribution**: no offending token in commit message, title or body; author and committer `frstrtr <frstrtr@gmail.com>`; the repo's own attribution-gate is PASS.
- **Not CONFLICTING**: `mergeable=MERGEABLE`; `BLOCKED` is awaiting approval, not a conflict.
- **No serve gate weakened, no null-commitment fallback.** `ours_only` was never enforcing code — a struct field and a log string; the "do NOT enable" text was documentation, never a branch. `MempoolValidityGate::open()` has **zero consumers** in `src/`; nothing arms on it yet. `test_mempool_accept` returns null json on failure, so a dead daemon classifies `UNPROBED`, never a pass.
- **Test-count claim is arithmetically honest**: 264 `TEST(` across the 17 sources of the target, 18 in the new gate file plus 3 new builder KATs.

**Do not merge until (1) is fixed.** (2) should ride the same push. CI was still running at review time, so "green" is not yet CI-confirmed either.

frstrtr:
Re-review after `11ef3aa2`: the blocking defect is **fixed**, and the one risk that could have made the gate unreachable is now **measured away**.

## The blocking defect, and what it actually meant

The pre-fix headline is worth recording because it is worse than "the window was short":

```
626 counter ticks at a SINGLE height,  g.open() == true
```

The gate could arm mempool serving on **one block**. Not a 4.8-hour window instead of 25.2 — a *one-height* window, if the tip happened to sit still.

Fixed with a monotone height cursor. Strictly-increasing was chosen over "differs from last" so an alternating reorg (H, H-1, H) cannot count one block twice; the cost is at most one lost height per rollback, which fails toward a **closed** gate.

The asymmetry I asked for is present and is the right way round: a repeat cannot advance the window, but the INVALID branch sits **before** the distinct-height test, so a defect on a repeat still resets the run. A repeat can only hurt, never help. Verified three ways by the reviewer, including a measured run: after four clean heights, an INVALID re-probe of the last one gives `consecutive_clean=0`, `heights_with_evidence` unchanged at 4.

Counters renamed so the names stop lying: `heights_seen` → `samples_seen`, plus `samples_repeat_height` and a printed disposition.

## CAVEAT B — checked against the hotel's own daemon, and it does not bite

The reviewer flagged, honestly and unmeasured, that the narrowed exemption keys on the literal `"missing-inputs"`, and that this string is an RPC-layer rewrite present in some Core versions and not others. If our dashd emitted only `bad-txns-inputs-missingorspent`, every in-set child would classify INVALID and **576 would be unreachable** — a failure that would surface only after a day of soak.

Measured on the production hotel just now, with a transaction spending a nonexistent outpoint:

```
$ dash-cli testmempoolaccept ...
  "allowed": false,
  "reject-reason": "missing-inputs"

$ dash-cli getnetworkinfo
  "subversion": "/Dash Core:23.1.7/"
```

Exactly the string the exemption keys on. The gate is reachable on the daemon we actually run. Worth re-checking on any dashd upgrade — the risk is real, it simply is not present at 23.1.7.

## CAVEAT A — the fix author's own claim was too strong, and the reviewer caught it

The claim "the 6 KATs were written using only pre-existing field names so they compiled against the unfixed header" is **false as committed** — the test file references `samples_seen`, `samples_repeat_height`, `last_disposition` and `SampleDisposition`, all added by `11ef3aa2`.

The redness proof survives anyway, because the reviewer did not rely on that claim: they reverted each hunk **themselves** on the finished branch. `advances = true` → 4 failed. Widening the exemption back → 1 failed. 4 + 1 = the 5 claimed.

And the sixth KAT was volunteered as **not** a defect witness: `AnEmptyProbeAtAHeightDoesNotBurnThatHeight` passes both before and after, because it guards against the fix over-correcting. Stating that rather than counting it is the difference between six red tests and five.

## Remaining, non-blocking

`CAVEAT C`: an evidence-bearing repeat sample's `unprobed` count reaches neither `txs_unprobed` nor a repeat-specific counter. Harmless to the window arithmetic, documented, no KAT. Fine to leave.

Registered at `test/CMakeLists.txt:863` inside `test_dash_embedded_gbt` — not the unregistered-binary fake-green pattern. 270/270, 24 in `DashMempoolValidityGate`.

**Unblocked from my side.** Merge once the full CI rollup is green on the current base.

PR #1236, opened 2026-08-14
dash(coin-p2p): header-sync stall watchdog — re-arm getheaders on the fresh bring-up path (#129)

frstrtr:
## Item #129 — FRESH daemonless node: coin-p2p header sync stops re-arming getheaders

**DASH lane only. No wire/consensus change. Draft — integrator review required, do NOT merge.**

### Defect

A FRESH cut node (EMPTY data-dir, NO `--coin-rpc`) wedged on coin-p2p header sync around h=2406000: `getheaders` stopped re-arming, the header chain never reached tip, the node could not bootstrap. Upgrading an EXISTING data-dir works (contabo cut node resumed a persisted cursor) — only the brand-new-from-empty path bites, which blocks standing up fresh public daemonless nodes.

### Root cause (main_dash.cpp, master f15d2310)

The tip-lane header sync is a **one-token relay with no token custodian**:

- **Armed once**: `set_on_handshake_complete` (main_dash.cpp:5379) fires `send_getheaders` — the same kick that also queues the cold-start monsters on the same primary: `mnlistdiff(base=null, ~3200 MNs)` + `qrinfo` + mempool.
- **Re-armed only by its own response**: the self-propel subscriber (main_dash.cpp:5160) re-issues `send_getheaders` **only when a non-empty headers batch arrives**, and `send_getheaders` (p2p_client.hpp:1296) targets **only `m_primary`** — silent no-op when the primary is null.
- **No timer, no stall watchdog, no height-vs-peer-tip condition anywhere.** Contrast the pre-anchor bulk walker, which has BOTH a self-propel AND a periodic re-kick backstop (`maybe_kick_backfill`, replay_bulk_fetch.hpp:1095, 15 s, peer round-robin). The tip lane has no analog.

On the FRESH path the token dies: dashd serves one peer's requests sequentially, so the next headers reply sits **minutes** behind the null-base mnlistdiff/qrinfo computation the same handshake queued. Live repro on vm905 (fresh empty dir, no `--coin-rpc`): first batch answered in 2 s, then **93 s of dead silence** (qrinfo + 3211-MN mnlistdiff arriving 09:08:29/09:08:35), after which sync only limped **one 2000-header batch per ~2.5-min block inv** via the per-block rescue (~19.5 hdr/s over a 121k deficit), with delayed duplicate batches (`accepted==0`) interleaving. Single-peer — or with the announcer falling back to the same starved primary — the rescue yields nothing and the chain parks forever. The historical wedge at h=2406000 = anchor 2400000 + 3 batches is exactly this signature. A warm data-dir never sees it: small deficit, no cold monsters (#1151 cursor restore).

Same **armed-once-never-re-armed** class as #1151 (cursor restore cold-start-only) and #1162 (`begin_fold` had to set `m_rerequest_from_cursor`).

### Fix

A 5 s watchdog timer on the coin-p2p arm — the tip-lane analog of `maybe_kick_backfill`:

- if `header_chain.height()` is **behind** the best peer-advertised height (`peer_tip_height`, stamped from version handshakes) **and** has not advanced for **20 s**, re-issue `getheaders` off the **current** locator;
- **round-robin across all handshaked peers** (`send_getheaders_to`), not pinned to the possibly-starved primary; primary fallback if no named peer takes it;
- min 10 s between kicks; logs `[COIN-P2P] header-sync stall watchdog: re-armed getheaders h=... peer_tip=... stalled_s=... peer=... kick#N`.

### Safety

- **No wire/consensus change** — only re-sends an existing, well-formed `getheaders`; a duplicate reply de-dups as `accepted==0`; at/above the advertised peer tip the height guard makes the watchdog inert (steady-state tip follow stays inv-driven, by design).
- **Reward safety unchanged** — no serve gate reads the watchdog; serving stays gated on chain freshness / `is_synced` (header_chain.hpp:486) exactly as before. A re-request grants no serve authority; a node still cannot serve from an incomplete header chain.
- **DASH lane only** — `main_dash.cpp` + the existing DASH `p2p_client.hpp` API; no other coin touched.

### Reproduction / proof (vm905)

- **Without fix** (binary 596a4b35, header path identical to f15d2310): `vm905:~/repro-129-fresh/{RUN.sh,run.log}` — fresh empty dir, no `--coin-rpc`; 93 s starvation after batch 1, then one-batch-per-block-inv limp (`rate=19.5hdr/s eta=1h40m`), duplicate `accepted==0` loops; the h=2406000 wedge signature.
- **With fix** (5ee18384): `vm905:~/repro-129-fix/{RUN.sh,run.log}` — same config, fresh empty dir; watchdog fires during the cold-monster starvation window and drives header sync continuously to the live tip.

(Exact with-fix log excerpts in the PR comments after the soak run completes.)

### DIFF
--- a/src/c2pool/main_dash.cpp
+++ b/src/c2pool/main_dash.cpp
@@ -3420,6 +3420,10 @@ int run_node(bool testnet, const std::string& rpc_endpoint,
     // Phase-1 mempool-ingest telemetry (30s); constructed only under
     // --embedded-mempool-ingest.
     std::unique_ptr<core::Timer>                               mempool_ingest_timer;
+    // #129 fresh-path header-sync stall watchdog (coin-p2p arm only).
+    // Declared after header_chain/coin_p2p so it unwinds FIRST at return —
+    // its callback reads both.
+    std::unique_ptr<core::Timer>                               header_sync_watchdog_timer;
     // W5 integration: the fold engine the lane drives, and its consumer.
     std::unique_ptr<dash::coin::replay::DmlFoldEngine>         replay_fold_engine;
     std::unique_ptr<dash::coin::replay::FoldReplayConsumer>    replay_fold_consumer;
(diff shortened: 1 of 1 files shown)


frstrtr:
## Fresh empty-data-dir bring-up proof (vm905, daemonless, NO `--coin-rpc`)

Two identical fresh bring-ups (EMPTY data-dir, same 4 pinned peers + discovery, 8/8 handshaked, same host, same wall-clock window). Only variable: the binary (fix absent/present). Deficit in both: anchor 2400000 -> live tip ~2521650 = **~121,650 headers**.

### WITHOUT fix — `596a4b35` (header path byte-identical to master f15d2310) — `vm905:~/repro-129-fresh/run.log`

- Start 09:06:27. **68 min later (10:14): 34,008 headers accepted = 28% of the deficit.** Average **8.3 hdr/s**; batches arrive only at block-inv cadence (the per-block rescue), with **dead gaps up to 13 min** (09:23:39 -> 09:36:47 -> 09:46:27) and long `accepted==0` duplicate loops between them. Extrapolated finish: **~4 h** — and only because 8 peers keep announcing invs. Single-peer, the rescue never fires and the chain parks forever: the historical h=2406000 wedge (anchor + 3 batches, then silence).
- `stall watchdog` lines: **0** (mechanism absent).

### WITH fix — `5ee18384` — `vm905:~/repro-129-fix/run.log`

- Start 09:46:29, first accepted batch 09:46:36.
- **kick#1 fired 09:46:59** — inside the cold-seed starvation window (`mnlistdiff(null-base) +3213 MNs` applied 09:46:42, qrinfo behind it) — and the watchdog then carried the lane end-to-end: one accepted 2000-header batch per ~25–30 s, **69 kicks round-robined across all 8 handshaked peers** (8 distinct `peer=` targets in the log), sustained **~72 hdr/s**.
- **Live chain tip reached at 10:14:43** (`tip advanced h=2521658 ... new_tip fired`): **28 min 14 s** for the full ~121.7k catch-up — at the same wall-clock the baseline sat at 28%. ~9x the baseline rate; vs infinity in the single-peer wedge case.
- After tip: watchdog **quiet by construction** (`h >= peer_tip` guard) — kick#69 (10:14:31) is the last line; steady-state tip follow is inv-driven exactly as before.
- The run also shows why "bridge the initial 90 s window" would have been insufficient: the SML follower issues a further `mnlistdiff` at every checkpoint the header advance crosses (h=2402000, 2410000, 2414000, 2416000, ...), so the primary stays busy for the WHOLE catch-up. dashd's stance — never wait on a starved sync peer, ask another — is the only shape that works, and is what this ports.

### Reward-safety check

- Watchdog state is a lambda-local struct in `main_dash.cpp`; **no serve gate reads it** (`grep header_sync_watchdog` outside main_dash.cpp: 0 hits). No wire/consensus change: it re-sends an existing well-formed `getheaders`; duplicate replies de-dup as `accepted==0` (visible in both logs, harmless).
- Serving remains gated on chain freshness (`is_synced`, header_chain.hpp:485-492) and the serve gates exactly as before — observable in this very run: at 10:16:03 the node **demotes its serve tip** while the h=2521658 tip body is pending (`tip-body-overdue ... demoting the serve tip`). A fully header-synced node still refuses to serve past what it can prove. The watchdog only shortens the time to REACH the gates; it grants no serve authority.

### dashd alignment (HYBRID port, per the recon verdict)

Ported skeleton is dash-core `net_processing.cpp`: the tick-driven re-arm (`SendMessages` re-evaluating `fSyncStarted`, :6060-6088), the per-sync-peer headers timeout (:109-110, :6462-6493) taken in its **NoBan branch** — re-arm WITHOUT disconnect (:6477-6486) — because our primary also carries the MN cold-seed lane and must not be dropped, plus the `MaybeSendGetHeaders` pacing gate (`HEADERS_RESPONSE_TIME`, :3130-3146). Peer rotation replaces dashd's disconnect-then-reselect with select-without-disconnect; timeout scale is seconds (intra-startup starvation), not dashd's 15 min IBD scale.

PR #1237, opened 2026-08-14
dash(fees): W5-A fold-backed fee pricing behind a hash_serialized_2 graduation gate

frstrtr:
## W5-A: fold-backed fee pricing behind a hash_serialized_2 graduation gate

**Goal (dashd-cut Tier-2 "W5", leg A):** close the DOMINANT dashd-only tx class behind the `gbt-xcheck-txmerkle` mismatches — 20/34 txs across the 26-event diagnosis are ordinary fee-payers spending coins OLDER than the replay horizon. The forward-built `UTXOViewCache` cannot price them, `compute_fee_locked` marks them fee-unknown, the conservative exclusion drops them, embedded serves a strict SUBSET of dashd, and the guard swaps to the dashd arm. This PR wires the W3 full-history UTXO fold (`replay_utxo_fold.hpp`) in as the mempool's THIRD coin source — behind a reward-safety graduation gate.

### DARK / flag-gated
`--embedded-utxo-fold-fees DBPATH` + **required** `--embedded-utxo-fold-expect HEX` (the operator RESTATES the anchor hash — a stale/foreign DB cannot graduate on its own say-so). **Default OFF: nothing is constructed, `Mempool::m_fee_coin_lookup` stays empty, production byte-identical.** DASH lane only. No wire/consensus/share-format change.

### The graduation gate (reward-safety)
A wrong fee VALUE can overstate coinbasevalue = an invalid found block = a lost block — the tx-merkle guard does not cover fee values. So fold answers are arithmetically unreachable from the fee path until ALL of:

1. **GRADUATED** — `ReplayUtxoFold::try_graduate()` runs with the cursor standing exactly on the pinned anchor (h=2,516,758) and proves `hash_serialized_2` byte-equal to dashd's `gettxoutsetinfo` (`3d14913768a9d492bfa7a42fe9b111cff625b80e35bb4133e1d60cf3991c2319`). PASS writes the durable `'G'` record (format v2); FAIL **poisons** the fold (latched refusal, nothing written). Restart restores trust ONLY from a recorded pass that also matches the operator restatement.
2. **LIVE at the tip** — fold cursor within 2 blocks of the last observed tip (a catching-up fold could hold since-spent coins ⇒ superset ⇒ guard trip).
[...]

PR #1238, opened 2026-08-14
dash(dstx): W5-B CoinJoin broadcast-tx intake — inv pull, BLS operator-sig gate, zero-fee prioritised admission

frstrtr:
## W5-B: CoinJoin DSTX intake — inv pull, BLS operator-sig gate, zero-fee prioritised admission

**Goal (dashd-cut Tier-2 "W5", leg B):** close the second dashd-only tx class behind the `gbt-xcheck-txmerkle` mismatches — 14/34 txs across the 26-event diagnosis are zero-fee CoinJoin DSTX-es. dashd admits every BLS-verified `CCoinJoinBroadcastTx` at base fee 0, prioritises it +0.1 COIN (`net_processing.cpp:3609`) and sorts on `GetModFeesWithAncestors`; embedded never even pulled the object. Mirrors the #1232 isdlock architecture at every hop, fail-closed at every arm. Necessary-but-not-sufficient by the diagnosis: this leg alone closes 0/26 events (every event also contains ≥1 fee-unknown ordinary tx — the sibling W5-A fold PR).

### Load-bearing corrections vs the original brief
- **MSG_DSTX = 16**, not 30 (dashd `protocol.h:507`; 30 is the RETIRED legacy `MSG_ISLOCK` — a getdata for 30 gets notfound from every v23 peer).
- **The DSTX inv hash is the PLAIN TXID** (`net_processing.cpp:2567`), not `SerializeHash(payload)` — so the pull shares the budgeted MSG_TX machinery keyed by txid.

### DARK / flag-gated
`--embedded-ingest-dstx`, **default OFF**: no getdata type-16 is ever sent, the handler discards, wire and templates byte-identical to master. Unlike isdlock the GETDATA itself is gated — a DSTX has no fee-only-safe unconditional consumer. Template EFFECT additionally requires `--embedded-serve-mempool-txs`. DASH lane only; no wire/consensus/share-format change. **The #1218 gbt-xcheck-txmerkle guard is untouched and stays the referee.**

### Pipeline (acquire → verify → use, fail-closed)
1. **Wire**: `dstx = 16` inv type; `message_dstx` codec (v23 `CCoinJoinBroadcastTx`: tx, protxHash, vchSig, sigTime); UNCONDITIONAL registry membership (#1077 silent-drop rule — the flag gates behaviour, never parsing). Pull rides the MSG_TX budget (strict tip-body priority, inflight cap, txid-keyed slots; getdata echoes the announced type). Handler releases the slot by COMPUTED txid, applies dashd's `IsValidStructure` refusals (KAT-pinned free predicate `dstx_is_valid_structure`: vin∈[3,180], vin==vout, all-denominated P2PKH vouts, 96B sig), fires `new_dstx` only when armed.
2. **Verify**: `CoinStateMaintainer::on_new_dstx` — UNSET verifier IS FAIL-CLOSED. The installed verifier BLS-verifies the mixing MN's OPERATOR signature over the KAT-pinned pre-image `dstx_signature_hash` = SHA256d(ser(tx) ‖ protxHash ‖ sigTime LE) (SER_GETHASH excludes vchSig, `coinjoin.cpp:68-80`) against the SML `pubKeyOperator(protxHash)` — REUSING `verify_govvote_operator_sig` (the byte-identical `VerifyInsecure` contract, key wire-scheme vs post-V19 basic signing handled inside).
3. **Use**: `Mempool::add_dstx` — bounded `mapDeltas` (recorded pre-admission, retro-applied if the plain tx arrived first, pruned on confirm); fee=0 forced only when inputs are unpriceable (**understate-only**: a hostile signed nonzero-fee DSTX can only understate `total_fees` ⇒ coinbase pays ≤ allowed ⇒ never an invalid block). **SCORE everywhere MODIFIED, ACCOUNT everywhere BASE** — dashd's exact split: feerate index, ancestor `modfee_wa` (== `nModFeesWithAncestors`), `ModEntry`, `update_for_parent_inclusion`, and the blockMinFeeRate gate use `modified_fee()`; `total_fees` and `SelectedTx.fee` stay base (coinbasevalue byte-parity). `delta==0` for every non-DSTX entry keeps all ordering byte-identical to master.

### Documented phase-1 divergences (both guard-covered, subset/superset only, never validity)
- **No per-MN rate limit** (dashd `IsValidForMixingTxes`, MAX=5 since last DSQ): mirroring the counter without the DSQ leg (inv 32) would only ever increment ⇒ permanent false refusals. An MN spamming >5 DSTX between queues makes embedded a SUPERSET; the tx-merkle guard refuses serving. Phase 2 = the DSQ leg if soak attributes mismatches here.
- **Current-SML-only operator lookup** (dashd walks ≤24 blocks back for recently-removed MNs): we refuse those ⇒ SUBSET for that rare window; phase 2 = operator-key tombstone ring.

### KATs (`test_dash_dstx.cpp`, new target)
Registry gate (RED if de-registered); codec golden decode + byte-exact re-encode (hand-assembled wire bytes); `GetSignatureHash` pre-image vs an independently hand-assembled stream (also pins `pack(tx)` network serialization); `IsValidStructure` refusal matrix; **the USE red/green** — plain zero-fee admission stays OUT on the min-feerate floor (master's behaviour) vs the verified maintainer path ranking it FIRST on the +0.1 COIN modified fee with `total_fees` still base; fail-closed — no verifier / bad BLS sig through the real `verify_govvote_operator_sig` entry point admits nothing; unpriceable-inputs admission at exactly fee=0.

### Measurement plan
Rides the W5-A plan (same canary posture, `--coin-rpc` kept, guards active): AFTER-2 arms both flags; per-event dashd-only decomposition should show the DSTX class → 0; count any embedded-only (superset) events attributable to the omitted rate limit — expected 0. Acceptance = 24 h, ZERO `gbt-xcheck-txmerkle-mismatch` events, guard still armed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -205,7 +205,7 @@ jobs:
                     core_test sharechain_test share_test btc_share_test \
                     test_threading test_weights \
                     test_header_chain test_mempool test_template_builder \
-                    test_doge_chain test_compact_blocks test_dash_x11_kat test_dash_header_chain test_dash_block_replay test_dash_conformance test_dash_subsidy test_dash_mempool test_dash_simplifiedmns test_dash_quorum test_dash_quorum_root test_dash_mn_state test_dash_embedded_gbt test_dash_underfill_guard test_dash_smldiff test_dash_p2p_messages test_dash_p2p_connection test_dash_p2p_node test_dash_node_interface test_dash_config test_dash_broadcaster test_dash_broadcaster_full test_dash_share_hash_link test_dash_block_relay test_dash_rpc_request test_dash_rpc_conf test_dash_block_producer test_dash_embedded_relay_e2e test_dash_block_relay_plan test_dash_version_activation_latch test_dash_embedded_startup_invariant test_dash_block_relay_binding test_dash_block_relay_dual_arm test_dash_coinbase_parity test_dash_coinbase_muldiv test_dash_donation_combined test_dash_g3_assembled test_dash_work_target test_dash_work_job_targets test_dash_stratum_binding test_dash_job_notify_roundtrip test_dash_cb_payee test_dash_stratum_notify_roundtrip test_dash_poolnode_messages test_dash_peer test_dash_share_tracker test_dash_node test_dash_share_messages test_dash_stratum_extranonce_split test_dash_stratum_submit_reassembly test_dash_difficulty_parity test_dash_auto_ratchet test_dash_min_protocol_gate test_dash_serve_gate_journal test_dash_work_source test_dash_node_coin_state test_dash_bestcl_gate test_dash_coin_state_maintainer test_dash_node_embedded_wire test_dash_node_reception_wire test_dash_get_work test_dash_stratum_work_source test_dash_superblock test_dash_oracle_byte_parity test_dash_chainlock_verify test_dash_isdlock test_dash_dashd_cut_arm \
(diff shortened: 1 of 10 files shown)


frstrtr:
**vm905 build + KAT evidence (BLS=REAL)**

- Build: `cmake -DC2POOL_DASH_BLS=ON -DDASHBLS_ROOT=~/dashbls-prefix` on vm905, exit 0.
- `test_dash_dstx`: **8/8 PASSED** — registry gate, codec golden decode + byte-exact re-encode, `GetSignatureHash` pre-image vs independent byte assembly (also pins `pack(tx)`), `IsValidStructure` refusal matrix, USE red/green (plain zero-fee admission OUT on the min-feerate floor vs verified path ranked FIRST at +0.1 COIN modified with `total_fees` still base), fail-closed (no verifier / bad BLS sig via the real `verify_govvote_operator_sig` entry point), unpriceable-inputs admission at fee=0.
- Regression: `test_dash_mempool` 63/63, `test_dash_isdlock` 12/12 — the modified-fee refactor with delta==0 is byte-identical to base ordering (D1/D2/D3 selection suites unchanged-green).

Sibling: #1237 (W5-A fold fee pricing — the dominant 20/34 class). Measurement rides that PR's canary plan; AFTER-2 arms both flags, acceptance = 24 h ZERO `gbt-xcheck-txmerkle-mismatch` with guards armed.

PR #1251, opened 2026-08-16
dash(daemonless): seed the fast-start anchor's full header so the MN-payee bridge can authenticate its snapshot (#1250 cold-cut park)

frstrtr:
### The park (2026-08-16, vm905 `wa0655re8`, cut binary master `01dbba13` = #1243+#1250, `--coin-rpc` removed, `--embedded-null-arm`, **no dashd**)

After #1250 the header tip seeds instantly **at** the MN anchor (2513000), so the MN-set fold is *eligible* at t≈0 (the coincide gate is closed — see `test_dash_header_checkpoint_coincide`). Yet the embedded arm still **never served**:

```
[EMBED-STATUS] arm=would-decline cause=no-tip populated=0 have_mn=0
               bridge_wait=fold-mnlist-reply@h=2513000 hdr_tip=2517000
```

Over ~4.5 min: zero `populated>0` / `have_mn=1` / embedded-serve events. On a `--coin-rpc`-removed binary there is **no dashd fallback** → no template → rigs shed before the first serve (reserve 25→0 in ~14 min).

### Root cause — the anchor's `hashMerkleRoot` is null

The bridge requests a `base=ZERO` snapshot at the anchor. The peer answers; the demux correctly routes it to `on_historical_snapshot`; it is *consumed*. But DIP-4 authentication **step (b)** in `historical_sml.hpp` binds the snapshot's cbTx merkle proof to **our own PoW-verified header's `hashMerkleRoot`**:

```cpp
auto header_root = merkle_root_of_hash(diff.blockHash);
if (!header_root || header_root->IsNull())
    return fail("block header not held");
```

`main_dash.cpp`'s `set_merkle_root_at_fn` returns `get_header(anchor)->header.m_merkle_root`. The #1250 fast-start seed (`header_chain.hpp`) built the anchor `IndexEntry` setting **only** `hash/height/chain_work/prev_hash/status` — it never set `entry.header`, so `entry.header.m_merkle_root` is default (all-zero). Step (b) fails closed, `apply_fold()` never runs, the payee cursor never leaves the anchor, `have_mn` stays 0, `populated = have_tip AND have_mn` stays 0, and nothing is served.

Pre-#1250 this root was present because the header was **crawled in** from the older 2400000 pin; #1250 correctly collapsed that 31-min crawl but dropped the anchor header with it.

### The fix — pin the anchor's full header, self-verified

Carry the anchor block's full 80-byte header alongside its already release-pinned hash, and populate `entry.header` from it in the cold seed — **only after self-verifying `X11(header) == hash`**.

- Restores the pre-#1250 DIP-4 trust anchor while keeping #1250's instant start.
- **Mints no new trust.** The `hashMerkleRoot` is *derived* from the already release-pinned block hash: the seed refuses the header unless it X11-hashes to `hash`, so a corrupt pin **fails closed** (header left unseeded) rather than admitting a forged SML.
- **Reward-safety:** the served template's MN-payee / quorum-root bytes are unchanged — only the *latency* of acquiring `have_mn` on a cold daemonless start changes.

### Files
- `bitcoin_family/coin/chain_params.hpp` — `Checkpoint` gains optional full-header fields (`has_header`, `hdr_*`); other coins leave `has_header=false` and are byte-identical. Feature macro `C2POOL_FAST_START_CHECKPOINT_HAS_HEADER` lets the KAT compile on a pre-fix tree.
- `dash/coin/header_chain.hpp` — pin mainnet 2513000's header (version/prev/merkle/time/bits/nonce); seed populates + X11-verifies `entry.header`, fail-closed on mismatch.
- `test/test_dash_anchor_header_merkle_root.cpp` — regression (RED on master: null anchor root), the exact `set_merkle_root_at_fn` bridge callback, fail-closed guard, and X11 self-consistency. Folded into the allowlisted `test_dash_node_reception_wire` target.

### Seams reused
#999 checkpoint-lane recovery · #1162 `begin_fold` cursor · #1151 cursor restore · #1128 replay seam (payee cross-check + SML freshness) · #1250 coincident checkpoint (this is its direct successor). No new mechanism; the anchor header is the same release-pinned trust anchor the MN checkpoint already commits to.

### Verification
- **KAT red→green** on vm905 (BLS=REAL): pre-fix → anchor `merkleRoot` null (tests 1–2 fail); post-fix → real root, bridge callback returns it, corrupt-pin refused, X11 self-consistent.
- **vm905 daemonless cold-start measurement** (no dashd, `--coin-rpc` absent, `--embedded-null-arm`): time-to-first-embedded-template — see PR thread.

Draft — integrator review, not for merge.

### DIFF
--- a/src/impl/bitcoin_family/coin/chain_params.hpp
+++ b/src/impl/bitcoin_family/coin/chain_params.hpp
@@ -11,6 +11,11 @@
 #include <cstdint>
 #include <string>
 
+// Signals that ChainParams::Checkpoint carries the optional full-header fields
+// (has_header / hdr_*). A KAT can key its fail-closed self-verification case on
+// this so the same test source still compiles on a pre-fix tree.
+#define C2POOL_FAST_START_CHECKPOINT_HAS_HEADER 1
+
 namespace bitcoin_family
 {
 namespace coin
@@ -26,7 +31,29 @@ struct ChainParams
     uint256 genesis_hash;            // genesis block hash (SHA256d for identification)
 
     // Optional: checkpoint for fast-sync (skip syncing from genesis)
-    struct Checkpoint { uint32_t height{0}; uint256 hash; };
+    struct Checkpoint {
+        uint32_t height{0};
+        uint256  hash;
+        // Optional: the FULL 80-byte block header at `hash`, carried as its
+        // consensus fields. When present, HeaderChain seeds entry.header (not
+        // just entry.hash) so downstream consumers that need the block's
+        // hashMerkleRoot — notably DIP-4 historical-snapshot authentication in
+        // the DASH MN-set bridge — have their trust anchor at t~=0, WITHOUT
+        // crawling headers forward from an earlier checkpoint.
+        //
+        // Reward-safety: this mints NO new trust. The seed self-verifies
+        // pow(header) == `hash` (X11 for Dash) before adopting it, so the
+        // merkle root is DERIVED from the already release-pinned `hash`; a
+        // wrong pin fails closed (header left unseeded) instead of admitting a
+        // forged SML. has_header=false keeps every other coin unchanged.
+        bool     has_header{false};
+        uint32_t hdr_version{0};
+        uint256  hdr_prev_block;
+        uint256  hdr_merkle_root;
+        uint32_t hdr_time{0};
+        uint32_t hdr_bits{0};
+        uint32_t hdr_nonce{0};
+    };
     std::optional<Checkpoint> fast_start_checkpoint;
 
     // Halving
--- a/src/impl/dash/coin/header_chain.hpp
+++ b/src/impl/dash/coin/header_chain.hpp
@@ -116,6 +116,28 @@ inline DashChainParams make_dash_chain_params_mainnet() {
     p.fast_start_checkpoint->height = 2513000;
     p.fast_start_checkpoint->hash.SetHex(
         "000000000000002114d621d90f28b52c07746414491cbffc7cd373b3a56bc950");
+    // The FULL header of the anchor block (mainnet 2513000), so the cold seed
+    // populates entry.header — not just entry.hash. Without it the seeded
+    // IndexEntry carries a default (all-zero) hashMerkleRoot, and the DIP-4
+    // historical-snapshot authentication in the MN-set bridge fails closed at
+    // step (b) ("block header not held"): the base=ZERO anchor snapshot is
+    // consumed but never APPLIED, the payee cursor never leaves the anchor,
+    // have_mn stays 0, and the embedded arm never serves a template on a
+    // --coin-rpc-removed binary (2026-08-16 cold-cut park). Pre-#1250 this
+    // root arrived by crawling headers in from the older 2400000 pin; #1250's
+    // fast-start seed dropped it. These six fields are the consensus header of
+    // the SAME release-pinned block `hash` above; init() REFUSES them unless
+    // X11(header) == hash (self-verifying, fail closed) — so no new trust is
+    // minted. Machine-checked by test_dash_anchor_header_merkle_root.cpp.
+    p.fast_start_checkpoint->has_header = true;
+    p.fast_start_checkpoint->hdr_version = 536870912u;   // 0x20000000
+    p.fast_start_checkpoint->hdr_prev_block.SetHex(
+        "000000000000002affa937f36922946849675632312981c160133b94365408c1");
+    p.fast_start_checkpoint->hdr_merkle_root.SetHex(
+        "01470ce31e4ec9934a229f6eef4c9f561e42898be48d0bd646ab80f4bcc15b9c");
+    p.fast_start_checkpoint->hdr_time  = 1785337727u;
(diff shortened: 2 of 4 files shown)


frstrtr:
## KAT red→green proven on vm905 (BLS=REAL, C2POOL_DASH_BLS=ON)

Same test source (`test_dash_anchor_header_merkle_root.cpp`), folded into the allowlisted `test_dash_node_reception_wire` target.

**RED — master `01dbba13` (no fix), worktree at the exact cut base:**
```
[  FAILED  ] DashAnchorHeaderMerkleRoot.ColdSeedPopulatesVerifiedAnchorMerkleRoot
    e->header.m_merkle_root Which is: 0000…0000   (anchor merkleRoot NULL — DIP-4 step (b) 'block header not held')
[  FAILED  ] DashAnchorHeaderMerkleRoot.MerkleRootCallbackYieldsAnchorRootForBridge
    set_merkle_root_at_fn returns a NULL root for the anchor
 2 FAILED TESTS
```

**GREEN — this branch `5f2030f5` (fix):**
```
[ OK ] ColdSeedPopulatesVerifiedAnchorMerkleRoot   merkleRoot=01470ce3…bcc15b9c
[ OK ] MerkleRootCallbackYieldsAnchorRootForBridge (the exact main_dash.cpp callback returns the real root)
[ OK ] CorruptPinnedHeaderIsRefusedFailClosed      ([EMB-DASH] X11 MISMATCH — refusing to seed header (fail closed); header=null)
[ OK ] PinnedHeaderX11EqualsPinnedBlockHash        (X11(pinned header) == pinned block hash — trust-neutral)
[  PASSED  ] 4 tests.
```
The `PinnedHeaderX11EqualsPinnedBlockHash` pass is decisive: the pinned 80-byte header genuinely X11-hashes to the already release-pinned block hash `000000000000002114d621d90f28b52c07746414491cbffc7cd373b3a56bc950`, so the merkleRoot is derived from an already-approved constant — no new trust minted, and the fail-closed guard refuses any header that does not.

vm905 daemonless cold-start (no dashd, `--coin-rpc` absent, `--embedded-null-arm`) time-to-first-embedded-template measurement to follow in this thread.

frstrtr:
## vm905 daemonless cold-start measurement (no dashd, `--coin-rpc` absent, `--embedded-null-arm`, fresh data-dir) — HONEST RESULT

Binary: this branch built with `C2POOL_DASH_BLS=ON` (BLS=REAL, `BLS_LINKED` verified). Cold data-dir, mainnet coin-P2P peers only, **no dashd**.

**The fix is live and does exactly what it claims.** At t≈6s the seed self-verified and adopted the anchor header:
```
[EMB-DASH] Fast-start checkpoint header VERIFIED (X11==hash) merkleRoot=01470ce31e4ec993
[EMB-DASH] Fast-start checkpoint: height=2513000 hash=0000000000000021 header=seeded
```
At **t≈30s** the anchor base=ZERO snapshot **DIP-4-authenticated and folded** — the exact step that is structurally impossible on master (null merkle root → "block header not held" → consumed-but-never-applied):
```
[MN-CKPT] chain position VERIFIED: anchor h=2513000 matches our header chain
[MN-CKPT] PoSe FOLD at h=2513000 ... payee-eligible 2068 -> 2068 ...
          List was DIP-4 client-verified against the block's own coinbase commitment.
```
So the bridge **advances past the reported RED park** (`bridge_wait=fold-mnlist-reply@h=2513000`). `folds` goes 0 → 1.

**But first-embedded-template is NOT reached within the work-timeout — a *distinct, downstream* blocker gates it.** After the anchor fold, forward replay wedges:
```
[LANE-WATCHDOG] lane=mn-ckpt state=bridging cursor=2513001 target=2521000 frozen_for=90s
   waiting_for=block-bodies@h=2513001..2513064 applied=0 folds=1
   — the payee queue is NOT advancing; the embedded arm cannot arm until it does
[EMBED-STATUS] arm=would-decline cause=not-populated have_tip=1 have_mn=0
   bridge_wait=block-bodies@h=2513001..2513064 payee_cursor=0/2522547
```
Over 5 min `have_mn` stayed 0 because the forward payee replay never fetches the block bodies for h=2513001.. — the **#1225 / task-#138 "reseed-wedges-on-tip-body" family** (no getdata on the forward/reseed path), a KNOWN, separately-owned blocker, **not** the merkle-root blocker this PR fixes.

### Bottom line (measured, not claimed)
- **RED master** parks at the **auth wall**: `fold-mnlist-reply@h=2513000`, snapshot consumed-but-never-applied, `folds=0`, `have_mn=0`.
- **This fix** clears the auth wall: anchor snapshot DIP-4-verified + folded (`folds=1`) at t≈30s cold, daemonless. Proven deterministically by the red→green KAT.
- The bridge then meets a **second wall** (`block-bodies@h=2513001..`, #1225/#138) that also must clear before `have_mn`→1 and the arm serves.

**This PR is necessary but not by itself sufficient** for first-embedded-template. It is the correct, reward-safe fix for the investigated blocker (the null anchor merkle root) and moves the wall from the auth step to the forward-body-replay step. Reaching a served template additionally requires #1225/#138.

PR #1255, opened 2026-08-16
dash(daemonless): rotate the mn-checkpoint fold getmnlistd snapshot across the archival pool (dashd rotate-on-stall)

frstrtr:
## What

Rotate the mn-checkpoint anchor→tip fold's `getmnlistd` snapshot request across the archival peer pool, instead of pinning it to `m_primary` and re-asking the same peer on stall. This is the **last single-primary funnel** in the daemonless cold fold path.

Stacks on #1254 (block-body `CanServeBlocks` archival convergence) — base branch is `dash/dashd-cut-bulk-fold-canserve`, so the diff here is the single follow-up commit.

## Why (measured)

A live cold soak (vm905, cut binary, **dashd ABSENT**, cold data-dir) reached `State::Bridging`, then **froze ~26 min** behind ONE slow primary waiting for the deep base→anchor mnlistdiff while seven other handshaked archival peers (all `services=0x1c45`, NODE_NETWORK) sat idle:

```
LANE-WATCHDOG state=bridging frozen_for=95s..1056s waiting_for=fold-mnlist-reply@h=2513000 warn=1..9
MN-CKPT no reply to h=2513000 mnlist after 3 drives — re-asking (same block)
```

The block-body legs already fan out (#1253) and converge onto archival deliverers (#1254). The fold's `getmnlistd` snapshot leg was the one that still went to `m_primary` only, with `tick_pending_fold()` re-asking the *same* primary — so a slow-but-alive primary wedges the whole fold.

## dash-core parity

dashd selects a sync peer for a `mnlistdiff` and **rotates to another on stall** rather than re-hammering one peer. Ported as `send_getmnlistd_rotating()` → `next_stateful_peer()`: round-robin across the eligible (`CanServeBlocks` + handshaked + non-demoted) pool, preferring a carrier other than the last. `main_dash` wires the fold's `request_snapshot_fn` through it. Only one `getmnlistd` is ever outstanding and the reply is matched by block hash, so any carrier's reply satisfies the same await — rotating the carrier is safe.

Falls back to any eligible peer → any handshaked peer → `m_primary`, so it never wedges (mirrors `next_bulk_peer`). The live **tip-follower** `getmnlistd` stays pinned (unchanged); only the deep fold snapshot leg rotates.

## Reward safety

Fetch/connection-side only. This changes **which peer is asked**, never the served MN-payee / quorum-root bytes, which stay DIP-4 client-verified against the block's own cbTx commitment before they are believed.

## Tests

`DashStatefulFold.*` (3 new KATs, red/green in-file):
- `fold_getmnlistd_rotates_across_archival_peers_and_skips_pruned` — the pinned leg lands all 3 asks on the primary (RED contrast); the rotating leg fans out 1-1-1 across three archival peers and never touches the pruned (limited-only) peer.
- `fold_getmnlistd_reask_leaves_the_slow_primary_for_a_neighbour` — the re-ask rotates off the current carrier.
- `fold_getmnlistd_never_wedges_when_only_pruned_peers_exist` — fallback still sends, never drops.

Full `test_dash_p2p_node`: **139/139 pass**, no regressions (existing `stateful_request_legs_go_to_the_primary_only` still green — that path is unchanged).

Build: BLS=REAL.

_Draft — integrator review; not for merge until the vm905 cold-fold measurement lands._

### DIFF
--- a/src/c2pool/main_dash.cpp
+++ b/src/c2pool/main_dash.cpp
@@ -5960,7 +5960,10 @@ int run_node(bool testnet, const std::string& rpc_endpoint,
                 auto* cp = coin_p2p.get();
                 mn_ckpt_lane->set_request_snapshot_fn(
                     [cp](const uint256& block_hash) {
-                        cp->send_getmnlistd(uint256::ZERO, block_hash);
+                        // Rotate the fold snapshot's getmnlistd across the
+                        // eligible pool so a slow/non-serving primary cannot
+                        // wedge the anchor->tip fold (dashd rotate-on-stall).
+                        cp->send_getmnlistd_rotating(uint256::ZERO, block_hash);
                     });
                 // DEMUX (reward-critical): the reply comes back on the SAME
                 // mnlistdiff message the tip sync uses. Routed here it is
--- a/src/impl/dash/coin/p2p_client.hpp
+++ b/src/impl/dash/coin/p2p_client.hpp
@@ -668,6 +668,16 @@ private:
     // reconnect to the same non-serving address. Incremented on NOTFOUND for a
     // bulk body and on stall-eviction; cleared when that peer delivers a body.
     std::map<std::string, int> m_bulk_nonserver_strikes;
+    // Round-robin cursor for the STATEFUL request legs (the mn-checkpoint
+    // anchor->tip fold's getmnlistd snapshot). dashd picks a sync peer for a
+    // mnlistdiff and ROTATES to another on stall rather than re-asking one slow
+    // peer forever; a live cold soak froze 26 min pinned to a single slow
+    // primary waiting for the deep base->anchor fold snapshot. This cursor
+    // spreads the fold's getmnlistd (and its re-asks) across the eligible
+    // (CanServeBlocks) pool. m_last_stateful_peer lets a re-ask prefer a
+    // DIFFERENT carrier so it actually fans out.
+    std::size_t  m_stateful_rr{0};
+    std::string  m_last_stateful_peer;
     // The peer select_block_peer() last sent a block getdata to, so
     // request_block_tracked() arms the watchdog against the peer actually
     // asked (not an unrelated primary).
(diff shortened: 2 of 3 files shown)


frstrtr:
---

## vm905 cold-fold measurement (48cdc945, BLS=REAL, dashd ABSENT, fresh data-dir)

Launched 2026-08-16 14:52:13 UTC, `--coin-rpc` omitted, `--embedded-mainnet --embedded-utxo --coin-p2p-peers 8 --coin-p2p-discover`.

**The getmnlistd fix behaves exactly as designed — no single-primary freeze:**
- Anchor PoSe fold snapshot (h=2513000) landed in **~47s** (14:52:11 arm → 14:52:58 applied), vs the prior soak's **26-min** single-primary freeze on the same request.
- **0** `frozen_for` LANE-WATCHDOG warnings, **0** "re-asking (same block)" stalls over 10+ min (the prior soak logged warn 1..9 / `frozen_for=1056s`).
- Body window fans out across **21 distinct archival peers** (8 concurrent handshaked, refilled), all delivering bodies — the multi-peer parallel fetch (#1253/#1254 base) is fully engaged.
- Note: in this cold start the anchor snapshot was actually served by the on-handshake initial-sync `getmnlistd` the instant the first peer came up, racing ahead of the lane's re-drive (which correctly dropped at t=0.6s with no peer yet). The rotating re-drive is exercised at interval fold points (h≥2513500); the KATs prove that path directly.

**HONEST: the cold node did NOT reach tip / populated in the measurement window.**
- `reached_tip: false`, `populated_reached: false`, `first_embedded_serve: false`.
- Fold cursor 2513000 → **2513047** in 620s (anchor→tip window ≈ 9858 blocks to hdr_tip 2522858).
- The binding limiter is **fold-apply throughput** (~0.09–0.18 blk/s), NOT peer-fetch: bodies are buffered thousands ahead of the cursor (3461 received vs cursor +47). That is the ordered replay/UTXO fold-apply belt (#1252 / #146), a separate lane from the getmnlistd/peer-selection path this PR fixes.

**Conclusion:** this PR removes the getmnlistd single-primary freeze (KAT-proven + live no-freeze), which was a real cut-blocker, but it is **not sufficient on its own** to make the daemonless cold node fold anchor→tip within a usable window — the remaining wall is fold-apply throughput. That reframes the investigation's premise: the 26-min freeze was one stall, not the whole gap.

frstrtr:
Subsumed by #1257 (squash-merge 22:25Z): the getmnlistd rotate-on-timeout content (send_getmnlistd_rotating / next_stateful_peer) is on master via #1257's squash, verified by symbol presence in origin/master src/impl/dash/coin/p2p_client.hpp. Closing to avoid a redundant/conflicting re-apply. Not reverted — the feature is live on master.

PR #1256, opened 2026-08-16
DASH daemonless: head-of-line staller detection for the mn-checkpoint fold (dashd nStallingSince) — DRAFT, honest negative on throughput

frstrtr:
## DRAFT — integrator review, do NOT merge

Stacks on #1255 (`dash/dashd-cut-fold-getmnlistd-fanout`, HEAD 48cdc945). Base of this PR is that branch, not master.

## What this does

Ports dashd's front-of-download-window staller (`net_processing.cpp` `FindNextBlocksToDownload` `nStallingSince` + the `SendMessages` 2s adaptive disconnect-on-stall) into the mn-checkpoint anchor→tip fold's body fetch.

The fold is forward-CONTIGUOUS: `MnStateMachine::apply_block` cannot cross a hole, so the single body at the fold cursor (`m_next`) gates the whole ordered fold even while up to `kWindow-1` (64) later bodies have already arrived and buffered ahead. Before this change that front body was one of ~64 EQUAL tracked slots: its `getdata` rode a blind round-robin carrier and its slow carrier was disconnected only after `BODY_REREQUEST_MAX(4)` rotations.

New behaviour:
1. `MnCheckpointLane::service_tick` tags the block at the fold cursor as the single **head-of-line** body (`set_hol_request_fn` → `CoinP2PClient::request_head_of_line`).
2. In `service_pending_bodies`, a HOL slot on stall **re-homes to the fastest-delivering peer** — ranked by a new per-address delivered-body tally (the live analog of `replay_bulk_fetch.hpp`'s `PeerTally.blocks`), reusing the existing `CanServeBlocks`/`IsLimitedPeer` filter — instead of a blind round-robin pick.
3. Its carrier is disconnected (dashd `fDisconnect`) only once it is a **chronic** staller: the adaptive stall window has doubled to ≥16s (dashd's `nStallingSince` doubling, 2→64s). Early stalls race the front to a faster peer WITHOUT dropping a willing-but-slow peer, so the pool keeps its parallelism.

Reuses the existing `PendingBody` slot, disconnect-on-stall path, `CanServeBlocks`/demotion machinery. The only new state is one privileged-slot marker + the per-address delivered tally + the fastest-peer comparator.

**Reward-safe by construction**: fetch/connection-side only. No served MN-payee / quorum-root / subsidy bytes, no `publish()` hold touched, and the fold still validates every block (`payee_ok` stays in lockstep with `applied`, `diverged=0`).

## Tests

`test/test_dash_coin_p2p_client.cpp` — `DashCoinP2PHeadOfLine`, red/green as untagged-vs-tagged on the same `service_pending_bodies` pass:
- **RED** (untagged front, `request_block_tracked`): blind round-robin re-request, no fastest-peer targeting, no HOL disconnect.
- **GREEN core**: tagged front re-homes to the fastest deliverer on the first stall with the pool intact.
- **GREEN chronic**: a front stuck through the adaptive window has its chronic carrier disconnected.

Full `test_dash_p2p_node` target: **142/142 green** (BLS=REAL).

## HONEST measurement — the speedup premise did NOT hold on the live network

Cold daemonless soak on vm905 (dashd ABSENT; `--embedded-mainnet --embedded-null-arm --embedded-utxo --coin-p2p-discover`). Fold arms correctly: anchor 2513000, eligible=2068 MNs, `applied==payee_ok` in lockstep, `diverged=0`. HOL fires live (re-homes to fastest; chronic-disconnect guard holds).

Same-network, same-wall-window A/B against the running #1255 baseline (no HOL):

| build | Δapplied | Δwall | rate |
|---|---|---|---|
| baseline #1255 (no HOL) | 71 blk | 198 s | **0.359 blk/s** |
| this PR (HOL port) | 67 blk | 203 s | **0.330 blk/s** |

**No material speedup** (difference is inside the block-arrival noise; both fluctuate 0.0–0.7 blk/s instantaneously). Neither reached `populated` in the window (~6–8 h from tip either way).

Why the 10–40× target did not materialise: the investigation's premise was that the front block's carrier is *uniquely* slow while *other peers are fast* (so re-homing wins). On this live network the archival peers serve deep (~10k-deep) history at a **uniform** rate — bodies arrive at ~1.9/s aggregate spread evenly across all 8 peers (~0.24 bodies/s each). There is no dramatically-faster peer to re-home the front to, so re-homing between equally-slow peers gives no gain, and the fold rate stays bounded by the per-peer deep-serve rate (shared, too, with the `--embedded-utxo` bootstrap lane competing for the same body bandwidth).

An earlier revision that disconnected the carrier on the FIRST HOL stall (no adaptive backoff) measured **slower** than baseline (0.23 vs 0.36 blk/s) from pool thrash; the chronic-staller guard in this revision removes that regression and lands at parity.

**Verdict for review**: mechanism is correct, dashd-faithful, reward-safe, non-regressive, and well-tested — but it does NOT deliver the cold-start fold speedup, because the bottleneck on this network is uniform deep-history serving rate + cross-lane body contention, not head-of-line-on-a-uniquely-slow-carrier. Recommend NOT merging on throughput grounds; keep as the honest record and re-evaluate the real gate (per-peer deep-serve rate / prioritising the fold's front over the UTXO bootstrap's bulk window).

### DIFF
--- a/src/c2pool/main_dash.cpp
+++ b/src/c2pool/main_dash.cpp
@@ -5891,6 +5891,24 @@ int run_node(bool testnet, const std::string& rpc_endpoint,
                     if (!e) return std::nullopt;
                     return e->hash;
                 });
+            // dashd nStallingSince (net_processing.cpp FindNextBlocksToDownload +
+            // the SendMessages 2s adaptive disconnect-on-stall). Tag the
+            // FRONT-OF-ORDERED-WINDOW block (the fold cursor's next height) as the
+            // head-of-line body so the fetch layer disconnects its slow carrier on
+            // the FIRST stall and re-homes JUST that block to the fastest-
+            // delivering NODE_NETWORK peer — instead of leaving it as one of
+            // kWindow equal round-robin slots that competes with the buffered-ahead
+            // bodies for re-request attention. This lets the ordered fold cursor
+            // track buffer-arrival rate rather than the round-robin's luck.
+            // Reward-safe: getdata routing only — no served MN/quorum/subsidy bytes
+            // and no publish() hold is touched, and the fold still validates every
+            // block. Same header-by-height lookup the block-request seam uses.
+            mn_ckpt_lane->set_hol_request_fn(
+                [cp = coin_p2p.get(), hc = header_chain.get()](uint32_t h) {
+                    auto e = hc->get_header_by_height(h);
+                    if (!e) return false;
+                    return cp->request_head_of_line(e->hash);
+                });
             // SML validity attestation — the ONLY carrier of a post-anchor
             // PoSe ban. dashd applies those from consensus, never as a special
             // tx, so the bridge's block replay is structurally unable to see
--- a/src/c2pool/main_dash.cpp
+++ b/src/c2pool/main_dash.cpp
@@ -5960,7 +5960,10 @@ int run_node(bool testnet, const std::string& rpc_endpoint,
                 auto* cp = coin_p2p.get();
                 mn_ckpt_lane->set_request_snapshot_fn(
                     [cp](const uint256& block_hash) {
-                        cp->send_getmnlistd(uint256::ZERO, block_hash);
+                        // Rotate the fold snapshot's getmnlistd across the
+                        // eligible pool so a slow/non-serving primary cannot
+                        // wedge the anchor->tip fold (dashd rotate-on-stall).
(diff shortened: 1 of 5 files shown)


frstrtr:
Superseded on master by #1272 (cdbc6dde, BLOCK_STALLING_TIMEOUT parity) + this branch self-declared an honest-negative do-not-merge. Closing as stale; reopen if the HoL-staller angle is revisited.

PR #1262, opened 2026-08-17
dash: wire the W4 quorum-member resolver into the MN diff-store parallel fold engine (stacked on #1261)

frstrtr:
## What

Wire the already-ported W4 quorum-member resolver + W1 PoSePunish loop into the **parallel MN diff-store fold engine** on the cold MN-CKPT bridge (`main_dash.cpp`). The store-writer engine (`mn_bridge_fold_engine`) was stood up with **no** `set_members_fn`, so at the first punishing `qfcommit` it hit the fold's fail-closed:

> `block carries a non-null qfcommit (llmqType=5 ...) but no quorum-member resolver is installed — PoSe punishes cannot be folded, failing closed`

This PR mirrors the existing main replay-fold wiring onto the store engine: a second `QuorumReplayEngine` seeded at the same full-state anchor and a second `ReplayQuorumBridge` (whose ctor installs `set_members_fn`), with `observe()`/`after_fold()` driven around each `fold_block` from the store's block feed and `prime_at_anchor()` called after the deferred full-state seed lands. Stacked on #1261.

No new algorithm: the dashd `HandleQuorumCommitment` PoSe-punish loop (`replay_fold_engine.hpp`) and the rotated/non-rotated member derivation (`replay_quorum_engine.hpp`, dashd `CalcQuorumMembers`/`GetAllQuorumMembers` analog) were already ported, self-checked, and KAT-covered. This is the wiring.

## Reward-safety (by construction)
A wrong member set / wrong penalty → folded root diverges from the block's committed `merkleRootMNList` → `fold_block` **poisons** (HARD STOP) → the writer never sinks the row → store stays empty → callers fall through to the network path. Never a bad mint. Fail-closed stays the last resort.

## KAT — red / green / reward-safe (all pass, BLS=REAL)
- **RED**: `DashReplayQuorumSeam.Height2513130FailsClosedWithoutAResolver`, `DashReplayFoldReal.QfcommitWithoutMemberResolverFailsClosed`
- **GREEN**: `DashReplayQuorumSeam.Height2513130FoldsThroughDerivedMembership` (folds, punishes 4, reproduces `merkleRootMNList` byte-exact), `DashReplayFoldReal.PoSePunishAndCaseDBan2516412` (real ban-flip reproduces root)
- **REWARD-SAFE**: `DashReplayFoldSynthetic.RootMismatchIsAHardStopWithNamedHeight`, `DashMnDiffStore.FinalRootMismatchRefuses`
- 18/18 in the run above.

[...]

PR #1263, opened 2026-08-17
dash: getqrinfo rotated-quorum bootstrap across the anchor for the MN diff-store engine (stacked on #1262)

frstrtr:
## What

At cold MN diff-store arm, bootstrap the store-engine's second `QuorumReplayEngine`
across the anchor by fetching `getqrinfo(B0)` — where `B0` is the first rotated
(DIP-24, llmqType=5) cycle base after the anchor — DIP-4/merkle-authenticating the
returned quarter snapshots + work-block MN lists, and seeding them through the
existing bridge `seed_snapshots` / `seed_work_lists` guards so the cycle that
straddles the anchor becomes derivable.

Stacked on #1262 (`dash/mn-diff-store-quorum-resolver`). One file: `src/c2pool/main_dash.cpp`
(+1 include). Reuses the operator's store/engines verbatim; adds only the seed transport.

## Why (ROOT)

The store-engine's second `QuorumReplayEngine` is seeded at the anchor h=2513000 with
**0 pre-anchor quarter snapshots** (only pre-anchor header hashes). The first punishing
qfcommit at h=2513130 is a ROTATED llmqType=5 quorum whose cycle base is **2513088**,
whose quarter work blocks (2512792 / 2512504 / 2512216) are **all below the anchor**.
`self_contained_from(5, 2513000) = 2513952`, so the engine cannot self-derive cycle 2513088
from the post-anchor stream → `fold_qfcommit` gets `resolver has no member set` → the parallel
fold fails closed a window past the anchor → the store caps (~130 heights) → the MN-CKPT payee
lane falls to the capped network probe → the payee-desync fail-closed wedge (task #149 class).

`getqrinfo(2513088)` returns exactly the pre-anchor data the engine needs: the three
`quorumSnapshotAtHMinusC/2C/3C` (cycle bases 2512800/2512512/2512224) and the `mnListDiffAtHMinusC/2C/3C`
(work heights 2512792/2512504/2512216) — all below the anchor, which is precisely what the
bridge's `producible_from = anchor + 3C = 2513952` guard **accepts** and what the replay
must itself produce it **refuses**. One fetch bootstraps the whole straddle; 2513376/2513664
self-produce forward.

## Reward-safety by construction

Fetched snapshots are DIP-4/merkle-authenticated (`authenticate_historical_snapshot`, the shared
R3: applied-onto-empty SML root == cbTx.merkleRootMNList + blockHash tied to a PoW-verified header),
and every folded list is re-checked vs `merkleRootMNList` before the writer commits. A wrong/partial
snapshot → wrong members → folded root mismatch → writer refuses the row → store empty → callers use
the network path → **never a bad mint**.

## Ordering gate

`getqrinfo` is a P2P round-trip that races the fold advancing from the anchor. The block-applied
callback (which this PR owns) **holds** straddle blocks (≥ B0) until the seed lands, then **drains**
them in order; re-issues the fetch while holding; and — if the reply never arrives within a bounded
hold — caps the store below the straddle and lets callers fall through **rather than poison**. No
engine change.

## Verification

- **Build**: `c2pool-dash`, BLS=REAL (`C2POOL_DASH_BLS=ON`, dashbls static). Clean. Runtime banner
  `DASH BLS backend: REAL`.
- **Provenance**: `/proc/<pid>/exe` md5 `9ef2f5b8…` == built binary; `--version` =
  `archive/dash/tipfreeze-stale-tip-alarm-81-g7ac9d631` (this HEAD).
- **KAT** (`DashReplayQuorumSeam`, 13/13 PASS) encodes the red→green→reward-safe for the seed
  mechanism this PR feeds: `Height2513130FailsClosedWithoutAResolver` (RED),
  `Height2513130FoldsThroughDerivedMembership` (GREEN), `DerivesCycle2513088MembersMatchingDashd`
  (byte-parity vs dashd), `BridgeNeverArmsTheRootSelfCheckByItself` (reward-safe poison).

## Live cold daemonless measurement (vm905, dashd ABSENT, `--embedded-null-arm --embedded-utxo
`--coin-p2p-discover`, no `--coin-rpc`) — HONEST PARTIAL

Proven end-to-end **up to the DIP-4 gate**:
- Bootstrap ARMED with the exactly-correct `B0=2513088` (`getqrinfo(cycle_base h=2513088, type=5, C=288)`).
- `getqrinfo(2513088)` issued and re-issued; the reply is **received and correctly bound** by the new
  consumer (it authenticates the `AtHMinusC` diff at exactly `h=2512792 = B0−C−8`).
- Ordering gate holds the straddle blocks; **0 poison, 0 `resolver has no member set`, 0 rejects, 0 bad mint** — reward-safety fully preserved.

**Parked, reward-safe.** The seed does **not** land: the cold node's header chain starts at the
anchor and does **not** hold the PRE-ANCHOR work-block headers (2512792/2512504/2512216, all below
h=2513000), so `authenticate_historical_snapshot` fails DIP-4 leg (b) `block header not held` and
**fails closed** (3× observed). Store still caps below the straddle; `populated=0`; no embedded serve.

**Next residual** (names the real prerequisite, not a band-aid): the cold node must hold the ≥3-cycle
pre-anchor header span (≈2512216..2513000, ~784 headers) so the DIP-4 leg-(b) merkle trust anchor
resolves — e.g. a one-shot backward `getheaders` at cold store-arm before authenticating, or a
checkpoint-supplied pre-anchor header set. Relaxing leg (b) would rely only on the downstream fold
self-check and is deliberately **not** done (the HARD RULE requires the DIP-4 verify; dashd itself
authenticates qrinfo against its own full header chain).

DRAFT — do not merge.

### DIFF
--- a/src/c2pool/main_dash.cpp
+++ b/src/c2pool/main_dash.cpp
@@ -101,6 +101,8 @@
 #include <impl/dash/coin/replay_fold_consumer.hpp> // W5: bulk lane -> W1 DML fold + per-block root check
 #include <impl/dash/coin/replay_quorum_bridge.hpp> // SEAM: W4 quorum lane <-> W1 MembersFn
 #include <impl/dash/coin/replay_payee_publish.hpp> // SEAM: W1 fold -> the PAYEE queue that gates serving
+#include <impl/dash/coin/mn_diff_store.hpp>      // dashd evodb dmn_D4/dmn_S3 port: per-block MN list diffs + snapshots
+#include <impl/dash/coin/mn_gap_repair.hpp>      // payee-queue gap-repair seam (maintainer + checkpoint bridge)
 #include <impl/dash/node.hpp>          // dash::Node — sharechain pool-node (NodeBridge<NodeImpl,Legacy,Actual>)
 #include <impl/dash/config.hpp>        // dash::Config (PoolConfig/CoinConfig)
 #include <impl/dash/config_pool.hpp>   // dash::SharechainConfig — P2P_PORT / PREFIX / min-proto SSOT
@@ -3465,6 +3467,12 @@ int run_node(bool testnet, const std::string& rpc_endpoint,
     // Constructed only alongside the fold engine.
     std::unique_ptr<dash::coin::replay::FoldLiveTail>          replay_live_tail;
     std::unique_ptr<dash::coin::replay::ReplayPayeePublisher>  replay_payee_pub;
+    // MN DIFF/SNAPSHOT STORE (dashd evodb dmn_D4/dmn_S3 port,
+    // evo/deterministicmns.cpp:689-694): per-block list diffs + 576-cadence
+    // snapshots written by the fold's per-block commit, read by the payee-
+    // queue gap-repair seam. Constructed only alongside the fold engine.
+    std::unique_ptr<dash::coin::replay::MnDiffStore>           mn_diff_store;
+    std::unique_ptr<dash::coin::replay::MnDiffWriter>          mn_diff_writer;
     // PR-2 FORWARD: dashd's mined-commitment store, fed from the same replayed
     // bodies. shared_ptr because the qc-plan lambda (installed far above, on
     // the serve path) captures it by value and must see it appear later — it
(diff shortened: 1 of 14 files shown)


frstrtr:

PR #1266, opened 2026-08-17
dash: cold-arm pre-anchor header backfill so getqrinfo snapshots authenticate (stacked on #1263)

frstrtr:
## Cold-arm pre-anchor header backfill so the getqrinfo straddle seed authenticates

**DRAFT — stacked on #1263** (base branch `dash/mn-diff-store-getqrinfo-straddle-seed`). Do not merge; review only.

### The gap this closes
On a cold fast-start daemonless node the #1263 getqrinfo straddle seed is fully wired but never lands: the three rotated quarter snapshots for the straddle cycle 2513088 sit at WORK blocks **2512792 / 2512504 / 2512216**, all *below* the fast-start anchor (2513000), and the cold header chain holds only the lone anchor. DIP-4 leg (b) of `authenticate_historical_snapshot` binds the snapshot cbTx merkle proof to our PoW-verified header's `hashMerkleRoot`; with those pre-anchor headers absent it fails **"block header not held"** and fails closed, so the store caps below the straddle and the bridge falls to the network payee lane. The forward `add_header` path cannot supply them — a pre-anchor header's parent is not held, so it orphan-rejects.

### The fix (dashd-parity, option b)
Supply the span the way dashd holds it — from its own header tree. At cold store-arm, once the qrinfo reply is in hand, issue **one bounded BACKWARD `getheaders`** (locator = the h-4C work block from the `extraShare` leg, `hash_stop` = the trusted anchor) and install the returned span with a new `HeaderChain::install_preanchor_span` that walks **down** from the anchor: each parent is bound to its already-trusted child by an X11 prev-hash link **AND** independently `check_pow`-verified — exactly the two legs dashd requires of any header it stores. The tip / best-work are never moved (pre-anchor headers never compete for the tip; the synthetic-anchor daemon-tip honesty guards stay intact). `getqrinfo` is re-driven once the span lands so leg (b) authenticates on the next reply.

DIP-4 leg (b) stays **STRICT**: header must be held + merkle root must match. A wrong/forged/missing header produces no downward linkage from the trusted anchor → zero installs → leg (b) still fails closed → callers fall through to the network path → never a bad mint. Backfilled headers are themselves PoW-verified and prev-hash-linked to the release-pinned anchor. Cost is one-shot ~800 headers (~64 KB wire) per cold arm vs `--replay-bulk`'s ~2.5M, so fast-start is preserved. The reply reaches the installer via the `new_headers` event; the replay bulk headers-demux is null off `--replay-bulk`, so the daemonless posture routes it unchanged.

Design's flagged residual (getheaders starts at locator+1, so h-3C's own header would be excluded) is handled by seeding the locator from the **h-4C** `extraShare` leg one rotated cycle deeper.

### What is PROVEN (KAT + provenance-checked vm905 cold daemonless run)
- **KAT** `test_dash_preanchor_header_backfill` (5/5 green, folded into the allowlisted wire target). Demonstrates red→green *within each test*: before install `get_header`/leg (b) return nullopt ("block header not held"); after `install_preanchor_span` walks the span down from the trusted anchor, leg (b) returns each header's real merkleRoot. Fail-closed cases (PoW-broken header stops the walk; a batch that never reaches the anchor; an unheld anchor) install nothing.
- **Build**: DASH `C2POOL_DASH_BLS=ON` (real `libdashbls.a`), BLS backend REAL confirmed at runtime.
- **Provenance**: `/proc/<pid>/exe` md5 `f05c184fce4272b2b722e5c48498aa46` == built binary md5, `--version` = `...-82-g2692fb42` (this HEAD).
- **vm905 cold daemonless run** (fresh state dir, `--embedded-mainnet --embedded-null-arm --embedded-utxo --coin-p2p-discover`, **dashd ABSENT**, no `--coin-rpc`):
  - `getqrinfo issued` for straddle cycle base h=2513088 → `pre-anchor header BACKFILL issued: getheaders locator=h-4C stop=anchor` → `pre-anchor backfill INSTALLED 1071 header(s) down to h=2511929` (covers all three work blocks) → `getqrinfo rotated bootstrap SEEDED: 3 pre-anchor quarter snapshots + 3 work-lists ... cycle 2513088 is now derivable`.
  - **0** occurrences of "block header not held"; **0** "DIP-4 auth FAILED". Seed authenticates.
  - Seed authenticated ~2m18s after cold start.

### Honest status: end-to-end cold serve NOT yet reached — two DOWNSTREAM residuals (out of scope of this PR)
This PR does exactly what it claims (the seed now authenticates), which advances the store-engine frontier from "capped at/below the straddle" to the straddle cycle's own commitment. Two further, independent residuals block serve-ready and are **not** in the header-backfill's domain:

1. **Store-engine parallel fold caps at h=2513130.** The type-5 rotated quorum of cycle 2513088 is committed at 2513088+42 = **2513130** (its DKG mining window). The store-engine W4 resolver's `m_members_fn(5, quorumHash)` returns *no member set* for that quorum even though the seed made cycle 2513088 "derivable" (3 quarter snapshots + work-lists present). The resolver does not assemble/key the rotated quorum's member set for the seeded cycle. **This is the store-engine resolver's domain (#1262 W4 / #1263 seed→derive), one layer below the header backfill.**

2. **Main mn-ckpt bridge FAIL-CLOSED at h=2517703** (PAYEE DESYNC): projected MN `6b17d6e1…` not paid by the coinbase, and the height-exact DIP-4-verified SML attests the projected candidate VALID (counter-evidence, not a missed ban). Root: BAN-STATE PROBE cap exhausted (8/8) — 27 ProUpServTx applied without a pre-block-height MN list → queue-ordering divergence accumulates. **This is the known ban-state-probe / reinstatement-gap class (DEF5/#1043 territory), not caused by this change.**

`reached_tip=false`, `store_covers_past_qfcommit=false`, `first_embedded_serve=false`. The pre-anchor header backfill is proven at its own layer and is reward-safe; the next residual to attack is #1 (store-engine rotated-quorum member-set assembly for the seeded straddle cycle).

### DIFF
--- a/src/c2pool/main_dash.cpp
+++ b/src/c2pool/main_dash.cpp
@@ -6445,6 +6445,15 @@ int run_node(bool testnet, const std::string& rpc_endpoint,
                         uint32_t last_send_h{0};
                         uint256  cycle_base_hash;
                         std::vector<std::pair<uint32_t, dash::coin::BlockType>> hold;
+                        // Pre-anchor header backfill (cold fast-start): leg (b)
+                        // of the getqrinfo snapshot auth needs the FULL headers
+                        // of the three rotated work blocks, all BELOW the anchor
+                        // and absent from the lone-anchor cold chain. One-shot
+                        // backward getheaders installs the span DOWN from the
+                        // trusted anchor before the seed can authenticate.
+                        bool     backfill_sent{false};
+                        uint256  anchor_hash;      // trusted downward-link root
+                        uint32_t anchor_height{0};
                     };
                     auto sseed = std::make_shared<StoreStraddleSeed>();
                     {
@@ -6464,6 +6473,8 @@ int run_node(bool testnet, const std::string& rpc_endpoint,
                             // correct for both and costs at most a few dozen buffered
                             // blocks in the rare case the reply is slow.
                             sseed->first_need_h = sseed->cycle_base;
+                            sseed->anchor_hash  = ckpt.blockhash;
+                            sseed->anchor_height= anchor_h;
                             sseed->armed        = true;
                             break;   // one rotated type on mainnet (type 5)
                         }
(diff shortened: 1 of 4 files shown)


frstrtr:

PR #1267, opened 2026-08-17
dash: assemble+key the store-engine straddle cycle member set on getqrinfo seed

frstrtr:
## What

Makes the **W4 store-engine quorum resolver** compute and register the straddle
cycle's rotated quorum member set from the getqrinfo-seeded quarters, so
`fold_qfcommit`'s `m_members_fn(llmqType, quorumHash)` lookup resolves and the
parallel MN diff-store fold walks **through** the cycle's first punishing
qfcommit instead of capping there.

Stacks on #1266 (`dash/mn-diff-store-preanchor-header-backfill`).

## Why

With #1266, the pre-anchor header span authenticates the getqrinfo snapshots and
the 3 pre-anchor quarter snapshots + work-lists for the straddle cycle
(2513088, type 5, C=288) are **seeded** — the cycle is *derivable*. But the
store engine never drove `ComputeQuorumMembersByQuarterRotation` to a **keyed**
result for it: the member set was only ever attempted lazily at forward-observe,
and on the seeded straddle cycle that attempt did not populate a result the
qfcommit at `cycle_base + mining_window_start` (h=2513130) could consume. So the
fold hit:

```
[MN-DIFF-DB] ... quorum-member resolver has no member set for
llmqType=5 quorumHash=0000000000000025 — failing closed
```

capping the store ~130 heights past the anchor, and the main mn-ckpt bridge fell
to the capped network payee lane (the h=2517703 PAYEE DESYNC class).

## How

- `QuorumReplayEngine::derive_members_for_cycle(llmqType, cycle_base)` — drives
  the engine's own `compute_rotation_cycle` (the ported
  `ComputeQuorumMembersByQuarterRotation`, **verbatim**, no new consensus code)
  over the just-seeded quarters plus the self-derived new quarter, and registers
  the ordered member set under `m_members[{type, cycle_base + quorumIndex}]` —
  exactly the key the fold's `m_members_fn(type, quorumHash)` resolves a
  commitment quorumHash to (`GetQuorumMembers` derives the cycle base by
  `GetAncestor(nHeight − quorumIndex)`, so the commitment quorumHash's block sits
  at `cycle_base + quorumIndex`). Idempotent; returns a small result the caller
  logs.
- `ReplayQuorumBridge::derive_members_for_cycle` — pass-through that folds the
  outcome into the report counters.
- `main_dash.cpp` — invokes it the moment the getqrinfo straddle seed lands
  (right after `seed_snapshots` + `seed_work_lists` succeed), **before** the held
  blocks drain, so the qfcommit resolves and the fold folds through toward tip.

## Reward-safety (by construction)

Every folded list is re-checked against the block's committed `merkleRootMNList`
before the writer commits the row. A WRONG member set → root mismatch → the
writer refuses the row → the store poisons/empties → callers fall through to the
network path. **Never a bad mint.** The correct member set makes the folded
ban-state dashd-exact.

## Test — KAT-D (`test_dash_replay_quorum_engine`)

Real mainnet qrinfo fixture (`dash_mainnet_qrinfo_2516785.bin`), cycle 2516544,
32 quorumIndex slots, dashd's own `quorum info` member lists as goldens:

- **GREEN**: `derive_members_for_cycle` assembles **32/32** dashd-exact ordered
  member sets, each resolvable through `members_for(type, quorumHash)` by the
  per-index quorum base hash — the exact contract the fold consumes.
- **RED** (a missing quarter): derive returns `ok=false` with a **named** skip,
  and `members_for` is `nullopt` — the fold fails closed, never a guessed set.
- **REWARD-SAFE**: a tampered quarter still runs the math but the assembled set
  **differs** from dashd's — the divergence the writer's per-row
  `merkleRootMNList` self-check catches.

KAT-B (the 32-slot rotation math) and the full 20-test
`DashReplayQuorumEngine` suite stay green.

## Build / provenance

`c2pool-dash` built with **BLS=REAL** (dashbls linked). Draft — do not merge.

### DIFF
--- a/src/c2pool/main_dash.cpp
+++ b/src/c2pool/main_dash.cpp
@@ -6740,6 +6740,51 @@ int run_node(bool testnet, const std::string& rpc_endpoint,
                                             " fold_qfcommit past the anchor resolves its"
                                             " member set and the store extends toward"
                                             " tip.";
+                                // ── ASSEMBLE + KEY the straddle cycle NOW ────
+                                // The seed made the cycle DERIVABLE (snapshots +
+                                // work-lists present); this drives dashd's
+                                // ComputeQuorumMembersByQuarterRotation over them
+                                // (the engine's compute_rotation_cycle, verbatim)
+                                // and registers the ordered member set keyed by
+                                // (llmqType, cycle_base+quorumIndex) — the exact
+                                // key fold_qfcommit's m_members_fn(type,quorumHash)
+                                // resolves a commitment quorumHash to. Done BEFORE
+                                // draining the held blocks so the qfcommit at
+                                // cycle_base+mining_window_start (the height the
+                                // store used to cap at) resolves its member set
+                                // and the fold folds THROUGH toward tip.
+                                {
+                                    auto dr = qbr->derive_members_for_cycle(
+                                        sseed->llmq_type, sseed->cycle_base);
+                                    if (dr.ok)
+                                        LOG_INFO << "[MN-DIFF-DB] straddle cycle"
+                                                    " base h=" << sseed->cycle_base
+                                                 << " (type "
+                                                 << int(sseed->llmq_type)
+                                                 << ") rotated member sets"
+                                                    " ASSEMBLED + KEYED: "
+                                                 << dr.member_sets << "/"
+                                                 << dr.expected_sets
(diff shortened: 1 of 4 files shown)


frstrtr:

PR #1269, opened 2026-08-17
dash: seed W1 fold from EMPTY set at DIP3 activation (true from-chain self-derive)

frstrtr:
## Summary

Enables a **true from-DIP3 self-derive** so the fresh MN-checkpoint set is derived from chain (dashd `BuildNewListFromBlock` parity), not captured from dashd RPC.

The only missing piece was the anchor seed: the W1 fold must start from an **EMPTY** masternode set at the DIP3 activation height (`h = MAINNET_DIP3_HEIGHT = 1028160`). Empty is chain-true there — DASH's deterministic MN list is empty until the first ProRegTx at/after DIP3 enforcement — but the prestate loader (`replay_prestate.hpp`) rejected **every** empty entries set.

## Change (scoped, reward-safe, fail-closed)

`replay_prestate.hpp` — allow an empty entries set **only** when the prestate height equals the DIP3 activation height, reusing the **same** constant `--replay-bulk-start` defaults to (`rp::MAINNET_DIP3_HEIGHT`, no fresh literal). At any other height an empty set stays a hard reject (there it can only mean a half-parsed / records-dropped seed → silent divergence, the exact failure this loader exists to refuse).

Why this can never mint: the empty seed is not trusted on faith. The fold's per-block `merkleRootMNList` self-check (`replay_fold_engine.hpp` `poison()`) hard-stops at the first post-DIP3 ProRegTx block if the empty-at-DIP3 seed (or its anchor hash/height) is wrong — a wrong seed folds to the wrong root and poisons the engine instead of serving wrong bytes. DIP-4 / root self-checks stay strict; no unscoped relaxation.

## Test (red → green, proven)

`test_dash_replay_prestate_empty_dip3.cpp`, folded into the already-allowlisted `test_dash_mn_state` target so it is built + run in CI:

- empty prestate **at** DIP3 activation → **ACCEPTED**
- empty prestate **below / above / DIP3+1** → **REJECTED** (spot-exact, not a window)
- count/records mismatch still fails at DIP3 (other guards intact)
- the empty DIP3 seed reproduces the empty-list `merkleRootMNList` (`uint256::ZERO`) through the same `seed_engine_from_prestate` root self-check a live replay trusts

Verified on vm905 (BLS=REAL): with the fix all 6 GREEN; reverting only the guard scope to the unconditional pre-fix reject turns the two accept cases RED while the reject/scope cases stay green.

Stacked on #1268 (`dash/mn-checkpoint-dump-self-derived`). Draft — do not merge.

### DIFF
--- a/src/impl/dash/coin/replay_prestate.hpp
+++ b/src/impl/dash/coin/replay_prestate.hpp
@@ -31,6 +31,9 @@
 //   * every per-mn hash field is WIRE hex (raw internal bytes, memcpy'd).
 
 #include <impl/dash/coin/replay_fold_engine.hpp>
+#include <impl/dash/coin/replay_bulk_fetch.hpp>   // MAINNET_DIP3_HEIGHT — the
+                                                  // SAME constant --replay-bulk-start
+                                                  // defaults to (do NOT re-literal it)
 
 #include <cstdint>
 #include <cstring>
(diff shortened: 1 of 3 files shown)


frstrtr:

PR #1274, opened 2026-08-17
dash(replay-fold): gate payee cross-check on DIP3 ENFORCEMENT (fix h=1028163 self-derive divergence) [DRAFT]

frstrtr:
## Port (B) — payee-ordering fold-correctness (DIP3-genesis regime)

**Stacked on #1270** (`dash/uaf-coldbridge-replayfold-store-exclusion`). DRAFT, do not merge.

### Root cause (NOT the tie-break)
The full-DIP3 self-derive replay poisoned at **h=1028163** (3rd DIP3 block, 14 fresh MNs) with `DML FOLD PAYEE MISMATCH` even though `merkleRootMNList` **MATCHED** (computed==committed). The MN SET was right; the payee cross-check was wrong to run there.

I audited the tie-break the task fingered and it is **already dashd-exact**:
- `payee_before()` = `memcmp(a.data(), b.data(), 32) < 0` == dashd `uint256 operator<` (`base_blob::Compare`, uint256.h:55/59) — the storage-byte order, **not** the `base_uint` arithmetic order that bit Tier-2 (#1271).
- `compare_by_last_paid_height` / `nRegisteredHeight` / pass-5 bookkeeping all byte-match dashd `CompareByLastPaid_GetHeight` / `RebuildListFromBlock`.
- The SML-root match at the divergence height also proves the `proTxHash` byte layout is dashd-correct (it commits `proRegTxHash`).

The real bug: **DIP3 has two heights.** The deterministic list is *built* from ACTIVATION (1028160) but the coinbase paying `GetMNPayee(list)` is only *ENFORCED* from `DIP0003EnforcementHeight` (mainnet **1047200**). dashd's own consensus rule `CMNPaymentsProcessor::IsTransactionValid` (`masternode/payments.cpp:114`) early-returns valid **without** checking the det payee when `!DeploymentDIP0003Enforced(h)` (`deploymentstatus.h:53`). In `[activation, enforcement)` the historical coinbase pays the **legacy** masternode winner, not the deterministic projection — so the cross-check is stricter than dashd there and false-poisons a byte-correct fold.

### Fix (dashd-faithful, reward-safe)
Gate the payee cross-check on `DeploymentDIP0003Enforced(H)` (`H >= dip0003_enforcement_height`, default 1047200), mirroring dashd exactly.
- The `merkleRootMNList` **SET self-check stays UNCONDITIONAL**.
- The payee projection (pass 0) + `nLastPaidHeight` bookkeeping (pass 5) also stay unconditional — exactly as dashd `BuildNewListFromBlock` runs them from activation — so the payee axis is fully derived and self-consistent across the window and correct the instant enforcement begins.
- Production mines at tip (~2.5M), far above enforcement → **money-path check UNCHANGED**. No block is ever skipped; the fold now folds *through* the DIP3-genesis window instead of hard-stopping.

### KAT (red → green), BLS=REAL
`test_dash_replay_fold.cpp :: DashReplayFoldPreEnforcement` — same state + same payment-block (coinbase pays nobody):
- `AtEnforcementPayeeMismatchStillPoisons` — H≥enforcement → **poison** (strict where dashd enforces).
- `BelowEnforcementPayeeMismatchFoldsThrough` — H<enforcement → **folds through**, SET self-check still binds, `nLastPaidHeight` still bumped.
- `BelowEnforcementPayeePaidIsNotFlaggedSkipped` — control.

Verified **RED** (forcing the old unconditional check poisons at h=102 with the exact `merkleRootMNList self-check PASSED … PAYEE MISMATCH` message — the h=1028163 signature) **→ GREEN** (gated check folds through). Full `test_dash_mn_state` fold+prestate suites: **31/31 pass**, no regressions.

### DIFF
--- a/src/impl/dash/coin/replay_fold_engine.hpp
+++ b/src/impl/dash/coin/replay_fold_engine.hpp
@@ -352,6 +352,12 @@ struct ReplayMNState
 struct FoldGates
 {
     int32_t dip0003_height{1028160};   // fold start — no DML below this
+    // DIP3 has TWO distinct heights (dashd deploymentstatus.h:52 comment:
+    // "'active' and 'enforced' are different statuses for DIP0003"). The
+    // deterministic MN list is BUILT from activation (dip0003_height); the
+    // coinbase paying GetMNPayee(list) is only ENFORCED from here on. dashd
+    // mainnet chainparams.cpp:186 DIP0003EnforcementHeight = 1047200.
+    int32_t dip0003_enforcement_height{1047200};
     int32_t v19_height{1899072};       // basic BLS (per-entry nVersion wrapper)
     int32_t mn_rr_height{2128896};     // MN reward reallocation (Evo 4-in-a-row OFF)
     int32_t masternode_min_confirmations{15};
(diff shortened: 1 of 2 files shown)


frstrtr:

PR #1275, opened 2026-08-17
dash(replay): integrate bulk-throughput (A) + DIP3-enforcement payee gate (B) on #1270

frstrtr:
## Integration: bulk-throughput (A, #1273) + payee-enforcement gate (B, #1274)

Stacks BOTH dashd ports onto ONE branch on top of PR #1270
(`dash/uaf-coldbridge-replayfold-store-exclusion` @ `5dd71d1e`) so the full-DIP3
self-derive replay runs at dashd-parity throughput AND folds through the early-DIP3
payee-ordering regime. The two ports touch DISJOINT files — clean cherry-pick, no conflicts.

### (A) Bulk-fetch throughput (was #1273, commit `df10298d`)
Ports dashd `net_processing` into the `--replay-bulk` lane:
- **CanServeBlocks filter** — `pump()`/`eligible_peers` only assigns a height range to a
  peer advertising the service bits with an announced start_height covering it
  (NODE_NETWORK full-history; NODE_NETWORK_LIMITED within 288 of tip). Wires #148/#1254.
- **Stall-demote** (`BLOCK_STALLING_TIMEOUT`=2s) — head-of-line's assigned peer with a
  timeout-dominated tally is demoted and the range re-homed to a serving peer NOW, not
  after the flat 30s. No hard-disconnect of the coin-p2p peer.
- **Event-driven refill** — `on_block_body()` triggers an immediate `pump()` for the
  serving peer instead of waiting for the 1s Timer tick.

Reward-safe: fetch-side peer selection only — NEVER changes WHAT is folded. Per-block
`merkleRootMNList` self-check + poison fail-closed UNCHANGED; a per-height liveness bypass
guarantees no height is ever skipped.

### (B) Payee cross-check gated on DIP3 ENFORCEMENT (was #1274, commit `cb8a8c6e`)
The h=1028163 divergence was a FALSE poison: the coinbase payee cross-check ran in the
`[activation=1028160, enforcement=1047200)` window where dashd's own consensus rule
(`CMNPaymentsProcessor::IsTransactionValid`, `!DeploymentDIP0003Enforced(h)`) does NOT
enforce the deterministic MN payee — the coinbase pays the legacy winner, not
`GetMNPayee(list)`. The uint256 tie-break (`payee_before()` = memcmp over storage bytes)
was already dashd-exact (NOT the #1271-class arithmetic-order bug). Fix mirrors dashd's
enforcement gate (`dip0003_enforcement_height{1047200}`). SET self-check, payee projection,
[...]

PR #1273, opened 2026-08-17
dash(replay): bulk-fetch CanServeBlocks + BLOCK_STALLING_TIMEOUT demote + event-refill (throughput port, reward-safe) [DRAFT]

frstrtr:
**Stacked on #1270** (`dash/uaf-coldbridge-replayfold-store-exclusion`). DRAFT — do not merge.

## What
Ports dashd `net_processing`'s block-download peer policy into the full-history `--replay-bulk` lane (`replay_bulk_fetch.hpp` `BulkBlockScheduler::pump()` / `BulkFetchLane`). **Throughput-only, reward-safe.** This is open task **#148** (CanServeBlocks) plus the reactive stall-demote and event-refill.

## Why
The full-history run measured `timeout=753956 / 1.68M delivered` (45% re-request) with `notfound=0` across the whole 2.52M-block run: non-archival peers **silently drop** deep-history `getdata` rather than declining it, and nothing demoted a near-dead peer that kept its 32 slots the whole run. The in-order delivery cursor head-of-line-blocked behind each hole for the flat 30 s timeout — ~39 h of the 51.5 h wall were stall bands (≤2.3 blk/s) while healthy bands ran 75-107 blk/s.

## Three parts
1. **CanServeBlocks filter** (proactive) — `eligible_peers` seam → `CoinClient::eligible_bulk_peer_keys()` (handshaked ∩ archival ∩ non-demoted, reuses the #1254 helper); `pump()` gains a `peer_can_serve(peer,height)` predicate → `bulk_peer_can_serve()` (`CanServeBlocks` + `peer_covers_height` start_height coverage). Cites dashd `FindNextBlocksToDownload`/`CanServeBlocks`.
2. **Stall-demote** (reactive, `BLOCK_STALLING_TIMEOUT`) — `demote_after=3` consecutive unanswered getdata hold a peer off **fresh** ranges for `demote_cooldown_sec=60`; a delivered body clears it. Retries never gated; all-demoted bypass for liveness. `demote_after=0` = byte-identical pre-port behaviour.
3. **Event-driven refill** — a delivered body event-drives an immediate `issue_bulk_pump()` (same strict-priority guards, gated on the `now_sec` clock seam) instead of waiting up to one 1 s `core::Timer` tick — removes the per-tick issue ceiling.

## Reward-safety
[...]

PR #1276, opened 2026-08-17
dash(replay): CanServeBlocks + stall-demote for the header-backfill lane (extend #1273)

frstrtr:
## Extend #1273's CanServeBlocks + BLOCK_STALLING_TIMEOUT to the header-backfill lane

**Stacked on #1275** (`integrator/dashd-cut-bulk-and-payee`). DRAFT — review only, do not merge.

### The wedge (vm905, cold start, `--coin-rpc` removed, dashd absent)
The genesis→anchor header backfill selected its `getheaders` target by **blind round-robin** over `eligible_peers()` at `replay_bulk_fetch.hpp` `maybe_kick_backfill()` / `on_headers()`. That pool (`eligible_bulk_peer_keys`) has a liveness fallback that re-admits a pruned / `NODE_NETWORK_LIMITED` peer whenever the CanServeBlocks-filtered serving set is transiently empty — the exact cold-start case (one limited peer up). The limited peer **silently drops** the from-genesis `getheaders`, `m_backfill->tip_height()` never leaves 0, `BulkBlockScheduler` stays gated on `m_backfill->complete()`, **bodies=0, the fold never starts** (`hdr=0/2513000`).

This is the same class PR #1273 (`df10298d`) fixed for the **body** scheduler; the **header** lane was simply not covered.

### The port (faithful dashd `net_processing` parity)
The two halves of dashd's block-download peer policy, applied to the single-target header walk:

- **Proactive (CanServeBlocks):** reuse the `peer_can_serve` seam **verbatim** (main_dash → `bulk_peer_can_serve`: advertises full-block service AND `start_height` covers the span AND not bulk-demoted). CanServeBlocks is **not** reimplemented. The round-robin now only targets a peer that can serve `tip_height()+1`.
- **Reactive (BLOCK_STALLING_TIMEOUT):** a header peer that received `getheaders` but did not advance the walker within the re-kick window is **demoted** (`m_hdr_demoted_until`, cooldown parity with the scheduler's `demote_cooldown`) and the span **re-homes** to a different serving peer. This catches a peer that *advertises* service but lies/drops — the case CanServeBlocks alone cannot see. `on_headers` clears the demotion the instant a peer delivers (mirror of `on_body` clearing `demoted_until`).

**Liveness bypass (never skip a span):** if no peer advertises the height the filter degrades to a preference; if every serving peer is demoted the demotions are cleared and the walk re-homes — it never deadlocks. `backfill_demote_cooldown_sec = 0` restores byte-identical pre-port blind behaviour (like the scheduler's `demote_after=0`).

### Reward-safety
Header-fetch peer **selection only**. PoW + prev-hash header linkage stays strict (`HeaderBackfill::add_headers` unchanged); the anchor JOIN CHECK, the per-block `merkleRootMNList` / `merkleRootQuorums` fold self-check and poison fail-closed are untouched; **no header or block is ever skipped** — a demote re-homes the span, it does not drop it.

### Proof
KAT `DashReplayHeaderBackfillPeerSelect` in `test_dash_replay_bulk_fetch.cpp`, red→green:
- **red** — blind round-robin (peer_can_serve null, demote disabled) wedges on the non-serving peer, walker stalls short of the anchor, dead peer WAS targeted.
- **green** — CanServeBlocks filter joins the anchor without ever targeting the non-server; stall-demote re-homes off a lying full-node peer and completes; liveness bypass never deadlocks with all peers non-answering.

Built DASH `C2POOL_DASH_BLS=ON` (BLS=REAL). Full `test_dash_p2p_node`: 147/147 pass.

### DIFF
--- a/src/impl/dash/coin/replay_bulk_fetch.hpp
+++ b/src/impl/dash/coin/replay_bulk_fetch.hpp
@@ -1162,7 +1162,8 @@ public:
         uint32_t start_height{MAINNET_DIP3_HEIGHT};
         uint32_t tip_exclusion{12};          // live edge belongs to the tip lane
         int64_t  telemetry_interval_sec{30};
-        int64_t  backfill_rekick_sec{15};    // headers stall re-kick
+        int64_t  backfill_rekick_sec{15};    // headers stall re-kick (== dashd BLOCK_STALLING_TIMEOUT window for the header lane)
+        int64_t  backfill_demote_cooldown_sec{60};  // header peer held OFF targeting after a stall (#1273 demote_cooldown parity); 0 ⇒ demote disabled (byte-identical to the pre-port blind lane, like the scheduler's demote_after=0)
         uint32_t cursor_persist_every{512};  // delivered blocks between cursor writes
     };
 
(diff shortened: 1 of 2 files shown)


frstrtr:
Superseded on master by #1283 (b815fe10) + #1272 (cdbc6dde) — header-backfill CanServeBlocks + stall-demote landed there. Closing stale; reopen if a gap remains.

PR #1277, opened 2026-08-18
dash(replay): early-era (pre-V20) non-rotated quorum members — resolve the h=1081595 LLMQ_50_60 fold poison

frstrtr:
**Draft. Do not merge. Stacked on #1276 (`integrator/dashd-cut-header-backfill-canserve`).**

### The poison (live-proven)
The daemonless full-DIP3 self-derive fold (`dash-selfderive-hdrfix.log`) folded **1028161..1081594 byte-clean** (`folded=53434, roots_matched=53434, payees_verified=34395, DIVERGED=none`) — #1274 payee-DIP3 enforcement proven — then **failed CLOSED at h=1081595** on the FIRST non-null early `LLMQ_50_60` (llmqType=1) qfcommit:

```
fold FAILED at h=1081595 tx[1]: quorum-member resolver has no member set
for llmqType=1 quorumHash=0000000000000001... — failing closed
```

`quorum_lane_refusals=53435` (every block).

### Root cause (NOT commitment_is_null)
`commitment_is_null()` is already dashd `CFinalCommitment::IsNull`-faithful and unchanged. The h=1081595 commitment is genuinely **non-null** (41/50 signers, real quorumPublicKey / quorumVvecHash / sigs) — mainnet's first mined `LLMQ_50_60` DKG era (May 2019) with 9 invalid members that dashd's `HandleQuorumCommitment` PoSe-punishes. A member set is **required** for MN-list parity.

The gap is member **production below the V20 floor**. `QuorumReplayEngine::observe_block()` hard-refuses every block under `v20_floor` (mainnet 1'987'776) — the CL-modifier era, DIP-0024 rotation and the `merkleRootQuorums` self-check are all post-V20 / Phase-2 — so the W4 lane produced no member set the whole early-DIP3 run and the W1 fold's punish pass fell to the fail-closed site.

### Fix — faithful dashd pre-V20 `ComputeQuorumMembers` (verified vs dashpay/dash `develop`)
For a non-rotated DKG base block **B** (`B % dkgInterval == 0`) below the floor:
- **member list** = `GetListForBlock(B)` — the list AS OF the base block, **NO -8 work-block offset** (the offset lives only in `GetHashModifier`'s ChainLock lookup; `ComputeQuorumMembers` feeds the base-block list). Sourced as the W1 fold's just-proven DML at B (post_fold) = `GetListForBlock(B)` by construction.
[...]

PR #1278, opened 2026-08-18
dash(replay): progress-condition the bulk-lane defer so a frozen MN-ckpt bridge cannot wedge the self-derive at 0 blk/s

frstrtr:
## What

Progress-condition the W2 bulk-lane defer so a **frozen** MN-checkpoint bridge can no longer wedge the daemonless self-derive at **0 blk/s**.

The bulk genesis→anchor lane yields the shared coin-P2P response demux to the MN-checkpoint anchor→tip fold via `BulkFetchLane::Seams::defer_to_higher_priority`. That seam read `bridging_active()`, which is `true` for the **entire** `Bridging` state — including the multi-minute freezes the `LANE-WATCHDOG` measures when the cold bridge parks on a single-outstanding `getmnlistd` reply.

Wired straight to `bridging_active()`, one frozen fold became a **total priority inversion**: while the bridge sat `Bridging`-but-idle, **both** the genesis→anchor header backfill **and** the body pump issued nothing, so the whole self-derive held 0 blk/s.

## Why it is the dominant throttle (measured 2026-08-18, vm905)

Same host, same hour, same 7 archival peers, `-connect`-pinned:

| run | blk/s |
|---|---|
| dashd v22.1.3 IBD (26,514→175,658 in 1,568 s) | **95.1** |
| this-commit-base self-derive (vm905) | **0.0** — one 22-min wedge (hdr frozen 2,020,000) + one 26+-min wedge after a full process restart |

`LANE-WATCHDOG` bridge freezes on the wedged run: `frozen_for=217s / 337s` (`ondemand-mnlist-reply@2513860`), `98s+` (`fold-mnlist-reply@2515864`). `getmnlistd` round-trips on the `--replay-fold-quorums` **fold** path = **0** (0-RTT claim CONFIRMED against the same-commit reference run: 342,382 blocks folded, 299 total `getmnlistd` sends = 0.0009/block). The throttle is **not** a per-block round-trip and **not** raw CPU (3.5–5%) — it is a priority-inversion **freeze**.

dashd has no analogue: block download (`net_processing` `FindNextBlocksToDownload`) never waits on masternode-state sync; `CMasternodeSync` waits for blocks, never the reverse.

## The port (dashd `BLOCK_STALLING_TIMEOUT` parity)

New `MnCheckpointLane::bulk_should_yield()`. The defer yields the link to the bridge **only while the bridge's apply cursor advanced within `kBulkDeferStallMs` (20 s)**. Once the cursor stands still past that window the bridge is, by construction, idle-waiting on a reply and **not using the link** — the contention the defer guards against does not exist, so the bulk lane runs. The instant a reply lands and the cursor advances, `progress()` re-stamps the watchdog and the bridge reclaims priority in the same tick.

- Reuses the existing `BulkBlockScheduler` and the existing `StallWatchdog` (`m_watchdog` is already armed at the `Waiting→Bridging` edge and `progress()`-stamped on every cursor advance, so `elapsed_ms` is exactly "time since the bridge last moved"). **No new state, one predicate, one constant.**
- Seam signature unchanged — the lane owns its own clock via `m_now`.
- No band-aid, no cap, no skipped block.

## Reward-safety

Timing / throughput only. It reorders **who** fetches **when** on an otherwise-idle link, never **what** either lane folds. The bridge's forward-contiguous cursor, its per-block `merkleRootMNList` / `merkleRootQuorums` self-check and its poison fail-closed are **all untouched** — the same argument #1257 made for the reorder buffer.

## Proof

- **KAT red→green** — `DashMnCheckpointBridge.BulkDeferReleasesWhenBridgeFreezesAndReclaimsOnProgress` (folded into the allowlisted `test_dash_node_reception_wire` target). Drives a real 1519544..1519546 bridge with an injected clock: defer **holds** while progressing, **releases** once the cursor freezes past `kBulkDeferStallMs`, **reclaims** on the next advance. RED on the pre-port `bridging_active()` semantics (fails at the freeze assertion, `Actual true / Expected false`); GREEN after.
- **Full TU 174/174** green.
- **Build** — vm905, `C2POOL_DASH_BLS=ON`, BLS backend **REAL** (`dashbls` linked, RELIC), `c2pool-dash` links clean.

## Base / stacking

Stacked on `integrator/dashd-cut-early-quorum-memberset` (@ `d8ba8fba`, the #1277 cold-start fetch+fold stack) — the branch the throttle was measured on. Single commit, 3 files, +131/−1.

## Follow-up (out of scope here, named in the measurement)

The measurement also flags the coin-P2P **session hygiene**: c2pool demotes but never disconnect-cycles a starved coin-P2P session, where dashd disconnects a staller (`BLOCK_STALLING_TIMEOUT` → disconnect + redial) and recovers. That is a separate seam (the coin-P2P session pool, `POOL-STATUS` demote-only) and a separate, smaller PR — this one fixes the dominant 0-blk/s inversion.

---

Draft — do not merge. Review requested.

### DIFF
--- a/src/c2pool/main_dash.cpp
+++ b/src/c2pool/main_dash.cpp
@@ -7596,7 +7596,13 @@ int run_node(bool testnet, const std::string& rpc_endpoint,
             // false and the bulk lane resumes at full rate. Reward-safe: this
             // only reorders WHO fetches WHEN, never what either derives.
             seams.defer_to_higher_priority = [mnl = mn_ckpt_lane.get()] {
-                return mnl && mnl->bridging_active();
+                // PROGRESS-CONDITIONED (dashd BLOCK_STALLING_TIMEOUT parity):
+                // yield the shared link to the bridge only while its apply
+                // cursor is actually advancing. A bridge frozen on a pending
+                // getmnlistd reply is idle-waiting, not using the link -- so it
+                // must NOT starve the genesis->anchor backfill + body pump. See
+                // MnCheckpointLane::bulk_should_yield().
+                return mnl && mnl->bulk_should_yield();
             };
 
             rp::BulkFetchLane::Config rcfg;
--- a/src/impl/dash/coin/mn_checkpoint_lane.hpp
+++ b/src/impl/dash/coin/mn_checkpoint_lane.hpp
@@ -1417,6 +1417,53 @@ public:
     /// bulk path (a plain cut cold node, the KATs) nothing reads it, and it is
     /// pure telemetry-shaped state — reading it changes nothing.
     bool     bridging_active() const { return m_state == State::Bridging; }
+
+    /// Silence (no cursor advance) after which the bulk-lane defer treats the
+    /// bridge as FROZEN and stops yielding the shared coin-P2P link to it.
+    /// dashd BLOCK_STALLING_TIMEOUT is 2s and adapts up to 64s; the bridge
+    /// legitimately pauses a few seconds per on-demand getmnlistd fold, so this
+    /// sits well above a healthy fold pause (seconds) and well below a real
+    /// freeze (minutes) -- 20s cleanly separates the two, matching the
+    /// 2026-08-18 measurement's 15-30s parity band.
+    static constexpr int64_t kBulkDeferStallMs = 20000;
+
+    /// dashd BLOCK_STALLING_TIMEOUT parity, applied to the shared-demux defer.
+    ///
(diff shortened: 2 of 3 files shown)


frstrtr:
Superseded on master by #1272 (cdbc6dde) / #1254 (ba3e2f86) — the bulk-lane defer progress-condition landed there. Closing stale; reopen if a gap remains.

PR #1281, opened 2026-08-18
dash(replay): parallel getheaders window + event refill + stall-disconnect for the pre-anchor header backfill

frstrtr:
## Summary

Extends the #1263 pre-anchor header-backfill lane (task #151) to fetch headers the way dashd does: a **parallel getheaders window across serving peers**, **event-driven refill on each arrival**, and **stall-disconnect-and-redial** of a non-answering peer. Stacked on **#1278** (`integrator/dashd-cut-bulk-defer-progress` @ `ca7e9a5e`) — review/merge #1278 first.

**Draft** — not for merge. Reward-safe (header-fetch peer selection + timing only). DASH lane only.

## The diagnosis (wf `wx0agv6hk`)

Measured against a same-network **dashd v22.1.3 IBD = 95.1 blk/s** on the same 7 archival peers. The self-derive cold start did the full genesis→2513000 header backfill on a **single in-flight getheaders**:

- one getheaders outstanding at a time; on a peer that received it but never answered, the whole walk sat ~15 s until the re-kick timer demoted it and re-homed,
- advancing in lumpy bursts (0→8000 stall, burst to 180000 at ~1750 hdr/s, re-stall at 1394000),
- zombie sessions piled up (POOL-STATUS started/lost = 8/0, every socket at max RTO backoff 9.5 min).

dashd never serializes header fetch on one peer: `FindNextBlocksToDownload` keeps a parallel window across peers, refills event-driven on each message arrival, and on a staller **disconnects + replaces** (`BLOCK_STALLING_TIMEOUT`, "Peer is stalling block download, disconnecting").

## The port

1. **Parallel getheaders window.** Up to `Config::backfill_getheaders_window` (default 4) spans race concurrently across the CanServeBlocks-eligible peer pool — reuses the #1273/#1276 `peer_can_serve` seam and the `eligible_peers` filter verbatim. Racing is safe by construction: a stale racing batch whose prev-hash is still in `HeaderBackfill`'s recent window is *claimed* and then cleanly *rejected* (prev ≠ tip); the walk only ever extends its tip.
2. **Event-driven refill.** A delivered headers batch tops the window back up at the new tip from inside `on_headers` — no timer wait. The tick top-up is a cadence-gated backstop only.
3. **Stall disconnect + redial.** A peer that received a getheaders but sent nothing back within `backfill_stall_disconnect_sec` (default 15 s) is dropped and a replacement redialed (`CoinClient::stall_disconnect_and_redial` → `remove_peer` + immediate `refill_pool`) instead of demote-only. Fixes the zombie-session pileup. A silent peer is **not** credited by the walk advancing via *other* racing peers — the trap the single-inflight lane could not see.

**Liveness bypass preserved:** if every serving peer is demoted the selector clears demotions rather than deadlock; a disconnected peer's span re-homes; no header span is ever skipped. Null seams (KATs / any embedding without a live pool) fall back to single-inflight + demote-only, byte-identical to before.

## Reward-safety

Header-fetch **peer selection + timing only**. PoW + prev-hash header linkage stays strict; the anchor JOIN CHECK still pins the whole pre-anchor segment to genesis; the per-block `merkleRootMNList` / `merkleRootQuorums` fold self-check + poison fail-closed is **unchanged**; no header or block is skipped.

## KAT — red → green

New `dash_hdr_backfill_window_kat` drives the **real** `BulkFetchLane` header lane against a scripted peer set, RED (window=1, disconnect off) → GREEN (window≥2 + event refill + disconnect on). All 16 assertions pass:

- parallel window: 3 concurrent getheaders (GREEN) vs 1 (RED),
- event refill: next span issued from inside `on_headers` with no tick,
- stall disconnect+redial: the silent peer is dropped, the answering peer is kept,
- liveness: RED still completes via the demote-only fallback; GREEN never deadlocks; join pinned,
- serialization: **green = 4 ticks vs red = 19 ticks** — RED pays at least one full stall window, GREEN routes around the dead peer inside one.

## Build / provenance

`c2pool-dash` built on vm905, **BLS = REAL** (`DASH BLS backend: REAL (dashbls linked; quorum-commitment verify armed)`), binary md5 `dae6305dde828f69061e7cf6d631b832`, on HEAD `f701dd57` (base `ca7e9a5e`). KAT green. Live cold-start comparison against a same-host dashd IBD is the soak gate (not run here) — no floor is claimed without it.

### DIFF
--- a/src/c2pool/CMakeLists.txt
+++ b/src/c2pool/CMakeLists.txt
@@ -337,6 +337,30 @@ target_link_libraries(c2pool-dash
     ${Boost_LIBRARIES}
 )
 
+# KAT: pre-anchor header-backfill PARALLEL GETHEADERS WINDOW (extends #1263 /
+# task #151; base PR #1278). Drives the real BulkFetchLane header lane against a
+# scripted peer set: RED (window=1, no disconnect) serializes/wedges on a silent
+# peer; GREEN (window=N + event refill + disconnect-and-redial) completes. Links
+# the same DASH library set as c2pool-dash (header lane pulls block_producer /
+# header_chain / x11 / block-replay; BLS carried for the shared object graph).
+add_executable(dash_hdr_backfill_window_kat dash_hdr_backfill_window_kat.cpp)
+target_link_libraries(dash_hdr_backfill_window_kat
+    core pool sharechain
+    c2pool_node_enhanced
+    dash
+    c2pool_payout c2pool_merged_mining c2pool_hashrate c2pool_storage c2pool_difficulty
+    dash_x11
+    dash_rpc
+    dash_stratum
+    dash_block_replay
+    dash_bls_verify
+    btclibs
+    yaml-cpp::yaml-cpp
+    nlohmann_json::nlohmann_json
+    ${Boost_LIBRARIES}
+)
+add_test(NAME dash_hdr_backfill_window_kat COMMAND dash_hdr_backfill_window_kat)
+
 # Windows: Boost.Asio requires Winsock libraries; /bigobj for large translation units
 if(WIN32)
     target_link_libraries(c2pool ws2_32 mswsock bcrypt dbghelp)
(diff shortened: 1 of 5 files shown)


frstrtr:
Superseded: the parallel getheaders header-backfill fast-IBD content landed on master via #1283 (b815fe10). Closing as redundant.

PR #1288, opened 2026-08-19
dash(replay): CPUID-dispatched SIMD/SHA-NI SHA256 for the daemonless fold self-check (dashd-cut speedup)

frstrtr:
## What

Enable Bitcoin Core's **CPUID-dispatched SIMD/SHA-NI SHA256** for the DASH binary, so the daemonless replay fold's per-block `merkleRootMNList` / `merkleRootQuorums` self-check stops running the scalar C transform.

**Base:** stacked on `integrator/dashd-cut-combined-fetch-fold` (the profiled combined fetch+fold binary). **Draft — do not merge.**

## Why (profile)

Fold-thread perf (10541 samples, mns=4943): the reward-safety self-check is **83.5%** of fold CPU, and **69.6%** of total self-time sat in the vendored **scalar `sha256::Transform`** — the fold SHA256's every one of ~4900 SML leaves plus the whole merkle tree, single-thread CPU bound, every block.

Root cause: `btclibs` vendors the full CPUID-dispatched SHA256 family (sse4 single-block, sse4.1 4-way + avx2 8-way `TransformD64`, SHA-NI), but every optimized TU is wrapped in `#ifdef ENABLE_*/USE_ASM` and those macros were **never defined**, and `SHA256AutoDetect()` was **never called** — so the whole binary ran scalar for its entire life.

## Fix (two byte-exact parts)

1. `src/btclibs/CMakeLists.txt`: define `USE_ASM/ENABLE_SSE41/ENABLE_AVX2/ENABLE_SHANI` and give each optimized TU its `-march` flag, so the SIMD code compiles.
2. `src/c2pool/main_dash.cpp`: call `SHA256AutoDetect()` once at startup; it probes CPUID and installs the fastest transform the running CPU proves correct, then an explicit **NDEBUG-proof KAT** aborts startup rather than fold with a miscompiled primitive.

## Reward-safety

The fold logic and the `merkleRoot` self-check are **UNCHANGED** — only the SHA256 primitive underneath is swapped, and only to a path that passes `CSHA256::SelfTest` at runtime. A wrong primitive can only make a root **mismatch and hard-stop**; it can never silently mis-fold. A CPU lacking a feature never selects that path, so the binary stays portable (Broadwell → sse4; SHA-NI hosts → the big win).

## Tests

New KAT `test_dash_sml_simd_sha256` (over a realistic 4943-entry MN list):
- SIMD-dispatched `merkleRootMNList` **== pinned golden** AND **== the scalar root** computed before `AutoDetect` ran (byte-identical self-check input).
- a one-entry `isValid` flip **still changes the root** (poison intact).

Existing byte-parity KATs stay green against the SIMD-enabled `btclibs`:
`test_dash_simplifiedmns` (10), `test_dash_quorum_root` (11), `test_dash_conformance` (56).

Startup on the profiling host:
```
[init] SHA256 backend: sse4(1way),sse41(4way),avx2(8way)
[init] SHA256 SIMD KAT: OK (byte-exact vs scalar goldens)
[init] DASH BLS backend: REAL (dashbls linked; quorum-commitment verify armed)
```

## Measured

Per-block self-check recompute on vm905 (Xeon E5-2690 v4, Broadwell, **no SHA-NI**): scalar **13.4 ms → simd 9.5 ms (1.41x)** via sse4 single-block. This is the honest floor for a pre-SHA-NI AVX2 part — the leaf/merkle hot path is single-block SHA256, which only the SHA-NI transform accelerates 3-6x, so **SHA-NI production hardware gets the full 3-6x**. Byte-exactness is identical on all hosts.

## Follow-up (not in this PR)

The hardware-independent 10x is an **incremental SML-root cache** — cache each entry's `CalcHash` keyed by proTxHash and recompute only the <10 leaves that actually mutate per block (dashd's own posture; the self-check would still police a wrong invalidation). Higher yield, higher risk — needs its own soak; deliberately kept separate from this byte-exact primitive swap.

### DIFF
--- a/src/btclibs/CMakeLists.txt
+++ b/src/btclibs/CMakeLists.txt
@@ -43,6 +43,33 @@ set(sources
         )
 
 add_library(btclibs STATIC ${sources}) #${SOURCE} ${HEADER})
+
+# ---------------------------------------------------------------------------
+# Runtime-dispatched SIMD SHA256 (Bitcoin Core crypto/sha256.cpp).
+#
+# btclibs vendors the full CPUID-dispatched SHA256 family (sse4 single-block,
+# sse4.1 4-way + avx2 8-way TransformD64, and the SHA-NI transform), but every
+# optimized translation unit is wrapped in ifdef ENABLE_*/USE_ASM and stays
+# EMPTY unless the macro is defined, and SHA256AutoDetect was never called at
+# all -- so the whole binary ran the scalar C sha256::Transform. Define the
+# gates and give each optimized TU its -march flag so the code compiles;
+# SHA256AutoDetect() (called once at process start) then probes CPUID and
+# installs the fastest transform the running CPU proves correct via
+# CSHA256::SelfTest. Output is BYTE-IDENTICAL to the scalar path by
+# construction; a CPU lacking a feature never selects that path, so the binary
+# stays portable across the fleet.
+#
+# Reward-safety: the DASH daemonless replay fold recomputes the full
+# merkleRootMNList / merkleRootQuorums every block and byte-compares them to the
+# committed cbTx roots (poison fail-closed on mismatch). That self-check is
+# UNCHANGED and stays the authority -- a miscompiled SIMD transform can only
+# make a root mismatch and hard-stop, never silently mis-fold. This swaps only
+# the primitive underneath ~70% of fold-thread CPU (its scalar form).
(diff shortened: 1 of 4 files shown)


frstrtr:
Superseded — CPUID-dispatched SIMD/SHA-NI SHA256 for the daemonless fold landed on master via #1289 (merged 2026-08-20, standalone). Closing this earlier draft. Reopen if it carries unique content not in #1289.

PR #1293, opened 2026-08-20
dash(daemonless): combined fetch+fold onion — THE dashd-cut landing branch

frstrtr:
THE dashd-cut onion landing vehicle. This is a `git merge` of master into the combined fold-tip (conflicts pre-resolved, KATs green), NOT the granular #1263-#1287 drafts — those sit on a stale base (#1257) and carry pre-merge originals of #1259/#1260/#1261, so rebasing them onto master is a deadbase-revert of evolved reward-path fold code. This branch already carries the evolved fold engine.

Contents (all reward-path — operator review + merge tap, do NOT auto-merge): the self-derive fold engine (straddle/backfill/empty-DIP3-seed/checkpoint-dump/UAF fixes), DIP3-payee tiebreak, operator-reward split mirror, pre-v19 platform-quorum type-4 double-punish gate, bulk-throughput CanServeBlocks, tx-selection IS/CL-hold + DSTX exclusion-discipline, and distributed won-block broadcast.

+20 / -2 vs master. Supersedes the granular drafts #1262/#1263/#1266-#1270/#1274/#1275/#1277/#1284/#1287 (their content is here, evolved). #1290 (incremental SML root) and #1292 (anti-wedge re-ask) fold on top of this. #1289 (SIMD) lands to master independently first, then rebuild this on master+#1289 so SIMD isn't duplicated.

Draft: opened now so CI validates the full onion payload in parallel with the cold-start anchor derivation (task #154). Merge sequence at cut-time: #1289 → #1229 → THIS → #1290 → #1292.

PR #1292, opened 2026-08-20
dash(replay): wall-clock watchdog re-asks ondemand getmnlistd (anti-wedge)

frstrtr:
Anti-wedge fix at the tip of the dashd-cut onion stack (stacked on the combined base 8480d28e).

Problem: the #1286 getmnlistd rotate-on-timeout is wired to the ondemand/bridging lane, but its only trigger is pump(), driven solely by HeaderChain::on_tip_changed. With headers already at tip (no new header events), the timeout->advance never fires, so the node freezes permanently on one unanswered ondemand-getmnlistd.

Fix: a wall-clock watchdog_tick() every 45s (= dashd EXTRA_PEER_CHECK_INTERVAL) re-asks a DIFFERENT peer and raises outbound_behind so the pool dials past 8.

Reward-safe: 3 src + 2 test files, ZERO apply-path moved. mnlistdiff-apply / merkleRoot self-check / payee cross-check / poison all untouched. KAT green-on-fix / RED-on-base; adversarial verify PASS.

Draft: merges after the onion stack, as part of the cold-start speedup (task #154). CI here runs on the full combined anchor tree.

PR #1294, opened 2026-08-20
dash(daemonless): emit the self-derived MN checkpoint from the MN-CKPT bridge (task #154)

frstrtr:
## What

Give the **MN-CKPT bridge** the one-shot `--dump-mn-checkpoint` seam that until now
lived only on the empty-DIP3 **replay-fold** consumer. This lets a fresh anchor be
**bumped** from the baked checkpoint via a short forward fold (bridge SEEDs the baked
2513000 checkpoint and forward-applies to `H`), instead of a from-DIP3 replay (~a day).

Reward-safe: **no change to the fold / apply / consensus path** — only a dump hook and
a serialization overload.

## The change (3 parts, each mirroring existing code)

1. **`mn_checkpoint_dump.hpp`** — factor the reward-safety-critical payload/`.inc`
   assembly out of `emit_mn_checkpoint_dump()` into `finish_mn_checkpoint_dump()`
   (byte format **unchanged**, now single-sourced so the two dumpers cannot diverge),
   and add a **bridge overload** of `emit_mn_checkpoint_dump()` / `write_mn_checkpoint_inc()`
   that serializes `MnStateMachine::entries()` (`std::map<uint256, MNState>`) via a pure
   `MNState -> ReplayMNState` field-copy. Reuses `emit_mn_record_line()` and the
   **self-verify-through-`parse_mn_checkpoint()`** guard VERBATIM — a set that fails its
   own self-parse is never written.
2. **`mn_checkpoint_lane.hpp`** — add `set_dump_hook(target_height, fn)` + a one-shot
   `maybe_fire_dump()` fired from the **same clean-fold event** that feeds
   `set_on_block_applied` (both apply sites). It therefore only fires on a
   forward-contiguous, non-gapped, non-payee-desynced, non-empty fold. Exact analog of
   `FoldReplayConsumer::set_dump_hook`. Unset = no-op.
3. **`main_dash.cpp`** — in the cold MN-CKPT bridge arm (the `g_replay_fold_prestate.empty()`
   branch), install the hook when `--dump-mn-checkpoint` is set, mirroring the replay-fold
   install; mutually exclusive by the prestate predicate. Block hash comes from OUR
   PoW-validated header chain.

`getmnlistdiff` is payout-stripped and **cannot** back a checkpoint — the bridge's
`apply_block` (dashd `BuildNewListFromBlock` parity) is the source, so every field is
self-derived from chain, no dashd.

## Test evidence

- **Build**: `c2pool-dash`, DASH + real-BLS recipe (`C2POOL_DASH_BLS=ON`), Release — clean.
- **Run** (`--run --embedded-mainnet --coin-p2p-discover --coin-p2p-peers 16 --web-port 0
  --dump-mn-checkpoint 2522504 <out>`, **no** `--replay-bulk`, **no** `--replay-fold-prestate`):
  - `--dump-mn-checkpoint ARMED on the MN-CKPT BRIDGE` — the new install fires.
  - Bridge SEEDs baked-2513000 (`entries=2974`) and folds forward with
    **`diverged=0`** and `payee_ok == applied` throughout (reward-safe).
- **Oracle-verify** against fleet mainnet dashd (`protx list registered 1 2522504`):
  _pending the dump landing_ (see note).

### Note on pacing

On the build host the fold to `H=2522504` is **pace-limited by coin-P2P peer
contention** from two co-resident full-history self-derive runs saturating the same
peer pool: each periodic getmnlistd fold-point stalls until the next mainnet tip
re-drives it (this base predates the on-demand-mnlist re-ask, #1292). It advances
reward-safely and lands the anchor, but in hours rather than minutes on a contended
host — an environment constraint, not a defect in this change. On an uncontended host
(or the combined onion base with the reduced-mismatch payee logic) it bumps in minutes.

## Base / landing

Stacked on `integrator/dashd-cut-header-backfill-canserve` so the diff is exactly this
one commit. Intended to ride into the combined onion (**#1293**), which carries the same
dump feature + bridge and where the fold-consumer `set_dump_hook` install already lives.

**DRAFT — reward-path, operator review. Do not merge / undraft.**

### DIFF
--- a/src/c2pool/main_dash.cpp
+++ b/src/c2pool/main_dash.cpp
@@ -6180,7 +6180,23 @@ int run_node(bool testnet, const std::string& rpc_endpoint,
             // matches, the store COVERS heights and the ban-state probe /
             // on-demand fold source 0-RTT/uncapped from it.
             if (mn_ckpt_lane && header_chain && ckpt.ok
-                && !mn_diff_store) {
+                && !mn_diff_store
+                // UAF FIX (yield to the deep replay-fold): the cold MN-CKPT
+                // store-bridge and the W5 replay-fold arm (main_dash.cpp ~7288)
+                // both drive the SINGLE mn_diff_store/mn_diff_writer, but they
+                // are MUTUALLY EXCLUSIVE by design -- this block's own header
+                // comment names it "the COLD MN-CKPT bridge path (no
+                // --replay-bulk / --replay-fold-prestate)". The guard never
+                // actually enforced that: this arm runs FIRST (cold bridge),
+                // creates store/writer A and installs set_on_anchor_snapshot()
+                // whose lambda captures raw w=mn_diff_writer.get(); the later W5
+                // make_unique then FREES writer A behind that lambda's back, and
+                // the historical-snapshot callback dereferences the freed writer
+                // -> MnDiffWriter::arm()->save_snapshot() UAF (SIGSEGV/SIGBUS).
+                // When --replay-fold-prestate is set the deep replay owns the
+                // store; the cold bridge YIELDS. Normal cold start (no prestate)
+                // is UNCHANGED -- this predicate is true there.
+                && g_replay_fold_prestate.empty()) {
                 namespace rp = dash::coin::replay;
                 mn_diff_store = std::make_unique<rp::MnDiffStore>(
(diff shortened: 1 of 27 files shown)


frstrtr:
Superseded by #1371 (rebased onto master; 9/17 commits survive).

PR #1303, opened 2026-08-21
dash: MNHF/EHF-signal auto-prioritise for daemonless tx-serving parity (flag-OFF, reward-safe)

frstrtr:
## What

Ports dashd's **one automatic `mapDeltas` write** that every node performs at mempool admission and that the embedded selector did **not** reproduce: `CTxMemPool::addUnchecked` prioritises every **type-7 `TRANSACTION_MNHF_SIGNAL`** (EHF-signal) tx by **+0.1 COIN** — dashd `src/txmempool.cpp:701-702` (v23.1.7):

```cpp
if (tx.nType == TRANSACTION_MNHF_SIGNAL)
    PrioritiseTransaction(entry.GetTx().GetHash(), 0.1 * COIN);
```

These signal txs are ~0-fee on the wire but must sort to the very top of the block template. Without the delta the embedded ancestor-score selector scored them at their base (~0) feerate, so a higher-fee ordinary tx preceded them and the served **tx-set + tx-merkle root diverged** from dashd's `getblocktemplate`. This is the "one automatic delta not yet reproduced at admission" named in the Phase-2 tx-serving-parity diagnosis (Rank-5 residual).

## How

The delta is recorded into the existing bounded `mapDeltas` store (`record_fee_delta_locked`) so it is picked up by `add_tx_locked`'s `fee_delta` read and survives / retro-applies **identically to the DSTX delta**. Unlike the DSTX delta (network-DSTX-only, `add_dstx`), this one is intrinsic to the tx *type*, so it is applied on **every** admission path (plain `MSG_TX` included) — exactly as `addUnchecked` does.

## Reward-safe / fail-closed

- The delta only raises `modified_fee()` — the scoring / ordering / `blockMinFeeRate` basis (dashd `nModFeesWithAncestors`).
- The **raw base fee** that feeds `total_fees -> coinbasevalue` and `SelectedTx.fee` is **untouched**, so the coinbase can never be overstated by the delta.
- A type-7 tx whose input the UTXO fold cannot price stays `fee_fold_proven == false` -> template-excluded (withheld).
- **Dormant unless `--embedded-serve-mempool-txs` is armed (default-OFF).**

## KAT (allowlisted target, not a standalone CTest sentinel)

Folded into `test_dash_mempool` (allowlisted in `build.yml`):

- `MnhfSignalAutoPrioritiseSortsToTopButFeeStaysBase` — a low-base-fee type-7 **outranks** a higher-fee ordinary tx in the template (**RED on master**), while `SelectedTx.fee` and `total_fees` stay **base-only** (reward-safety anchor).
- `MnhfSignalUnpriceableIsFailClosed` — an unpriceable type-7 is **withheld** from the template (fail-closed).

Red-on-master / green-with-fix verified; full 66-test `test_dash_mempool` suite green.

## Scope / honesty

This closes the type-7 auto-prioritise ordering gap only. **Full byte-exact tx-SET parity vs dashd is NOT reached by this PR** and cannot be in one — it remains blocked on the two structural pieces the diagnosis ranked #1/#2:
- special-tx -> CbTx fold (ProTx/collateral-spend -> `merkleRootMNList`), issue #146;
- full-UTXO replay so every input is priceable daemonlessly, the #154 replay-speedup blocker.

Draft — review is the operator's tap (no self-merge).

### DIFF
--- a/src/impl/dash/coin/mempool.hpp
+++ b/src/impl/dash/coin/mempool.hpp
@@ -531,6 +531,34 @@ private:
         auto packed      = ::pack(tx);
         entry.base_size  = static_cast<uint32_t>(packed.size());
         entry.time_added = std::time(nullptr);
+        // W5-B: dashd's ONE automatic mapDeltas write that EVERY node performs.
+        // CTxMemPool::addUnchecked prioritises every MNHF/EHF-signal tx by
+        // +0.1 COIN at admission — src/txmempool.cpp:701-702 (v23.1.7):
+        //     if (tx.nType == TRANSACTION_MNHF_SIGNAL)
+        //         PrioritiseTransaction(entry.GetTx().GetHash(), 0.1 * COIN);
+        // These type-7 EHF-signal txs are ~0-fee on the wire but MUST sort to
+        // the very top of the block template; without the delta the embedded
+        // ancestor-score selector orders them at their base (≈0) feerate and
+        // drops/misplaces them, diverging dashd's tx-set and tx-merkle root.
+        // Unlike the DSTX delta (recorded in add_dstx, network-DSTX-only), this
+        // one is intrinsic to the tx type, so it is reproduced on EVERY
+        // admission path (plain MSG_TX included), exactly as dashd does it in
+        // addUnchecked. Recorded into m_fee_deltas (dashd mapDeltas) so the
+        // entry.fee_delta read just below picks it up AND it survives /
+        // retro-applies identically to the DSTX delta.
+        //
+        // REWARD-SAFE: the delta only raises modified_fee() — the scoring /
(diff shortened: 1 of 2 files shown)


frstrtr:
Adversarial-verify finding (keep as DRAFT, do NOT merge yet): this type-7 MNHF auto-prioritise delta is a faithful dashd port (txmempool.cpp:701-702) and reward-safe, but it is INERT on the DASH served path — embedded_gbt calls the selector with exclude_special=true, so type-7 (special) txs are dropped BEFORE the delta could affect ordering. The proving KAT uses exclude_special=false, a config the serve path never runs. So this advances the --coin-rpc cut by zero until #146 (special-tx→CbTx fold) makes exclude_special flippable. Re-point the KAT at the served (exclude_special=true, post-#146) config when that lands, then merge. Blocked-useful-on #146.

PR #127, opened 2026-06-17
btc(donation): forrestv P2PK on v36 gentx, not LTC P2SH (FLAG6) [cut-only, do-not-merge]

frstrtr:
### Problem
`btc/config_pool.hpp::get_donation_script(int64_t share_version)` returned the
**LTC** `COMBINED_DONATION_SCRIPT` (a 23-byte P2SH `OP_HASH160 <hash160> OP_EQUAL`,
LTC mainnet dev-fund address `MLhSmVQxMusLE3pjGFvp4unFckgjeD8LUA`) for any
`share_version >= 36`. The BTC v36 gentx leg at `share_check.hpp:2940`
(`get_donation_script(int64_t(36))`) therefore emitted the **LTC** P2SH donation
output on BTC v36 shares — contradicting the in-code comment, which states the
BTC donation leg intentionally stays forrestv's canonical **BTC-mainnet P2PK**.

This is **latent**: it fires only once BTC v36 is *confirmed* active. The
deployed canary build (`2f2e8282`, a1) predates the v36 layer, so vote-state is
unobserved — this determination is **static-analysis-backed (code-derived), not
node-confirmed**.

### Fix (BTC-only)
`get_donation_script()` now returns the 67-byte forrestv P2PK
(`0x41` OP_PUSHBYTES_65 · `04ffd03de4…664b` · `0xac` OP_CHECKSIG) on **all**
share versions, including v36+. The LTC `COMBINED_DONATION_SCRIPT` constant is
retained as inert byte data (LTC-only; never selected for BTC). The version arg
is kept for signature parity with the LTC config but is ignored.

Touches only `src/impl/btc/config_pool.hpp`. No other coin's config is affected.

### gentx verification (static)
- `DONATION_SCRIPT` = 67 bytes, leading `0x41` (OP_PUSHBYTES_65), pubkey
  `04ffd03de4…`, trailing `0xac` (OP_CHECKSIG) — forrestv canonical BTC P2PK.
- Post-fix, the v36 gentx path (`share_check.hpp:2940`) writes this 67-byte P2PK
  txout, matching the v35 leg (`share_check.hpp:2387`).

### Scope
Per operator approval (decisions@ ✅) and integrator GO: **cut only** — this PR is
NOT for merge. The consensus merge into a release path requires a separate
explicit operator approval on its own card.

PR #102, opened 2026-06-16
ci: add dgb/pool-pillars-port to PR branch filters (symmetry)

frstrtr:
Additive-only: appends `dgb/pool-pillars-port` to the `pull_request.branches` allow-list in build.yml and coin-matrix.yml, mirroring the existing `btc-embedded` and `dash-spv-embedded` entries already on master.

**Why:** btc-embedded and dash-spv-embedded are on master for symmetry; leaving dgb out makes the canonical filter list inconsistent and invites the next contributor to hit the ZERO-CI footgun before their branch merges up. Redundant once pool-pillars-port lands, but the consistency is cheap (two filter lines, no job/behavior change).

Scope: 2 lines, both `pull_request.branches` only. No push-trigger or job changes.

Merge-gated: holding for operator push-approval per standing rule.

PR #1316, opened 2026-08-23
dash(gbt-xcheck): classify quorumroot-dashd-stale at DKG boundary (keep embedded); mnroot re-scoped

frstrtr:
## Summary

The #127 null-arm reward-safety backstop (`work_source.cpp`) cross-checks our
embedded CbTx `merkleRootQuorums` / `merkleRootMNList` against dashd's
`getblocktemplate` and **hard-swaps to dashd on ANY difference**. That is the
right default — but it has one **inverted** case that this PR fixes, and one
that this PR re-scopes to a follow-up (with an honest correction to the premise).

This closes the classifier half of the researched gap *"Transient
mnroot/quorumroot GBT-xcheck fallback — fold-lag ordering hypothesis vs measured
reality"*. The measured reality: the gap is in the **#127 quorum/MN xcheck
CLASSIFIER**, not in tip/fold ordering. The body-first 4-axis promotion gate
(`coin_state_maintainer.hpp` header-tip / credit-pool-seed / payee-cursor /
sml-currency conjuncts) already prevents serving a tip before its fold, so the
`#102/#119` "tip published before fold" mechanism is **not** present on the
serve path.

## The fix (quorumroot — the inverted-safety case)

**Measured (mainnet, 48h hotel window + dashd RPC chain truth):** at each LLMQ
DKG commitment boundary (a clean 24-block cadence: 2525987, 2526011, 2526035,
2526059, …; episodes <1s) dashd's `getblocktemplate` momentarily serves the
**previous cycle's** quorum root while our embedded arm has already advanced to
the freshly-committed one. Chain truth:

- block `2525987` CbTx `merkleRootQuorums = 8b5f6344…` **== our EMBEDDED root**;
  dashd's own GBT at that instant served the stale `164eeaf7…` (== our root one
  cycle earlier).
- same at `2526011` (chain = `275ded6e…` == embedded).

So in that ~1s window the reward-safety direction was **inverted**: swapping to
dashd served the chain-would-reject `bad-cbtx-quorummerkleroot` template, and it
is the **embedded** root the chain commits.

**Change:** a pure, stateless classifier `coin/gbt_quorum_staleness.hpp`
(`quorumroot_dashd_is_stale`) wired into the quorum/MN xcheck branch. It **keeps
the embedded template** only when **all** of:

1. the quorum roots differ, **and**
2. the MN-list root axis **agrees** (a simultaneous quorum+MN divergence is
   body-divergence, not a pure DKG boundary → ordinary swap), **and**
3. a prior **agreed** quorum root exists **and** dashd's current root **equals**
   it (dashd regressed to the pre-boundary value) **and** our embedded root
   **advanced past** it.

The caller keeps a small `last_agreed_quorum_root_` cache (the most recent root
both arms matched on), updated on every agreeing serve.

### Safety basis
- **Inverse skew** (embedded stale, dashd fresh) makes `dref` the NEW value, so
  `dref != last_agreed` → predicate false → the caller swaps to dashd (the
  reward-safe direction). **No history**, **null sentinel**, **mixed
  quorum+MN divergence**, and a **two-cycle dashd lag** all return false → swap.
- **Byte-neutral** whenever the roots match, which they do for every
  non-boundary template — current hotel behaviour is unchanged outside the ~1s
  DKG windows.
- Runs entirely on the two already-parsed CbTx roots + a member cache; it does
  **not** touch `coin_state_`, so the #1134 off-io ownership boundary is
  preserved (verified: `test_dash_node_coin_state` / the work-source ownership
  guard stays green).
- The `last_agreed` update is guarded by `sel.source == Embedded` so a prior
  creditPool swap (which MOVES `dref`) can never make it read a moved-from
  `dref.m_height`.

## Red → green KAT
`test/test_dash_gbt_quorum_staleness.cpp` (rides the already-allowlisted
`test_dash_stratum_work_source` target — no new NOT_BUILT sentinel):

- pins the classifier on the chain-proven `8b5f6344 / 164eeaf7` vectors and
  every swap-direction case (inverse skew, no-history, null sentinel, mixed
  MN divergence, equal roots, two-cycle lag).
- **RED** (classifier reverted to pre-fix always-swap): `StaleBoundaryKeepsEmbedded`
  and `AnyAdvancedEmbeddedOverRegressedDashdKeeps` **FAIL**; the 6 swap-direction
  cases still pass.
- **GREEN** (with the classifier): 8/8 pass; full binary **123/123**;
  `test_dash_node_coin_state` 103/103 (ownership guard); `test_dash_work_source`
  27/27. Built on gcc-13 Release via the CI Conan toolchain on the build host.

## The mnroot axis — re-scoped, with a premise correction

The research also flagged the `mnroot-mismatch` tokens (29/94 in the window) as a
**by-design body divergence**: our coinbase-only template's tip-state MN root is
chain-valid for its own body; dashd's differs only because its body carries
ProRegTx / ProUpReg / collateral-spend txs whose SML effect it folds
(e.g. 2526182 carried 2 ProRegTx; 2526358 carried 2 `deletedMNs` via collateral
spends inside plain type-0 txs). It is **not** a fold lag, and it **self-closes
under tx-serving** (191 soaks with `--embedded-serve-mempool-txs` armed show 0
mnroot / 0 quorumroot events across all 4 logs).

I did **not** ship an mnroot code change in this PR, for two honest reasons:

1. **Premise correction.** The gap suggested mirroring "the creditPool
   modulo-special pattern" to *keep embedded*. But the creditPool `explained`
   branch (`work_source.cpp:~967`) actually still **serves dashd** — it only
   relabels so the graduation stats stop counting a by-design exclusion as
   drift. A true *keep-embedded* mnroot decision needs strictly more than the
   creditPool precedent provides.

2. **A correct keep-embedded mnroot classifier requires a recompute the serve
   thread structurally cannot do here.** To prove the divergence is fully
   body-explained (not an embedded SML bug) we must apply dref's body MN-effects
   (type-1/2/3/4 special txs **and** type-0 collateral-spend removals — the
   2526358 case has **zero** MN-type txs, so type inspection alone is
   insufficient) to a copy of the **tip SML** and check the recomputed root ==
   dref's. The tip SML lives in `coin_state_`, which the xcheck block runs on
   the sourcing thread and is **forbidden** to touch (#1134). Doing it means
   plumbing a ~450 KB SML (+ the collateral index) across the ownership boundary
   into `CoinStateArm`, on the reward path, plus a byte-exact re-implementation
   of dashd `RebuildListFromBlock` (type-1/2/3/4 apply incl. the legacy↔basic
   operator-key scheme — the #154 SML-parity wall). That is a genuinely large,
   reward-critical change; half-doing it on a reward path is worse than not.

Because the current mnroot behaviour is **already reward-safe** (it serves
dashd's own valid template) and self-closes under tx-serving, the mnroot axis is
correctly a **classification refinement**, not a safety fix, and is deferred
behind this contained, higher-value quorumroot fix.

### mnroot follow-up spec (for the deferred PR)
- Copy the tip SML (+ collateral index) into `CoinStateArm` in
  `resolve_coin_state_arm()` (owning thread) so the sourcing thread can read it
  by value without crossing #1134.
- New pure helper `mnroot_body_explained(tip_sml, mnstates, dref.m_txs,
  dref_root)`: apply the confirmation pass (`project_sml_confirmations`, exists),
  the collateral-spend removals (`tx_spends_mn_collateral`, exists), and the
  type-1/2/3/4 SML applies (new, byte-exact vs dashd), recompute `CalcMerkleRoot`
  and compare to `dref_root`. On equality classify `mnroot-modulo-body-explained`
  and KEEP embedded (our coinbase-only root is correct for our coinbase-only
  body); otherwise swap.
- Also evaluate the #130 txmerkle branch **before** the quorum/MN branch when
  `--embedded-serve-mempool-txs` is armed, so a genuine body divergence gets its
  correct label instead of a quorum/MN one.

## Merge / deploy gate

**reward_or_consensus = true.** This changes a swap decision on the reward path
(it makes the embedded arm win in the DKG-boundary window instead of swapping to
dashd). **Do not auto-merge.** Merge and any hotel deploy are **operator-gated**,
after CI + a canary soak that confirms the `quorumroot-dashd-stale (KEEP
embedded)` classification fires exactly on the 24-block boundaries and never
produces a reject on hashpower.


frstrtr:
Superseded by #1322 (dash daemonless full-stack serve-path landing — the 8-PR consolidation, merged 2026-08-23). This PR's content is already on master via #1322; its branch is based on the pre-#1322 divergent line (d1000d605 era), so merging now would revert master. Closing as superseded — branch retained, reopenable if a unique net-new delta is identified.

PR #1319, opened 2026-08-23
dash(daemonless): close Window-2 eviction lag — utxo-stale-at-tip serve gate

frstrtr:
## Window-2 close: stop throwing away rarely-found blocks

Ports dashd's `CTxMemPool::removeForBlock`-before-`SetTip` invariant to the embedded serve path as a **fail-closed currency gate**, so a winning share is never wasted on an intra-node ordering race. **Default OFF; reward/serve path → operator-gated merge.**

### The two windows
- **Window 1 (untouched)** = propagation lag: our tip is still H-1, block H unseen. Serving H's competing txs is normal — dashd serves the same set, it's a valid fork competitor, self-healing. This PR does not change it: when the serve tip is H-1 the UTXO view is also H-1, the predicate is true, nothing is suppressed.
- **Window 2 (closed here)** = intra-node eviction lag: the serve tip was promoted to **H** by the diff-driven path (`CoinStateMaintainer::on_mn_list_update` / `on_mnlistdiff`, `coin_state_maintainer.hpp:511` / `:906-910`) **before** the embedded UTXO lane connected+evicted block H. In that sub-second window (median tip-body **0.771s**, p90 **2.9s**) the UTXO view is still at H-1, so H's spent outpoints resolve as unspent and the selection stale-input guard (`mempool.hpp:944-949`) is structurally blind to them → a template for H+1 packs a tx already spent by H → `bad-txns-inputs-missingorspent` on a winning share = **a thrown-away found block**.

dashd can't expose Window 2 because `removeForBlock` runs synchronously inside `ConnectBlock` before `SetTip`. c2pool can't hold one lock across two async network events (header vs body arrival), so we reproduce the invariant fail-closed.

### The change
1. **`utxo_lane.hpp`** — `UtxoLane::utxo_current_at(tip_hash)`: the cache flushes its `best_block` atomically per `connect_one` (which also runs `Mempool::remove_for_block` on the same io thread before returning), so `get_best_block()==tip_hash` is an exact "H connected AND evicted" test.
2. **`node_coin_state.hpp`** — `set_utxo_current_fn` + the decision folds `utxo_stale_at_tip = m_serve_mempool_txs && fn-set && !fn(m_prev_hash)` into `suppress_mempool_txs` with a NEW most-specific, never-silent cause **`utxo-stale-at-tip`**. Fn unset ⇒ byte-identical to prior behaviour.
3. **`main_dash.cpp`** — wire the fn beside `set_utxo_ready_fn`. On the LIVE full-block path the utxo-lane subscriber (`:2438`) connects+evicts *before* the maintainer promotes the serve tip (`:3908`) on one io thread, so the predicate is already true there — the live path had no gap; the diff path is the one this closes.
4. **`embedded_gbt.hpp`** — unchanged; `suppress_mempool_txs` already threads to the coinbase-only body + telemetry.

Degraded mode is coinbase-only (consensus never requires a mempool tx; fees exactly 0, every committed value exact — same argument as `utxo-immature-serving`), **not** a promotion stall: the diff path exists to serve promptly when the body is late. Strictly tightening; self-heals on body arrival via the existing recompute path.

### KAT — `test_dash_node_coin_state` (`test_dash_embedded_gbt` currently fails to LINK on the pre-existing #154 `opkey_*` gap, so the gate KAT lives in the node-coin-state target that does build)
- `StaleTipSuppressesBodyWithNamedCause` — **RED→GREEN in one place**: fn UNSET (== shipped master) serves the doomed tx (`suppress=false`); fn WIRED with a stale view suppresses, cause `utxo-stale-at-tip`. RED additionally proven by neutering the gate branch (rebuild+run: the stale assertion fails, body served) and GREEN with it restored.
- `CurrentTipServesFullSet` — view current at tip ⇒ full set (self-heal).
- `InertWhenServingOff` — serving OFF keeps cause `mempool-txs-disabled` (gate can't fire unarmed; default posture byte-unchanged).
- `ImmatureCauseWinsOverStale` — cold-start immature cause wins the naming.
- `DoomedTxAbsentFromTemplateWhenSuppressed` — `build_embedded_workdata` twice: T selected when served, **absent** when the gate suppresses.

Full suite **108/108 green**; `c2pool-dash` binary builds with the wiring (workstation-191, ccache).

### Reorg / disconnect
Out of scope by construction. The daemonless UTXO lane connects forward only (`on_block_connected` `height > best_height` guard); there is no disconnect/undo consumer wired — reorgs are handled by re-derivation, not mempool re-admission. The gate is fail-closed, so a serve tip that moves across a reorg without the view catching up serves coinbase-only until current; re-admitting a disconnected block's txs is a fee-completeness (%-only) concern with no seam to hook. dashd's `removeExpiredAssetUnlock` analog is the Type-9 lane (#143) and likewise stays out of scope.

### Safety
Default OFF and byte-identical unless `--embedded-serve-mempool-txs` is armed with a live UTXO lane. Coinbase-only, dashd-fallback and unarmed postures unchanged. **Not for merge without operator tap.**

_Optional follow-up (not in this PR): a startup wiring assert pinning the utxo-lane subscription before the maintainer's when both are armed, so a future re-wire can't invert the already-correct live-path order. The gate makes any inversion fail-closed regardless, so it's hardening, not a blocker._


frstrtr:
Superseded by #1322 (8-PR full-stack serve-path consolidation, merged 2026-08-23) — content already on master; branch predates #1322 so merging would revert. Closing superseded, branch retained.

PR #1321, opened 2026-08-23
dash daemonless: cross-restart cumulative serve-gate accounting + QC-NULL-SERVE counters (#127 null-arm soak measurability)

frstrtr:
## What

Closes the **cumulative-accounting gap** the null-arm completeness critic found: the `[EMBED-GATE-ROLLUP]` denominator `observed=` resets every restart (measured on hotel-primary 2026-08-23: `observed=7211s` at the 06:00 tick = wall-clock since the 04:00 restart), and there is **no explicit null-served counter**, so the dashd-cut gate — *"the null-arm covered the 4.51% DKG floor with 0 rejects over a **standing** soak"* — cannot be proven program-level.

This is **measurement only**. Nothing on the reward/consensus path changes and the flag defaults **OFF**.

## Why the serve path is untouched

R1/R2 re-confirmed against master that the null-arm serve-exactness is already correct: `daemonless_qc_commitments` tries `cache->verified_for` (real BLS-verified commitment) **first** and consults `null_evidence` **only on a miss**; `build_null_commitment` is the all-false byte-shape dashd itself emits on a local cache miss (`blockprocessor.cpp:748-763`); any unproven freshness fails the **whole height** closed to the fallback. So the null-arm fires on exactly the DKG-floor tips and never when a real quorum exists — the gap the critic flagged is the **accounting**, not the serve.

## Changes (in-scope: quorum/null-arm serve path + EMBED-GATE accounting only)

- **`serve_gate_persist.hpp`** (new) — the I/O layer `ServeGateJournal` deliberately isn't:
  - `ServeGateCumulative` (restart_count + observed/off_embedded/per-cause seconds + `QcNullServeCounters`) moved through a small JSON state file;
  - `combine(prior, live_rollup, live_counters) = prior + live`; the file holds **only** frozen prior totals, so write-through on every tick **replaces** (never adds) the live part — N saves == one;
  - load (missing/corrupt ⇒ clean zero, restart_count bumped) + **atomic** temp-file+rename save;
  - `QcNullServeCounters` — per-LLMQ-type `window_open/null_served/real_served/fell_back` with the invariant `window_open == null+real+fell_back`, plus the global never-a-reject conjunct `submits_from_null / rejects_from_null`;
  - `cumulative_rollup_line` / `qc_null_serve_line` renderers.
- **Live wiring** flag-gated by `--serve-gate-state-file PATH` (empty default ⇒ **OFF**, byte-identical serve path). When set, `note_arm_decision` loads prior once and, on each hourly roll-up tick, write-throughs the combined state and prints `[EMBED-GATE-ROLLUP-CUM]` + `[QC-NULL-SERVE]` alongside the per-process line. File I/O runs **off** `serve_gate_mutex_`.
- Public hooks `note_null_serve_disposition` / `note_null_submit`. The per-tip disposition call site is the `daemonless_qc_commitments` fold consumer, deliberately left as a **soak-gated follow-up** (it touches the consensus fold).

## KAT (red → green)

`test_dash_serve_gate_journal` (allowlisted, header-only-friendly, already links `nlohmann_json` — avoids the pre-broken `test_dash_embedded_gbt` opkey link gap). 5 new cases, **RED on master** (the new header does not exist there → TU does not compile, same red shape as `NoWorkIsTheDistinctThirdArmState`) / **GREEN here**:

- `StateSurvivesRestartAndAccumulates` — seconds **and** null-serve counters accumulate across 3 simulated processes; `restart_count 1→2→3`. The per-process wipe is gone.
- `NullServeCountersLockServeExactness` — the disposition invariant locks that the null-arm is counted null **only** on floor tips and real wherever a verified commitment exists; `rejects==0`.
- `MissingOrCorruptFileYieldsCleanZero`, `CombineIsReplaceNotAddWithinOneProcess`, `LogLinesRenderCumulativeShape`.

Full suite **17/17 green**; `c2pool-dash` binary builds + links with the wiring (workstation-191, gcc13/Release).

## Not merged — operator-gated

Serve-path-adjacent (money/consensus). Fix item **(c)** — flip `--embedded-null-arm` default / pin `--serve-gate-state-file` in the money-node drop-in — is an operator decision, made once (a)+(b) prove the standing soak.

### DIFF
--- a/src/c2pool/main_dash.cpp
+++ b/src/c2pool/main_dash.cpp
@@ -302,6 +302,12 @@ std::string g_replay_mnlist_seed_file;        // --replay-mnlist-seed-file FILE 
 // serve are both fine WITHOUT proving the serve is daemonless.
 bool        g_no_dashd_mn_seed = false;       // --embedded-no-dashd-mn-seed
 
+// --serve-gate-state-file: path to the cross-restart cumulative serve-gate
+// accounting state (JSON). Empty (default) => OFF; the serve path is
+// byte-identical. Non-consensus, non-reward: a soak convenience whose
+// bad read fails to a clean zero and never wedges the node.
+std::string g_serve_gate_state_file;           // --serve-gate-state-file
+
 // ── PR-2 FORWARD: --replay-mined-commitment-index ─────────────────────────
 // Arms the forward half of dashd's mined-commitment store
 // (mined_commitment_index.hpp, ported from v23.1.7 llmq/blockprocessor.cpp)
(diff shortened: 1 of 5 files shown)


frstrtr:
Superseded by #1322 (dash daemonless full-stack serve-path landing — the 8-PR consolidation, merged 2026-08-23). This PR's content is already on master via #1322; its branch is based on the pre-#1322 divergent line (d1000d605 era), so merging now would revert master. Closing as superseded — branch retained, reopenable if a unique net-new delta is identified.

PR #1324, opened 2026-08-23
dash(daemonless): self-validate served tx set from OUR OWN state — demote dashd validity gate to measurement

frstrtr:
## DASHD-FREE self-validated own-set tx serving

Successor to the tx-serve referee (#1314) and to the validity-gate window-1 work (#1318). Makes the **serve decision self-sufficient for the `--coin-rpc` cut**: deciding *what* to serve now makes **zero dashd calls**. All behind the existing `--embedded-tx-serve-own-set` flag — **default OFF is byte-identical**.

### What changed
1. **Demote the dashd testmempoolaccept validity gate out of the serve decision.** Own-set serving previously required `validity_gate_open()` (576 clean testmempoolaccept heights) — a *permanent* dependency on dashd's external view. The serve decision now depends **solely on the internal-consistency referee over our own state** (coinbase == subsidy + Σfee + superblock; every served tx `fee_fold_proven` against our spent-aware UTXO view at build; no intra-set double-spend). `validity_gate_open()` remains a **pre-cut confidence measurement** (logged, never gated); post-cut it simply vanishes.
2. **Run the referee unconditionally on the embedded serve path.** It used to live only inside the `gbt_xcheck_ && dashd_fallback_` divergence branch, so post-cut (rpc==null) an armed template served with *zero* serve-time self-validation. It now runs on every armed embedded template carrying mempool txs, making zero dashd calls. Fail action is **dashd-free**: refuse the internally-inconsistent template (fail-closed to no-work; last good work held) — never ask dashd, never hand-strip a reward-critical coinbase at serve time.
3. **Classifier successor to #1318 (the h=2526495 incident class).** `missing-inputs` while dashd's probe-time tip is *strictly ahead* of our serve parent is a **Window-1 propagation artefact**, not a defect — both leaked txs at 2526495 were confirmed *in block 2526495 itself*, the exact height our template competed for; valid on our fork. `classify` now emits `PendingPropagation` (advances nothing, resets nothing, counted separately so the skew stays measurable) instead of `Invalid`, keyed on a **fact** (`dashd_ahead_of_serve_height`, a height comparison from dashd's GBT height) never a reason-string. Narrowing is strict: same reject **without** the ahead fact stays `Invalid`; `bad-txns-inputs-missingorspent` stays `Invalid` even under the skew.

### Residual (stated, not hidden)
Own state validates UTXO-spentness / no-double-spend / coinbase fee-exactness / topology / maturity / islock / already-confirmed — but **not tx scripts/signatures** (c2pool ports no `CheckInputScripts`). The reliance is the **sole-ingestion-path invariant**: every pooled tx arrived only via dashd-peer relay (full script checks ran there) and we never mutate bytes. That invariant **survives the `--coin-rpc` cut** (coin-P2P peers remain full nodes; `testmempoolaccept` was never the script check). It **breaks at T3.1 miner tx-injection (#157)** — script validation becomes mandatory there, a T3.1 gate, not a cut gate.

### Test — `dash_tx_serve_self_validation_kat` (no node, no daemon)
- **PART A (classifier, red→green):** `missing-inputs`+dashd-ahead → `PendingPropagation` (was `Invalid`); same without the fact stays `Invalid`; `missingorspent` stays `Invalid`; a PendingPropagation-only height does **not** reset the clean run; a genuine `Invalid` still does.
- **PART B (referee, zero dashd):** serves our own valid set; refuses double-spend / coinbase over-claim / unstamped template — a pure function of the template.

Verified **red** (incident assertion fails with the new branch disabled) → **green** (12/12). `c2pool-dash` links clean on the 191 build.

### Rollout
Serve/reward path. **Default OFF, operator-gated. Do NOT merge and do NOT restart the money node without the operator's tap.** Base is `integrator/tx-serve-own-valid-set` (#1314); review this delta on top of the referee.

### DIFF
--- a/src/c2pool/CMakeLists.txt
+++ b/src/c2pool/CMakeLists.txt
@@ -361,6 +361,31 @@ target_link_libraries(dash_hdr_backfill_window_kat
 )
 add_test(NAME dash_hdr_backfill_window_kat COMMAND dash_hdr_backfill_window_kat)
 
+# KAT: DASHD-FREE self-validated own-set tx serving (successor to #1318). Pure
+# red/green over two header-only units with NO node and NO daemon: the
+# mempool-validity classifier (Window-1 dashd-ahead propagation -> demoted to
+# measurement, does not reset the clean run) and the serve-time internal-
+# consistency referee (own-state self-validation: coinbase fee-exact, no
+# intra-set double-spend, fee_fold_proven) that decides own-set serving with
+# ZERO dashd calls. Links the same DASH library set as c2pool-dash.
+add_executable(dash_tx_serve_self_validation_kat dash_tx_serve_self_validation_kat.cpp)
+target_link_libraries(dash_tx_serve_self_validation_kat
+    core pool sharechain
+    c2pool_node_enhanced
+    dash
+    c2pool_payout c2pool_merged_mining c2pool_hashrate c2pool_storage c2pool_difficulty
+    dash_x11
+    dash_rpc
+    dash_stratum
+    dash_block_replay
+    dash_bls_verify
+    btclibs
+    yaml-cpp::yaml-cpp
+    nlohmann_json::nlohmann_json
+    ${Boost_LIBRARIES}
+)
+add_test(NAME dash_tx_serve_self_validation_kat COMMAND dash_tx_serve_self_validation_kat)
+
 # Windows: Boost.Asio requires Winsock libraries; /bigobj for large translation units
 if(WIN32)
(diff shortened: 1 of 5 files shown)


frstrtr:
Superseded by #1325 — same delta rebased clean onto master (6e0318cf8, post-#1322 which landed the 8 serve-path PRs incl the referee base #1314). #1324 was stacked on #1314's pre-#1322 branch. Branch integrator/tx-serve-dashd-free-selfvalidate is preserved (not deleted).

PR #1311, opened 2026-08-22
dash daemonless: pay block H's masternode from the pre-block payout (fixes same-block-ProUpReg payee_desync fail-close @ h=2525714)

frstrtr:
## consensus / money-path — needs operator merge tap. Do NOT auto-merge.

Draft for review + tracking. The masternode payee cross-check is a reward-path consensus check, so this stays gated on an operator merge tap even when CI is green.

## Root cause — same-block ProUpReg LIVE-READ

`MnStateMachine::apply_block` (`src/impl/dash/coin/mn_state_machine.hpp`) cross-checked the coinbase against the projected payee's **live** `scriptPayout`, read in **pass 3 — AFTER pass 1 folds the block's own special txs**.

When the projected payee's `ProUpRegTx` rides the very block it is paid in, pass 1 rewrites `scriptPayout` before pass 3 reads it. dashd pays block H's masternode from the **H-1** list — `payments.cpp GetBlockTxOuts` reads `oldList.GetMNPayee(pindexPrev)->pdmnState->scriptPayout`, i.e. the payout as of H-1, before `deterministicmns.cpp BuildNewListFromBlock` folds the ProUpReg in (`specialtxman.cpp:399`). So the coinbase correctly pays the pre-block script but pass 3 no longer finds it — a **FALSE `payee_desync`** that fail-closes the daemonless masternode-set bridge and blocks the full dashd-cut.

Observed on mainnet **h=2525714**: proTx `f5a49888ec1d00db`, in-block ProUpReg moves `scriptPayout a7f7252d -> Xv4Ys1mi`; the coinbase pays the pre-block `a7f7252d`.

## Fix

Snapshot the projected payee's payout script at **pass 0** (before pass 1) and cross-check the coinbase against that pre-block script — mirroring dashd `GetBlockTxOuts`, exactly as `DmlFoldEngine::fold_block` already captures `payee_script_pre` (`replay_fold_engine.hpp`). Tip commit `832fa7703`, `src/impl/dash/coin/mn_state_machine.hpp` (+26 lines, 2 hunks).

## KAT

`DashMnState.ProjectedPayeePaidByPreBlockScriptWhenSelfProUpRegSameBlock` pins h=2525714 **RED before** the snapshot (`payee_desync`, paid=0) → **GREEN after** (paid=1). Full `DashMnState` suite green (41/41 on this build).

## End-to-end proof — pure-daemonless fold (the decisive gate)

[...]

PR #1340, opened 2026-08-24
dash(daemonless): PR-C3 ConnectBlock fold-before-publish ordering (flag default-OFF, byte-identical master)

frstrtr:
## PR-C3 — ConnectBlock fold-before-publish ordering (dormant defense-in-depth)

Adds `--embedded-connectblock-ordering` (default OFF, byte-identical to master when absent). When armed on the body-first coin-P2P arm, `NodeCoinState::set_tip()` refuses to advance the published serve-tip to a block whose credit-pool axis is not yet folded (early-return, no derived-state write — strictly more fail-closed than master).

### Scope correction (measured, not claimed)
Review of the wired path shows the per-tip creditpool/payee/dmn-SML publish window is **already closed** by the pre-existing 4-axis `maybe_promote_pending_tip` -> `promote_serve_tip` machinery (header + credit-pool-seed + payee-cursor + SML-currency all proven current at the promoted block before `republish()->set_tip`). On that path `connectblock_ordering_deferrals == 0` — this guard never fires. It is therefore a **dormant defense-in-depth backstop on the credit-pool axis**, not the mechanism that closes the window.

The one residual — the serve tip promoting ahead of the quorum-folded state (`emit-qc-*`) — is caught fail-closed at emit time (no wrong quorum root is ever emitted) and is the known qc-availability propagation floor, addressed by the MinedCommitmentIndex arming + null-arm, not by fold-ordering. This PR does not close it.

### Safety
- Flag OFF => byte-identical to master (single guarded early-return, `m_connectblock_ordering` default false; new members read by no other logic). Proven by KAT `OffPublishesWithoutFold` + untouched 65/65 maintainer + 109/109 node_coin_state suites.
- When armed: only ever REFUSES to advance to a not-yet-folded tip (retains the prior fully-folded tip). No derived value is ever changed — reward-safe ordering discipline only. Re-run on every maintainer event + independent `check_tip_body_overdue` demote => cannot wedge.
- KAT (5 cases) folded into the allowlisted `test_dash_coin_state_maintainer` target (no NOT_BUILT fake-green).

Reward-safe, dormant, off the dashd-cut critical path.

### DIFF
--- a/src/c2pool/main_dash.cpp
+++ b/src/c2pool/main_dash.cpp
@@ -992,7 +992,17 @@ int run_node(bool testnet, const std::string& rpc_endpoint,
              // of THIS node's own dial targets — never changes WHAT is fetched
              // or how a reply is verified. OFF => dial plan byte-identical to
              // master.
-             bool embedded_asn_diversity = false)
+             bool embedded_asn_diversity = false,
+             // --embedded-connectblock-ordering (PR-C3): DEFAULT OFF. Port of
+             // dashcore ConnectBlock atomicity to the tip-publish seam
+             // (NodeCoinState::set_tip). When on, the arm refuses to publish a
+             // tip to the work source ahead of that tip block's body-derived
+             // credit-pool fold — closing the intra-node per-tip ordering
+             // window (#1154) that surfaces as creditpool/dmn/payee-stale.
+             // Ordering discipline ONLY (no derivation change); composes with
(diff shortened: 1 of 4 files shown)


frstrtr:
Closing: C3 fold-before-publish ordering premise refuted — window is pre-closed on master, reorder is a no-op. Not a node in the daemonless land chain (#1311 -> #1314 -> #1315 || #1318 -> #1316). Green was cosmetic, not landing-ready.

PR #1368, opened 2026-08-27
Peer-found blocks invisible on dashboard: carry share PoW across threads (LTC/BTC/DGB/BCH)

frstrtr:
## Live defect (ltc.voidbind.com, 2026-08-27)

Operator report: p2p-spb found a valid DOGE block; neither the DOGE block nor the LTC-side block appeared on ltc.voidbind.com.

Verified on the live node (contabo, merged LTC+DOGE v36):

- 13:57:26 CEST the node received peer share `521906df2ba641bf` and logged `[BLOCK] Peer share meets block target` -- a real LTC-target hit.
- That share carried an ACCEPTED merged DOGE block: height 6349371, hash `2f3e4d8aa870c238...`, coinbase tag `/P2Pool v36/`, is_aux=true, 10,001.7 DOGE reward (confirmed on the DOGE chain via blockchair).
- The LTC counterpart `521906df...` is NOT in the canonical LTC chain (litecoinspace: not found) -- genuinely orphaned/lost at the source node, not a c2pool defect.
- Neither `GOT BLOCK FROM PEER!` nor `*** MERGED BLOCK FROM PEER!` ever fired; `/recent_blocks` held only 2 stale relay rows (found_by=unknown) and `/recent_merged_blocks` had nothing newer than Aug-12.

## Root cause

Share reception is two-phase: phase 1 runs `share_init_verify()` (expensive PoW) on `m_verify_pool` WORKER threads; `attempt_verify()` runs later on the COMPUTE thread and calls `share_init_verify()` with `check_pow=false` (the hash is already set). The block verdict and pow hash lived only in thread_local scratch (`g_last_init_is_block` / `g_last_pow_hash`) set on the worker thread -- invisible to the compute thread. So for EVERY peer share:

- `m_on_block_found` never fired (flag false on the compute thread) -> no `record_found_block`, no `schedule_block_verification`;
- `m_on_merged_block_check` was gated on a null/stale pow -> peer-found merged (DOGE) blocks never detected;
- latent corruption: a STALE `g_last_pow_hash` from an unrelated share could be cached as `idx->pow_hash` for the wrong share.

Dash is unaffected (X11 identity hash is recomputed in-thread on every verify). DOGE has no own tracker (aux leg of LTC).

## Fix (coin-generic)

- `chain::BaseShare` gains transient `m_pow_hash` (never serialized). Phase-1 reception and local share creation store the freshly computed PoW on the share -- same thread, same call.
- `share_init_verify()` resets the thread_local scratch unconditionally at entry so `check_pow=false` calls can never expose stale values.
- `attempt_verify()` resolves the pow (share-carried -> same-thread recompute -> index cache) and derives the verdict from `pow <= min_header block target`; the merged check receives the resolved pow. No cross-thread thread_local reads remain in the trackers.
- The LTC peer-found record now labels authorship `sharechain_peer` (climb-only; cannot downgrade a local `this_node` record).

## Verification

- Structural KAT `peer_block_pow_carry_test` (folded into the ltc `share_test` target) pins all three legs across btc/ltc/dgb/bch; restoring the thread_local consumption or dropping the phase-1 carry goes red.
- ltc `share_test`: 97/97 pass. `btc_share_test`: 199/199 pass.
- `c2pool` (LTC lane), btc/dgb libs and `c2pool-bch` all build clean (gcc13 CI profile).
- Reward-path untouched: no consensus, PPLNS, or share-validation semantics change; the only behavioral change is that the already-wired found-block reporting callbacks now actually receive the events they were written for.

Source-side context: the finder node (p2p-spb) was intermittently peerless and had lost daemon contact around the find window, which explains the LTC orphan and delayed propagation. That does not diminish this defect -- the share DID reach this node, met the block target in its logs, and was still not surfaced.

### DIFF
--- a/src/c2pool/main_ltc.cpp
+++ b/src/c2pool/main_ltc.cpp
@@ -4331,10 +4331,15 @@ int main(int argc, char* argv[]) {
                              << " miner=" << miner_addr.substr(0,16)
                              << " height=" << height;
 
+                    // sharechain_peer: this hook fires for ANY verified share
+                    // meeting the block target. For our own find the submit
+                    // path records this_node and authorship only climbs, so
+                    // labelling peer here can never downgrade a local record.
                     mi->record_found_block(
                         height, block_hash, s->m_min_header.m_timestamp,
                         is_testnet ? "tLTC" : "LTC",
-                        miner_addr, share_hash.GetHex(), net_diff, share_diff, pool_hr, 0);
+                        miner_addr, share_hash.GetHex(), net_diff, share_diff, pool_hr, 0,
+                        core::MiningInterface::BlockAuthorship::sharechain_peer);
                     mi->schedule_block_verification(block_hash.GetHex());
                 });
             };
--- a/src/impl/bch/node.cpp
+++ b/src/impl/bch/node.cpp
@@ -392,6 +392,13 @@ void NodeImpl::processing_shares(HandleSharesData& data_ref, NetService addr)
                     {
                         share.ACTION({
                             obj->m_hash = share_init_verify(*obj, true);
+                            // Carry the PoW result WITH the share: the
+                            // thread_local scratch set here (verify pool)
+                            // is invisible to the compute thread that runs
+                            // attempt_verify() later. Without this, peer
+                            // shares meeting the block target were never
+                            // reported (block-found + merged callbacks).
+                            obj->m_pow_hash = g_last_pow_hash;
                         });
                     }
                     catch (const std::exception&)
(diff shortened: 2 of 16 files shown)


frstrtr:
Evidence closure: parsed the AuxPoW section of the accepted DOGE block 6349371 (raw block via blockchair). The parent block header at the end of the AuxPoW hashes (SHA256d) to exactly

`521906df2ba641bf155313b2574a9196070e97fcdaa64d87c04257aded38e96d`

which is the sharechain share this node received at 13:57:26 CEST, verified into the chain (chain=37776, verified=20970), and flagged with `[BLOCK] Peer share meets block target` -- and then failed to surface. The accepted DOGE block and the orphaned LTC block are the same PoW event, carried by that one share; both were dropped by the thread_local gap this PR closes.

PR #1412, opened 2026-08-30
dgb: port block subsidy from DigiByte Core (COIN=1e8), fix 100x embedded-coinbase underpay

frstrtr:
## Summary

CONSENSUS / MONEY-CRITICAL. The DGB embedded coinbase-build path underpaid the block reward ~100x. This ports the DGB block subsidy from DigiByte Core (COIN=1e8) so a solved DGB block pays the real reward. Report-first; **do not merge or deploy without operator review** — this changes a live money path and reverses an earlier ruling.

## The bug

`dgb::CoinParams::subsidy` (config_coin.hpp) replicated the p2pool-dgb-scrypt display oracle, which carries **COIN=1e6**. That oracle never built a coinbase in python p2pool — the real value came from `digibyted`'s GBT `coinbasevalue`, so its 1e6 unit was a display artifact. In c2pool the same function feeds the **live daemonless coinbase-build path** (`embedded_coinbase_value.hpp`), where the returned value is written raw as the coinbase vout in satoshis. Consensus (DigiByte Core `amount.h`) is **1e8**, and there is no scale anywhere on the path:

```
params.hpp (verbatim lambda) -> embedded_coinbase_value.hpp (subsidy+fees, no scale)
  -> work_source.cpp (verbatim) -> gentx_coinbase.hpp (raw 8-byte LE vout write)
```

So at the current Scrypt tip the built coinbase paid **2.53558070 DGB** instead of the on-chain **253.55810338 DGB** — a 100.0000132x underpay.

It has never fired: no DGB block has been mined through the embedded path, and the external-GBT fallback takes `digibyted`'s value verbatim. It would fire the **first time real hashrate solves a block** on the daemonless node (contabo, stratum :9434). DigiByte Core accepts underpay (ConnectBlock rejects only *overpay*, `bad-cb-amount`), so every such block would be **VALID but silently destroy ~251 DGB** of reward.

## Ground truth (triple-proven)

1. **DigiByte Core** `GetBlockSubsidy(nHeight, consensusParams)` (`src/validation.cpp`), COIN=1e8 (`src/consensus/amount.h`), mainnet constants (`src/kernel/chainparams.cpp`).
2. **Recomputed** from the Core formula at the tip -> `25,355,810,338` sat.
3. **Live chain**: on-chain coinbase at height 24,125,022 = `25,355,810,338` sat (253.55810338 DGB) — byte-identical.

## Fix

Port `GetBlockSubsidy` exactly: COIN=1e8, master floor semantics (sub-1-DGB reward clamped to 0). The six-period schedule shape, constants, truncating decay loops, and the Period VI `months = blocks * 15 / 2628000` count already matched Core — only the unit and floor differed. A naive x100 rescale at the build site is **not** parity: the 1e6 decay loop truncates coarser and lands thousands of sats off (e.g. x100 = 25,355,807,000, short 3,338 sat). The decay must run at full 1e8 precision.

This supersedes the earlier "the oracle IS the spec" ruling — for a real coinbase the spec is DigiByte Core. The schedule is *more* correct than the stale 3-phase farsider C module (which predates the block-1,430,000 DigiSpeed fork and returns ~125 DGB), so this does **not** move toward farsider. The DOGE lane's `get_doge_block_subsidy` was spot-confirmed already exact and is untouched.

## Red / green proof

Compiled the shipped `subsidy()` function text before and after the change:

| height | before (1e6, shipped) | after (1e8, ported) | DGB-Core / on-chain |
|---|---|---|---|
| 24125022 (live tip) | 253,558,070 | **25,355,810,338** | 25,355,810,338 |
| 67200 | 7,960,000,000 | 796,000,000,000 | 796,000,000,000 |
| 400000 | 2,434,410,000 | 243,441,000,000 | 243,441,000,000 |
| 1430000 | 1,078,500,000 | 107,850,000,000 | 107,850,000,000 |
| 20000000 | 331,934,836 | 33,193,486,585 | 33,193,486,585 |

Before: 5/5 mismatch (RED). After: 5/5 match + a 644-point full-domain byte-parity sweep against an independent Core reference (GREEN).

## Tests

`test/test_dgb_subsidy.cpp`, `test/test_dgb_coinbase_value.cpp`, `src/impl/dgb/test/coinbase_value_parity_test.cpp`, `src/impl/dgb/test/work_source_test.cpp` updated to DigiByte Core vectors (satoshis, 1e8). `test_dgb_subsidy` adds the live-tip on-chain vector and a full-domain byte-parity sweep vs an independent Core reference implementation.

## Do not

- Merge without operator go (consensus/reward path).
- Point real hashrate at the DGB pool until this is merged and deployed.

### DIFF
--- a/src/impl/dgb/coin/embedded_coinbase_value.hpp
+++ b/src/impl/dgb/coin/embedded_coinbase_value.hpp
@@ -42,7 +42,18 @@ namespace dgb::coin
 // Mirrors src/impl/ltc/coin/template_builder.hpp:
 //     uint64_t coinbasevalue = subsidy + total_fees;
 // but sources the subsidy through the coin's CoinParams::subsidy_func SSOT
-// (the DGB oracle decay schedule) rather than a hardcoded halving formula.
+// (the DigiByte Core GetBlockSubsidy port) rather than a hardcoded halving
+// formula.
+//
+// UNIT RECONCILIATION (money-path invariant): both terms are DGB satoshis
+// (1e8/DGB). subsidy_func == dgb::CoinParams::subsidy now returns DigiByte Core
+// GetBlockSubsidy() output in satoshis (COIN=1e8), and total_fees is the raw
+// satoshi sum of the included transactions' fees (the same unit digibyted folds
+// into GBT "coinbasevalue"). The sum is therefore unit-consistent and needs NO
+// rescale here. Before the subsidy fix the SSOT returned 1e6 units while fees
+// were 1e8 sats -- a mixed-unit coinbasevalue that underpaid ~100x. The fix is
+// entirely in the subsidy SSOT; this addition must stay a bare '+' with no
+// scaling factor (a stray x100 here would double-count the fix and overpay).
 //
 // Throws std::logic_error if subsidy_func is unset — an empty std::function
 // here means the CoinParams factory was not wired, which must fail loudly at
--- a/src/impl/dgb/config_coin.hpp
+++ b/src/impl/dgb/config_coin.hpp
@@ -119,65 +119,81 @@ public:
     static constexpr uint32_t DUMB_SCRYPT_DIFF = 65536;  // 2^16
 
     // -----------------------------------------------------------------------
-    // DGB subsidy schedule -- EXACT replica of the p2pool-dgb-scrypt oracle
-    // get_subsidy() (bitcoin/networks/digibyte.py, IDENTIFIER 4B62545B1A631AFE).
-    // V36 conformance = parity with DGB's OWN canonical reference, so the
-    // oracle behaviour IS the spec -- including COIN=1e6 (NOT 1e8) and the
-    // weeks/months + 1 decay count. These are not quirks to "fix" (card #156).
-    // 3-bucket: COMPAT (pre-v36 per-coin baseline, temporary for the crossing;
-    // NOT a v36-native cross-coin standardization).
+    // DGB block subsidy -- EXACT port of DigiByte Core consensus
+    // GetBlockSubsidy(nHeight, consensusParams) (src/validation.cpp), the
+    // authoritative supply curve every DGB mainnet node validates against.
+    // Ground truth is DigiByte Core, NOT the p2pool-dgb-scrypt display oracle:
+    // that oracle's get_subsidy() carried COIN=1e6 and never built a coinbase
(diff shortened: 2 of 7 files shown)


frstrtr:
### Follow-up commit `b6ddf440` — build/KAT proof + floor provenance correction

Pushed on top of the port (fast-forward, no history rewrite). No change to the shipped `subsidy()` value at any height; three tightenings:

**1. Floor provenance corrected (comment-only).** The earlier note said the released 8.22.x consensus clamped a sub-1-DGB reward to `COIN`. That is wrong. `GetBlockSubsidy` clamps it to **0** in BOTH the released consensus (`feature/8.22.0-final`) and the latest release (`v9.26.5`):

```
// Make sure the reward is at least 1 DGB   <- stale comment in Core
if (nSubsidy < COIN) { nSubsidy = 0; }      <- code assigns 0
```

The shipped port already clamps to 0, so this is a documentation fix — but on consensus/money code the provenance must be right. The floor is ~40 years out and never bites at the live tip.

**2. Task tip vector added.** KAT now pins `subsidy(24125001) == 25355810338` alongside `subsidy(24125022)`. Both are in the months=129 monthly-decay band and pay the identical value; 24125001 is the height a live coinbase would be built at, 24125022 is the block whose on-chain coinbase vout is byte-identical.

**3. Fee-unit invariant documented** in `embedded_coinbase_value.hpp`: coinbasevalue is now unit-consistent (subsidy sats + fee sats, bare `+`, no rescale); a stray `x100` there would double-count the fix and overpay.

### Verification

- `conan install` + `cmake --preset conan-release -DBUILD_TESTING=ON`, GCC 13.
- `test_dgb_subsidy`: **4/4 PASS** (Core vectors, live tip, full-domain byte-parity sweep 0..60M vs an independent Core reference, floor-to-zero region).
- `test_dgb_coinbase_value`: **5/5 PASS**.
- **RED-then-GREEN proven**: rebuilding `test_dgb_subsidy` against the pre-fix (1e6) `config_coin.hpp` fails every Core vector 100x low (`72000000000` vs `7200000000000`, …); rebuilding against this branch passes.
- `c2pool-dgb` links clean; `--selftest` OK.
- DOGE lane untouched (`get_doge_block_subsidy`, `COIN=1e8`, unchanged).

Consensus/money — **operator merge tap only, do not self-merge; no deploy without operator go.**

PR #1432, opened 2026-09-01
bip110 M2: mineable BLAKE2b Stratum-v1 (work source + submit-roundtrip KAT) [DRAFT]

frstrtr:
## BIP-110 M2 — make embedded c2pool-bip110 MINEABLE

Base branch: `bip110-m1-embedded-fork-following` (M1, #1431). M1 already proves the daemonless BLAKE2b fork-following header lane. **M2 makes it serve work + accept shares + submit blocks**, via the Stratum-v1 BLAKE2b work-shape that stock Sia BLAKE2b ASICs already speak (the mapping BIP-110 was designed for; matches Knots `primitives/block.cpp` GetHash and the DATUM Gateway reference bridge).

**DRAFT — mining/money path. Do NOT merge.**

### The work-shape (all offsets proven against live fork block 961640)
- `coinb1 = u32(0) || h2 || u32(0) || u32(0)` (44 B), empty `coinb2`, **literally-empty** merkle branch array. `coinb1 || en1(4) || en2(4)` is the 52-byte BLAKE2b root preimage `u32(0)||h2||m_extranonce[16]`.
- `m_extranonce[16] = 8*0x00 || extranonce1(4) || extranonce2(4)`.
- submit `nonce` (8 B hex): low4 → nNonce, high4 → m_nonce2. `ntime` (8 B): low4 → wire time area, high4 → m_nonce3.
- prevhash on the wire is the raw Sia `prevblock_hidden` (NOT a Bitcoin block hash) — a new additive `StratumConfig::raw_prevhash_wire` flag (default false → every other coin byte-identical).

### Change-set (per-coin isolation held; only `src/impl/bip110/*` + `src/c2pool/main_bip110.cpp` + one additive core flag)
- **`stratum/pseudoheader.hpp`** — SSOT fold shared by the KAT and the work source (`compute_h1_h2`, `wire_coinb1`, `rebuild_header_v2`, `parse_h2_from_wire_coinb1`), byte-derived from the M1 `pow.hpp` 164-byte v2 pipeline.
[...]

PR #1436, opened 2026-09-01
bip110 M3: daemonless mempool tx-serving (ingest+price+select+reward-safe xcheck) [DRAFT — DO NOT MERGE]

frstrtr:
## BIP-110 M3 — daemonless mempool tx-serving (good-citizen mandate)

**DO NOT MERGE — money path (mining coinbase/reward). DRAFT for operator review.**
Stacked on `bip110-m2-mineable` (M2 PR #1432).

### Why this is real work
BIP-110 is a Bitcoin fork (`src/impl/bip110/` cloned from `src/impl/btc/coin/`) with **no coin daemon**, so there is no `getblocktemplate` to hand the pool a tx list or per-tx fees — the model the BTC/LTC lanes rely on. M2 therefore mined **coinbase-only EMPTY blocks** while the fork network had pending txs (bad-citizen). M3 does the daemon's job daemonlessly, in **pure Bitcoin-native terms — zero DASH consensus machinery** (no special-tx / creditPool / payees / SortForBlock).

### What it does
- **Ingest**: subscribe `new_tx` + `full_block` on the coin P2P; maintain an own post-anchor UTXO view (`UTXOViewDB`/`Cache`) by connecting every full block.
- **Price (three tiers, fail-closed)**: T1 in-mempool CPFP parent; T2 own UTXO view; T3 parent-tx fetch over `getdata(MSG_WITNESS_TX)` into a **self-authenticating** (txid-recompute) priced-parent side table (`parent_tx_resolver.hpp`). A tx that no tier can price is **EXCLUDED** — never a guessed fee, never a double-spend.
- **Select** (`block_assembler.hpp`, ports Bitcoin-Core `addPackageTxs`): ancestor fee-rate ordering + CPFP under the RDTS **800000 WU** cap. **HARD topological invariant** — a child never enters the template without its in-mempool parent at a lower index (fixes the cloned flat-feerate selector's child-without-parent hazard); a final O(n) pass drops any offending suffix. Every exclusion is **named** (`unpriced-in-mempool-parent` / `weight-cap`).
- **Reward-safety**: `coinbasevalue = subsidy + Σ(provable fees)` EXACTLY; real witness commitment over selected wtxids; txid-merkle over `[coinbase]++txids` and `txcount = 1+N` folded into the 164-byte v2 header. An **independent pre-serve XCHECK** re-derives merkle / fees / coinbase-value / weight / witness-commitment from the frozen bytes and **refuses to serve** on any mismatch. The submit path re-derives the body merkle vs the header before relay and refuses on mismatch.

Sv1 work-shape is unchanged (txs live in the body; only merkle/txcount inside `h1` change). Serving is default-on with `--no-serve-mempool-txs` as the canary escape hatch.

### Honest limitation
`getdata(tx)` answers only from a peer's mempool/relay set, not confirmed history, so a tx spending an **old pre-fork confirmed coin** resolves via neither T2 nor T3 and is excluded (fail-closed = fewer txs, never a guessed fee). Coverage grows with uptime as the UTXO view deepens.

### Live proof (legion64-class run, standalone, no daemon)
`c2pool-bip110 --run --coin-p2p-discover --fork-checkpoint --serve-mempool-txs`
- 20 NODE_BLAKE2B fork peers discovered, handshaked, **header-synced to tip 964347 on the BLAKE2b fork**.
- **5 real mempool txs INGESTED over P2P** (distinct real txids); **T3 parent-fetch fired 19×**.
- All 5 spend pre-fork confirmed coins → correctly **fail-closed EXCLUDED** (coinbase-only, honest limitation above). **0 XCHECK failures, 0 crashes.**
- The priced path (**ingest → price(T2 UTXO) → include → txcount=2 → coinbasevalue=subsidy+fee → merkle**) is proven deterministically with the **real production `Mempool` + `UTXOViewCache` + `assemble_block_txs`** in the new KAT (fee 10000 priced and included).

### KAT — `bip110_m3_serving_kat` (21/21 PASS)
CPFP topological parent-before-child ordering; fail-closed exclusion of an unpriced-parent child and an over-weight tx (named); coinbasevalue conservation to the satoshi; txid-merkle / witness-commitment SSOT; plus the real-production priced-path (add_tx → UTXO price → assemble → txcount=2).

### Isolation
`src/impl/bip110/*` + `src/c2pool/main_bip110.cpp` + additive CMake/KAT only. No btc/ltc/dgb/bch/dash file touched. **Zero DASH code imported.**

### Follow-up (separate, not this PR)
The cloned flat-feerate selector's child-without-parent hazard is latent in the live **btc** lane too — file a btc-lane issue; do not cross-fix here (isolation).


frstrtr:
Landed on master via integration #1439 (bdfed37c) — M2+config_pool+M3 merged together after local build+KAT (3/3) + green CI. Superseded.

PR #1451, opened 2026-09-02
bip110 M3 PR-C2: addrman-backed found-block fan-out broadcaster (DRAFT / do-not-merge)

frstrtr:
## DRAFT — do-not-merge. Folds onto the F1b result (`bip110-m3-pra-sharechain-lane`).

This branch is off `bip110-m3-pra-sharechain-lane`, NOT master. A parallel F1b workflow is editing that lane; this PR-C2 is delivered on a separate branch so it does not race F1b. It will be folded onto F1b's result before any merge — the base is shown as master only so the net PR-C2 delta is reviewable. **Do not merge as-is.**

## What this adds — the addrman-backed BLOCK-FANOUT broadcaster

The python fork's broadcaster does two jobs; bip110 now matches both, all behind `--bip110-sharechain` (default OFF).

### (1) FOUND-BLOCK over coin-core-P2P — the real single-peer money hole
Before: a won block reached exactly ONE NODE_BLAKE2B peer (`coin_node.m_p2p`, replaced on each dial). Now it fans out to the WHOLE reachable NODE_BLAKE2B set.

- `src/impl/bip110/coin/broadcaster.hpp` — `Bip110Broadcaster`: NODE_BLAKE2B slot pool. Real dialing slot factory (bip110's `NodeP2P::submit_block_raw` is real — the DASH #152 leaf was a scaffold whose factory never opened a socket). Addrman-backed `discover()` from `BtcCoinPeerManager` tried set + explicit fork peers; dedupe / cap / primary-exclude / backoff; `prune_dead`; `submit_block_raw_all` fans to every live slot, each guarded.
- `src/impl/bip110/coin/broadcaster_full.hpp` — `Bip110BroadcasterFull::on_block_found`: **ARM A** embedded fan-out + **ARM B** the existing `coin_node.submit_block_with_fallback` (primary relay + optional submitblock RPC). `reached_network()` iff ≥1 arm placed the block — a both-arms-fail block is logged LOUDLY, never silently dropped.

### (2) WON-SHARE over sharechain-P2P — re-spread trigger
The per-peer fan-out already existed (`broadcast_share` → `broadcast_and_mark` over the whole `m_peers` set). The missing robustness was the p2pool `best_share_var.changed.watch(broadcast_share)` re-spread trigger — now wired via `set_on_best_share_changed` so a think()-adopted new best re-spreads. Fires on the IO thread with no tracker lock held; `broadcast_share` takes its own shared try-lock.

### Wiring (`src/c2pool/main_bip110.cpp`), flag-gated
- `stratum_submit_fn` routes through `on_block_found` when the flag is ON; **OFF ⇒ the single `submit_block_with_fallback` call, byte-identical to M2**.
- Broadcaster constructed after `coin_peer_mgr` (the NODE_BLAKE2B addrman), with a 30 s self-rescheduling discovery tick.
- Lock discipline: `on_block_found` touches no sharechain tracker lock (stratum submit thread, reads only the coin-P2P pool).

## Isolation / scope
`git diff origin/master --name-only` for the PR-C2 delta = `src/impl/bip110/coin/broadcaster*.hpp`, `src/impl/bip110/coin/CMakeLists.txt`, `src/c2pool/main_bip110.cpp`, `src/c2pool/bip110_broadcaster_kat.cpp`, `src/c2pool/CMakeLists.txt`, `.github/workflows/build.yml` (KAT allowlist). **No `src/pool` or `src/core` edits** — the coin-P2P push primitive, addrman, sharechain fan-out and cross-coin policy already exist and are only wired.

## Build + tests
- `c2pool-bip110` builds clean.
- New `bip110_broadcaster_kat` (18 assertions) PASS: discover dedupe/cap/primary-exclude, prune_dead+backoff, `submit_block_raw_all` fans to N live slots, dual-arm verdict incl the never-silent-drop case, flag-OFF analog (empty pool ⇒ 0 fan-out).
- `bip110_blake2b_kat`, `bip110_m2_workshape_kat`, `bip110_m3_serving_kat` all PASS (no regression).

## Remaining map → PR-D (persistence + soak)
Per-peer broadcast scoring persistence, addr-gossip growth of the fan-out pool beyond the tried set, an inv→getdata WonBlockRelay handshake if raw-push proves insufficient against strict peers, and the live multi-node soak proving a won block reaches N>1 NODE_BLAKE2B peers.


frstrtr:
Superseded by #1449 (merged 32899fc4). broadcaster.hpp/broadcaster_full.hpp/bip110_broadcaster_kat.cpp all in master (patch-equivalent). Closing.

PR #1452, opened 2026-09-02
bip110 M3: flag-ON found-block witness reconcile (block validity / money) [DRAFT, do-not-merge, folds on M3 stack]

frstrtr:
> **DRAFT — DO NOT MERGE.** Stacks on the M3 PR chain (`bip110-m3-pra-sharechain-lane`, PR #1449 family). This branch is off the lane head, so the diff vs `master` includes the lane commits; only the top commit (`bip110 M3: flag-ON found-block witness reconcile`) belongs to this PR. Fold onto the M3 stack; merge after the lane lands.

## The bug (block validity / money)

On the flag-ON (`--bip110-sharechain`, default **OFF**) path the minted coinbase carries a P2Pool witness-commitment output, but:

1. the coinbase **block-body witness reserved value** was spliced as 32 zeros (`gentx_coinbase.hpp`), and
2. the **commitment** was computed over the `0xff` None-sentinel wtxid root (`2^256-1`, `SegwitDataDefault`).

A flag-ON share that **also meets the coin block target** and is submitted as a **found block** is therefore consensus-invalid. Segwit recomputes the coinbase witness commitment as `SHA256d(witness_merkle_root || reserved)` over the **real** witness root — coinbase-only ⇒ `merkle([0]) == ZERO` — which matched *neither* the `0xff` root *nor* the 0-reserved value ⇒ `bad-witness-merkle-match`, block rejected, **a won block lost**. This is not a share-acceptance blocker (peers verify the non-witness txid/merkle), which is why it is a separate must-fix.

## Fix — reconcile both axes to the python v36 reference

`p2pool-merged-v36 data.py:1090/1092-1094`: a segwit-active **coinbase-only** mined share uses `wtxid_merkle_root = merkle_hash([0]) = 0` and `witness_reserved_value = '[P2Pool]'*4`. Both are needed; fixing only the reserved value leaves the block invalid (the commitment would still be over `0xff`).

| file | change |
|---|---|
| `coin/gentx_coinbase.hpp` | `assemble_gentx_coinbase` takes an optional `witness_reserved_value` (empty ⇒ 32 zeros, byte-identical); block-body witness stack splices it |
| `stratum/work_source.cpp` | flag-ON commitment over the **real** root `witness_merkle_root({}) == ZERO` (not the `0xff` sentinel); assemble call passes `'[P2Pool]'*4` iff `pplns_fn_` (same discriminator that chose the P2Pool commitment). OFF/M2 passes empty ⇒ 32-zero reserved |
| `pool/share_check.hpp` | `create_local_share` stores a segwit-active coinbase-only share's `wtxid_merkle_root` as the real root `ZERO` — so mint==verify **and** the on-chain commitment is consensus-valid. Sentinel retained only for segwit-inactive (never on this v36-genesis chain) |
| `c2pool/main_bip110.cpp` | `ref_hash_fn` serializes the same `ZERO` root so the ref a peer recomputes off the stored share matches the coinbase |

The reconciled `ZERO` root **is** the python v36 wire value for a mined coinbase-only share (a present `segwit_data` with root 0), so this **converges** with a python-fork peer rather than diverging; the `0xff` sentinel is only the `PossiblyNoneType` wire encoding of "no segwit_data", never a real root.

## Proof they now agree (KAT)

`bip110_multiminer_pplns_kat` extended with `[WR]` assertions on the assembled flag-ON coinbase:

- `[WR]` block-body witness reserved value `== '[P2Pool]'*4`
- `[WR]` `SHA256d(witness_root(ZERO) || reserved_from_block) == coinbase commitment` — **a found block passes segwit CheckWitnessMalleation**
- `[WR-OFF]` no-reserved-arg path stays 32-zero and `SHA256d(ZERO || 0*32) == M2 commitment` (OFF path valid & unchanged)

`pool_mint_kat` / `extend_mint_kat` updated to assert the `ZERO` coinbase-only root.

## Build / test

- `c2pool-bip110`: **green**
- all 9 bip110 tests: **green** (`pool_mint_kat`, `extend_mint_kat`, `multiminer_pplns_kat` incl. new `[WR]`/`[WR-OFF]`, `pool_node_compile`, `pool_identity_kat`, `sharechain_compile`, `blake2b_kat`, `m2_workshape_kat`, `m3_serving_kat`)
- multiminer `[X] generate_share_transaction == minted PPLNS coinbase` still passes (mint==verify preserved)

## Constraints honored

`--bip110-sharechain` default **OFF**; OFF/M2 path **byte-identical** (empty reserved ⇒ 32 zeros; M2 commitment untouched). No `params.hpp` change. Per-coin isolation: changes confined to `src/impl/bip110/*` + a 2-line `ref_hash_fn` segwit-root change in `main_bip110.cpp` (unavoidable — it bakes the same root into the ref commitment).

### DIFF
--- a/src/c2pool/CMakeLists.txt
+++ b/src/c2pool/CMakeLists.txt
@@ -389,6 +389,14 @@ target_link_libraries(c2pool-dash
 add_executable(c2pool-bip110
     main_bip110.cpp
     ${CMAKE_SOURCE_DIR}/src/impl/bip110/stratum/work_source.cpp
+    # M3 sharechain MINT lane (behind --bip110-sharechain; default OFF): the
+    # namespace-copied sharechain node + single-protocol v36-genesis bridge +
+    # c2pool protocol handler. main_bip110 references bip110::pool::Node (ctor,
+    # listen, broadcast_share, notify_local_share, register_template_txs,
+    # start_outbound_connections) which are DEFINED in these TUs — same set the
+    # bip110_pool_node_compile KAT links. Dormant unless the flag is passed.
+    ${CMAKE_SOURCE_DIR}/src/impl/bip110/pool/node.cpp
+    ${CMAKE_SOURCE_DIR}/src/impl/bip110/pool/protocol_actual.cpp
     ${CMAKE_SOURCE_DIR}/src/impl/bip110/crypto/blake2b.c
 )
 c2pool_wire_version(c2pool-bip110)
@@ -410,6 +418,8 @@ target_link_libraries(c2pool-bip110
     yaml-cpp::yaml-cpp
     nlohmann_json::nlohmann_json
     ${Boost_LIBRARIES}
+    ${SECP256K1_LIBRARIES}
+    Threads::Threads
 )
 
 # Dashboard assets: the shared coin-generic WebServer serves the web-static/
--- a/src/c2pool/main_bip110.cpp
+++ b/src/c2pool/main_bip110.cpp
@@ -555,11 +555,13 @@ int run_embedded(bool coin_p2p_discover,
         //   • share_version = desired_version = 36 (no AutoRatchet — v36 always)
         //   • donation u16 = 66 (WIRE-GENESIS freeze, 0.1%; NOT btc's 50)
         //   • no merged mining (merged_addresses / merged_coinbase_info empty)
-        //   • segwit_data = SegwitDataDefault::get() none-sentinel (2^256-1 root):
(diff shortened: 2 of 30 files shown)


frstrtr:
Superseded by #1449 (merged 32899fc4). flag-ON found-block witness-reconcile absorbed into master share_check.hpp/work_source.cpp/gentx_coinbase.hpp (patch-equivalent). Closing.

PR #1457, opened 2026-09-03
sharechain: port restart-reorg supersede (Class-A self-fork) + periodic think tick to dash/btc/dgb

frstrtr:
## Summary

An audit of the shared C++ sharechain engine (all coin lanes) for the two
kr1z1s-class convergence defects found the LTC-only restart-reorg supersede fix
(#1357 + #1405) **unported and live** on the btc/dgb/bch/dash lanes:

- **Class A — self-fork / incumbent structural privilege (HIGH).** A warm-restarted
  node that loads its persisted, fully-verified sharechain then peers with a
  higher-work network STICKS on the old head and self-sustains its own minority
  fork. `TailScore` compares `chain_len` FIRST and `score()` short-circuits to
  `{verified_height,0}` below `CHAIN_LENGTH`, so a warm-loaded full-CL verified
  tail beats any higher-*received*-work challenger whose verified height is still
  short — and `think()` never backfills the unrooted challenger to the ~2*CL depth
  a CHAIN_LENGTH-tall verified tail needs. This is the C++ analog of the python
  fork's majority-escape (different mechanism, same outcome: own-share orphans).
- **Class B — single-peer wedge (liveness, btc/dgb/bch only).** `think()`/`clean_tracker()`
  fire ONLY on inbound share arrival because these three mains have no periodic
  think tick, so a node with no live share flow (all peers silent/hostile, no
  local miners) never re-attempts a timed-out bootstrap download. (The consumer
  request path already fails over across random peers with a 15s timeout and
  per-cycle reset — it is otherwise immune to a single hostile peer.)

This PR ports the complete, proven LTC stack per-lane (per-coin isolation
invariant — each lane's `share_tracker.hpp`/`node.{hpp,cpp}` edited separately):

- `compute_supersede_hint()` — fires only for a **genuine** fork (`is_child_of`
  guard) whose **RECEIVED** work is **strictly greater** than the incumbent
  best's **VERIFIED** work and whose verified height is still `< CHAIN_LENGTH`.
- `think()` — deepens that segment's backfill to `2*CL+10`, spends a **bounded**
  elevated verify budget (400/tick) through the ordinary `attempt_verify` loop
  (never bypasses verification, never scores unverified work), and bounds the
  per-cycle exclusive-lock hold with `THINK_MAX_WALL_MS` (2s).
- `prioritize_challenger_heads()` — gives the challenger Phase-2 budget priority
  (stable_partition; no-op when inactive → healthy-node order byte-identical).
- `gate_supersede_convergence()` — denylists an unconvergeable segment (missing
  parent unobtainable) after `SUPERSEDE_STALL_LIMIT`=20 stalled cycles, TTL 60m,
  to break the elevated-verify treadmill and re-enable GC.
- `clean_tracker()` — exempts the converging challenger from stale-head eating /
  tail-dropping while the hint is active.
- **FIX-2** (btc/dgb): 5s `clean_tracker()` safety-net tick in the main loop
  (mirrors `main_ltc.cpp`), closing Class-B and making `clean_tracker()`
  (previously never called on these lanes) actually run. DASH already has its tick.

Guarantee: a node **always converges to the best reachable valid sharechain**
(strictly-higher-work-only, verified-only selection preserved, deactivates once
full-length verified — no oscillation) and can no longer self-sustain a minority
fork or be wedged by a silent-peer bootstrap stall.

## Lane status

| lane | Class-A (FIX-1) | FIX-2 tick | notes |
|------|-----------------|------------|-------|
| ltc  | already fixed (#1357/#1405) | already has 5s tick | reference |
| dash | **this PR** | already has 15s tick | live money on the shared hotel sharechain |
| btc  | **this PR** | **this PR** | btc.voidbind live |
| dgb  | **this PR** | **this PR** | dgb.voidbind live |
| bch  | **NOT in this PR** | **NOT in this PR** | tracker is structurally divergent (see below) |
| bip110 | n/a | n/a | no sharechain node yet (M3); port the btc-cloned tracker as part of M3 |

## Not yet done (follow-up, called out honestly)

- **bch lane** — bch's `share_tracker.hpp` is structurally divergent from the
  ltc/btc/dgb template (no matching Phase-2 head-sort anchor; different struct
  layout). It needs the same port but adapted by hand, not mechanically. Deferred
  to avoid shipping unverified divergent consensus code.
- **Per-lane KATs** — the shared algorithm (`compute_supersede_hint` /
  `prioritize_challenger_heads`) is already covered by
  `src/impl/ltc/test/restart_reorg_supersede_test.cpp`; the per-lane clones (which
  need each lane's share-construction harness) are the immediate next step, to be
  authored once CI confirms the ports compile so they can be iterated.
- **Local build** — deliberately NOT run: this host is disk-constrained (96%) and
  shared with live infra. CI is the authoritative multi-lane build/test gate for
  this PR. Static verification done locally: symbol-coherence greps across all
  three lanes, brace balance, attribution scrub, structural review of the
  hand-edited DASH lane.

DRAFT pending CI (compile proof) + adversarial review.

### DIFF
--- a/src/c2pool/main_btc.cpp
+++ b/src/c2pool/main_btc.cpp
@@ -3203,6 +3203,30 @@ int main(int argc, char* argv[])
         LOG_INFO << "[BTC-POOL] dashboard disabled (no --http bind given).";
     }
 
+    // Restart-reorg / liveness safety-net: periodic clean_tracker() tick (5s).
+    // clean_tracker() runs think() inline (advancing verification + the restart-
+    // reorg supersede convergence) and prunes stale heads/tails. Without this tick
+    // think() fires ONLY on inbound share arrival, so a node with no live share
+    // flow (all peers silent/hostile, no local miners) never re-attempts a
+    // timed-out bootstrap download and never advances a warm-restart reorg — the
+    // consumer-side single-peer-wedge / stuck-tip liveness gap. Mirrors
+    // main_ltc.cpp:6682. Self-rescheduling on the run-loop ioc; stops when the
+    // timer is cancelled at shutdown (ec set).
+    auto think_timer = std::make_shared<boost::asio::steady_timer>(ioc);
+    std::function<void(boost::system::error_code)> think_tick;
+    if (p2p_node) {
+        think_tick = [&, think_timer](boost::system::error_code ec) {
+            if (ec) return;
+            think_timer->expires_after(std::chrono::seconds(5));
+            think_timer->async_wait(think_tick);
+            try { p2p_node->clean_tracker(); }
+            catch (const std::exception& e) { LOG_ERROR << "[CLEAN-TRACKER] error: " << e.what(); }
+            catch (...) { LOG_ERROR << "[CLEAN-TRACKER] unknown error"; }
+        };
+        think_timer->expires_after(std::chrono::seconds(5));
+        think_timer->async_wait(think_tick);
+    }
+
     LOG_INFO << "[BTC] io_context running. Ctrl-C to stop.";
 
     // Diagnostic [IOC-LAT]: when a single run_for slice takes far longer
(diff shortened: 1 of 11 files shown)


frstrtr:
Stale/superseded: dash SupersedeHint already in master; btc/dgb/bch restart-reorg reworked (improved, non-draft) in #1459. Closing in favor of #1459.

Issue #859, opened 2026-07-25
ltc: average_attempts_to_target UB above 2^64 — silently stops minting (same latent bug fixed for DASH in #858)


frstrtr:
Found by the sweep KAT while wiring the pool-share cap for DASH (PR #858). **The DASH copy is fixed there; LTC's inlined copy is untouched and this is a LIVE PRODUCTION lane.**

## The bug
`average_attempts_to_target` does `static_cast<uint64_t>(avg_attempts)` on a `double`. Above 2^64 that is **undefined behaviour**, and on x86-64 it yields **0** — which then produces a **"Division by zero"** thrown out of the producer-job path.

`work_source.cpp` catches it and degrades to the non-producer coinbase, so the symptom is not a crash: **minting silently stops.** The pool keeps serving work and looks healthy while producing no shares.

## Why it becomes reachable
Cap 1 of the pool-share cap scales the measured hash rate by ~1198x (`local_hash_rate * SHARE_PERIOD / 0.0167`), so the effective boundary lands around ~1.5e16 H/s of *measured* rate. Without the cap the boundary is far away; with it, it is merely improbable rather than impossible — and a bogus/stale rate reading could reach it.

## Fix
Saturate at `UINT64_MAX` instead of casting an out-of-range double. This is what PR #858 does for the DASH path.

**LTC site:** `src/c2pool/main_ltc.cpp:4538` (the inlined copy of the same computation).

## Why it matters here
LTC+DOGE is the shipping production line. A silent mint stop is the worst failure shape — no crash, no alert, dashboards keep reporting a healthy node. Worth fixing even though the trigger is unlikely, because the failure is invisible.

Related: #858 (DASH pool-share cap, where this was found and fixed for DASH).

frstrtr:
Fixed on master — closing.

Merged change: PR #1308 "ltc(#859): saturate average_attempts_to_target narrowing (UB / silent mint-stop)" (branch ltc-doge/ltc-859-ata-u64-overflow), merged 2026-08-22 as commit 1621386, which is an ancestor of current master (tip 2e0d9569, the dashboard persistence roll for #159).

Live-path evidence on master:

[...]

PR #1483, opened 2026-09-04
v37/W2: RDWR receipt ingestion (admission-first, carrier-position credit) [DRAFT]

frstrtr:
## Track A2 · W2 — RDWR receipt ingestion (DRAFT)

The receipt-intake path of the A2 C++ bring-up: takes a candidate share (the
*carrier*) plus the up-to-`R_MAX` RDWR receipts it embeds, validates each
statelessly against committed chain data + one small dedup set, and turns each
**accepted** work event into exactly one weight-credit push through the W1 lane
executor via the **W0 `V37Engine::submit_tracked`** seam. The C++ `Lane` never
sees a receipt object — by design it sees only `(descriptor, w_raw, flags)`.
All receipt semantics live **before** the push, in the share-check position.

Consumer-tree only (`src/c2pool/v37/`), alongside W0. The header-only consensus
module (`src/sharechain/v37/`) is untouched **except** the one owed
`L0F_RECEIPT` constant (below).

### The ruling this implements — W2-F-A option (a)

Operator ruling (2026-09-04): **carrier-position credit for V37.0**. Each
accepted receipt's work is credited at the carrier/arrival position — exactly
what `Lane::push` does natively (it always appends at `next_pos`; it has **no**
bin/position argument). **No origin-bin positioned insert.** So a receipt push
is an ordinary push of `work(T_origin)` at the carrier position, distinguished
only by the annotation-only `L0F_RECEIPT` bit. The origin-bin T-OQ1 refinement
is demoted to V37.x (there is no positioned-insert surface to call).

### What is here

- **Receipt/carrier model + PoW** (`w2_receipt.hpp`): real `sha256d` mini-PoW
  over the RDWR 4-tuple preimage (chain_id, payout-descriptor identity, recent
  mainchain block hash, prev on-chain share hash), so every binding check is
  stateless against committed bytes and tampering really breaks the PoW.
- **Stateless per-receipt validation** (`w2_admission.hpp`) in consensus order
  (spec §3.1): PoW → **R-1 origin-target equality** (easy-at-hard / hard-at-easy
  both rejected) → identity (self-carriage) → prev-own → chain → **N_CTX
  expiry** → **dedup**. Any failure ignores that receipt; the carrier still
  stands. Wrong pool/chain/payee and bad prev-own reject; a tampered PoW rejects.
- **Single dedup window** within `N_CTX`, keyed on lane **incarnation**
  (W2-F-D / F2 — version ABAs across `RemoveLane`→`AddLane`), pruned on clock
  advance, `R_MAX=4` per carrier with whole-carrier reject when exceeded.
- **Fixed push sequence** (spec §4.2): carrier first, then accepted receipts in
  wire order, each self-carried under the carrier's descriptor with
  `flags | L0F_RECEIPT`. Every push advances `next_pos` by 1 (W2-F-E: "receipts
  never advance the clock" is the *bin* clock, not the lane position).
- **`w_raw = work(target)`** with a **saturating** `uint288`→`u64` narrowing
  (KP-F2): clamp to `UINT64_MAX`, never silently truncate/wrap — the discipline
  the #859 fix established. Cross-checked against exact `2^256/(T+1)` arithmetic.

### 🔶 Consensus-surface flags for the integrator (do NOT merge without these)

1. **`L0F_RECEIPT = 0x04`** added to `v37_lane.hpp` (the one free low bit).
   Annotation-only and **digest-neutral** by construction — flags never enter
   `build_leaves()` and never affect weights — guarded by two `static_assert`s
   (value == 0x04; no collision with the assigned bits). Proven digest-neutral
   in C++ (KA-8) and by the **unchanged** `v37_test` lane golden. It touches the
   consensus header, so **finalizing it needs an integrator digest-neutral tap
   before MERGE** (standing no-self-merge rule). It does **not** block the impl
   or the tests.
2. **`N_CTX` (=2 bins) / `R_MAX` (=4)** are kept as **W2-layer consensus
   params**, NOT baked into the `LaneParams` header leaf. KP-F1: that leaf is
   digest-committed (it hashes geometry), so adding fields there is **NOT**
   digest-neutral (unlike `L0F_RECEIPT`). Whether they must become digest-
   committed `LaneParams` is an **open integrator question** — flagged, not
   silently decided.

### What is NOT here

- **No relay** — carrier relay over the p2p layer is **W3**.
- **No settlement / owed ledger** — W2 credits ledger #1 (in-window weight) only;
  owed balances, K_fair, coinbase, SETTLED are **W4**. No address monitoring.
- **No decayed-value pins** — every decayed leg is OI-7-blocked; W2 asserts
  raw-work / identity / ordering / dedup legs only.
- No positioned-insert surface (KR-1 branch (b)).

### Test evidence & reproduced golden

`v37_w2_ingestion_test` (stdlib-only, wired into `build.yml` **both** legs —
x86_64 and ASan/UBSan — plus `add_test`, to avoid the #1467/#1477 NOT_BUILT
hollow-green trap):

- **Ruling-agnostic KA-1..8** + **KR-1 branch (a)** (freshness-quirk Δ0/Δ1/Δ2 →
  identical credit; receipt push ≡ ordinary push of the same `w_raw` at the same
  position, shown intra-engine and OI-7-agnostically).
- **Golden reproduced bit-for-bit** — the proto RDWR admission reference
  `proto/w2-rdwr-kat` stamp
  `7ac46235e65022ed69afe426f5bf5f8767207ec201050b10e36b7106658d8b51`: push-
  sequence digests `d7adff5d…` (KA-1), `92df6bd2…` (KA-2 single, equal across
  all dup modes), `90679579…` (KA-4 receipt-absent), raw sums (768 / 1024),
  window sizes `[3,4,5,6,5,5,5,5,5,5,5]`, and the KR-1 age-delta vector
  `[0,0,1,2]`, embedded as literals.
- Every accepted push driven **end-to-end** through `V37Engine::submit_tracked`
  (`applied()`, interned `MinerId` self-carriage, `next_pos`, `raw_work_in_span`,
  intra-engine `digest`), plus the executor `SubmitStatus` handling
  (`RejectedZeroWork` / `RejectedInvalidDescriptor` / `RejectedUnknownChain`,
  stopped-engine future throw treated as carrier-reject, never a crash).

**Result:** 128 checks, 0 failures; green under ASan/UBSan; `v37_scaffold_test`
(6606), `v37_executor_test` (10551), `v37_test` (100576) unaffected.

**DRAFT — do not merge, no self-merge.** Blocking gate: integrator digest-
neutral tap on `L0F_RECEIPT`, and a ruling on the `N_CTX`/`R_MAX` digest
question, before this leaves draft.


frstrtr:
Ratification record (constant freeze), recorded post-merge for a citable reference:

- Disposition B adopted: N_CTX = 2, R_MAX = 4 fixed as V37.0 constants.
- Investigation verdict: SOUND. Digest unchanged — golden d659c801 unmodified; #1483 required no digest change.
- Change scope: consumer-tree static_asserts only (compile-time check).
- This comment supplies the review-record citation that was missing at merge time (merged 2026-09-05T02:58:04Z, merge 2c9f2be0, head b60b1310).

PR #1459, opened 2026-09-03
btc/dgb/bch: restart-reorg SupersedeHint stack + periodic think/clean tick [DRAFT]

frstrtr:
## What

Ports the LTC-proven restart-reorg **SupersedeHint** stack (#1357 + #1405) to the **btc**, **dgb**, and **bch** share trackers, and adds the periodic **think/clean tick** these three lanes were missing. Companion to the DASH PR (#1458) — same defect class, reviewable as a 3-lane mirror.

C++ analog of the kr1z1s Python self-sustained-minority-fork desync; confirmed present on all three lanes by the cross-lane audit.

## The defect (Class-A, all three lanes)

`score()` short-circuits to `{verified_height, 0}` below `CHAIN_LENGTH` and `TailScore` compares `chain_len` FIRST, so a warm, full-CL, fully-verified incumbent structurally beats **any** higher-work challenger whose verified height is still `< CL` — and the challenger can never verify to CL because Phase-2 never backfilled its unrooted segment to the ~`2*CL` depth a CL-tall verified tail needs. A node with local hashrate then self-sustains a minority fork.

## The fix (per lane — mirror of LTC; no reorg path, no weakening of anti-Sybil depth)

- **`share_tracker.hpp`**: `SupersedeHint`; `compute_supersede_hint()` (strictly-greater received work + `is_child_of` genuine-fork + sub-CL guardrails); `prioritize_challenger_heads()`; `think()` 5th **defaulted** param, bounded elevated verify pool, challenger `2*CL+10` backfill, `THINK_MAX_WALL_MS=2000` bound (bundled — the elevated pool is unsafe without it; scrypt on dgb makes it essential).
- **`node.hpp`/`node.cpp`**: `m_supersede_hint` + stall-denylist (`SUPERSEDE_STALL_LIMIT=20`, TTL 60m) + `gate_supersede_convergence()`; `run_think()` + `clean_tracker()` (Step-1 / Guard-1b / Step-3 / Step-4) compute+pass the hint and exempt the converging challenger from GC.

## The A2 tick (NEW behaviour on btc/dgb/bch)

[...]

PR #1488, opened 2026-09-04
btc: restart-reorg SupersedeHint stack + periodic think/clean tick (slice of #1459)

frstrtr:
BTC-only slice of the btc/dgb/bch restart-reorg SupersedeHint bundle (#1459), decoupled per integrator ruling (a) so it can land on its own review cadence.

Scope — btc-only files, no dgb/bch/shared hunks:
- src/impl/btc/node.cpp / node.hpp
- src/impl/btc/share_tracker.hpp
- src/c2pool/main_btc.cpp
- src/impl/btc/test/restart_reorg_supersede_test.cpp (+ CMakeLists registration)

What it does: rewires the BTC think() call sites to the 5-arg SupersedeHint form so a restart-time reorg supersedes the stale chain head instead of livelocking; adds the periodic think/clean tick and a KAT proving the supersede path (preservation semantics preserved).

do-not-merge is applied: HOLD until the GLM-5.2 adversarial verdict on the SupersedeHint reshape lands (bg stream in progress). Non-draft so review can proceed in parallel.

Head SHA f2780f4a (GPG-signed). Bundle #1459 retains the dgb/bch/build.yml hunks.


frstrtr:
Clearing do-not-merge: the hold condition was a GLM-5.2 adversarial verdict on the SupersedeHint reshape. The GLM tank (192.168.86.45) is off-net — probed 2026-09-06 08:52 UTC, no route on :8095 and :8000 (curl code 000), so that tier cannot produce a verdict. Per the standing peer-steward substitution rule, dgb-scrypt-steward has cross-read this head (a4c2fab4) and returned SHIP, covering (a) the 4->5-arg think() reshape being semantic-preserving at all three call sites with an inactive hint, (b) the KAT registering for real on both the linux and ASAN/UBSAN legs under btc_share_test with ctest --no-tests=error, and (c) no btc divergence from the already-merged dgb/bch mirrors #1490/#1491, which landed on the same reviewer verdict. Per-coin isolation clean (src/impl/btc + main_btc.cpp + btc/test only). 39/39 green. Merging.

PR #1551, opened 2026-09-09
[DRAFT] v37/xmr option B: canonical K_fair payees (W4Propose) + section-13 state-root commitment

frstrtr:
## [DRAFT] Multi-node XMR settlement: canonical K_fair payees + section-13 state-root commitment

Wires the **two operator rulings of 2026-09-10** into the X9 option-B settlement coinbase. This is multi-node **consensus** wiring, so it ships as a DRAFT: the operator merges and activates.

**Scope fence.** Consumer + impl trees only — `src/c2pool/v37/xmr/*`, `src/impl/xmr/template/*`, plus the new KAT and its CI wiring. **`src/sharechain/v37` canon is untouched, and so is the W4 `owed_digest` fold.** `w5_coinbase.hpp` is only *read* (the `StateCommitment` builder); `w4_settlement.hpp` is only *called* (`propose_coinbase`).

---

### Ruling 1 — K_fair source is `W4Propose`, and it is re-derived at every reward

The XMR coinbase selects payees by the ONE ratified K_fair rule: `OwedLedger::propose_coinbase` — oldest-owed-first (`first_eligible` ASC then identity ASC), `take = min(EffectiveOwed, remaining)`, **CARRY** a sub-`h_min` take, count-capped at C. That is whitepaper erratum **E-1**. Every node derives the SAME output set from the same ledger.

`W4Propose` was already the default — but the set was projected **once**, at the first reward hint, and then re-cut by X6 arithmetic at every other reward the assembler's count-invariance fixpoint visited. **That is the real gap this closes.**

| change | where |
|---|---|
| the projection moves out of `build()` into a free function `project_w4_owed()` — one body, one order, no second implementation | `xmr_o2_settlement_source.hpp` |
| `AssemblyInputs::reproject_owed`, called at every reward the fixpoint visits **and** in `split_reward`'s accept-a-different-reward branch, *before* X6 allocates | `xmr_block_assembly.hpp` |
| on adoption the re-projected rows are copied into the FINAL `CoinbaseInputs`, so the FOUND record and any peer ACCEPT re-derivation carry the canonical set for the reward actually paid | `xmr_block_assembly.hpp` |
| the callback receives the **RESOLVED weight-aware cap**, closing a real cap disagreement (proposal cut at the config ceiling vs X6 truncating at the weight-aware cap ⇒ the emitted tail was not the proposal at any budget) | `xmr_block_assembly.hpp` |
| `KFairSource::X6Allocate` REFUSED fail-closed at both the config gate and the build gate, ruling named in the message | fixture + source |
| the daemon binds the hook through one shared helper `make_reproject_owed()` — the same one the KATs bind | fixture + provider |

`same_set()` is **unchanged** and still governs adoption: a re-projection that changes the SET is refused, the reward is handed back, and the assembler rebuilds at it. `max_passes` and the cycle guard still fail closed with a named reason surfaced through `provider.last_error()`. **An empty `reproject_owed` is byte-identical to the previous release**, so every impl-only caller is unaffected (pinned by KC-4).

### Ruling 2 — the MM commitment binds the section-13 StateCommitment root

The MM-root leaf in the coinbase `tx_extra` `0x03` now commits the FULL whitepaper section-13 `StateCommitment` Merkle root — the same root `w5_coinbase.hpp` builds (summary leaf `V37S` + per-key `V37E` balance leaves) — instead of the narrower `owed_digest`.

* **Strict subsumption**: the section-13 summary leaf already *contains* `owed_digest`, so this strengthens the previous binding rather than replacing it.
* We commit the **root VALUE the engine computes**, never a re-derivation from pinned constants — so the wiring is robust to any later change in what section 13 hashes.
* `LaneCommitmentSource::StateRoot` is the new default; anything else is refused unless `allow_nonruled_local_only` is set for a single-pool experiment.
* New operator flags `--lane-commitment <state-root|owed-digest|explicit>` and `--lane-commitment-local-only`. The serve banner now prints the resolved K_fair source **and the 32-byte committed root**, so a node is auditable straight from its log.
* `lane_commitment` also seeds the deterministic tx secret key `r`, so `r`, `R`, every `P_i` and view tag move: **every pinned option-B byte golden was regenerated.** Expected, not a regression (pinned by KS-5).

---

### New KATs — `v37_xmr_o2_settlement_kat`, 62 checks, GREEN

The first test executable over the *consumer* option-B settlement path (the impl-side self-check covers only the halves that need no `OwedLedger`).

* **KW-1** the payee set equals an **INDEPENDENT** canonical K_fair walk written in the KAT, at the FINAL reward and the RESOLVED cap, in a penalty-zone case where the tail payee's take falls below `h_min`. With the hook unwired the same lane emits a sub-floor dust output the canonical rule forbids — **this is the regression pin.**
* **KW-2** `X6Allocate` refused at both gates. **KW-3** canonical order survives the template (owed ++ fixed ++ sink last, configured sink identity). **KW-4** payout-map honesty.
* **KS-1..KS-5** the section-13 root is what the block commits; a verifier holding the same ledger recovers it; a **one-piconero** ledger difference moves it; the section-13 **O(log n) balance proof verifies against exactly the committed root** (`Lane::verify_proof`) — proven end to end for the first time; and the `r`/`R`/`P_i` movement is pinned.
* **KC-1..KC-4** the fixpoint under a set that changes size (2 passes), cap agreement, fail-closed refusals with named causes, and byte-identical back-compat.
* fixture selftest extended: **S4** now pins the state-root default, **S7** the resolver, **S8** both RULED refusals.

### CI

`-DXMR_BUILD_OPTION_B_KATS=ON` on **both** the Linux x86_64 leg and the ASan+UBSan leg, with all three option-B targets in both `--target` lists. Locally green on both configurations (gcc-15):

| target | plain | ASan+UBSan |
|---|---|---|
| `v37_xmr_o2_settlement_kat` | 62/62 | 62/62 |
| `xmr_block_assembly_kat` | 192/192 | 192/192 |
| `xmr_template_primitives_kat` | ALL PASS | ALL PASS |

**Bug found by turning that option on**: `xmr_template_primitives_kat` carried a wrong hand-written golden — `LEB128(600000000000)` is `80e0a596bb11`, not `80a094a58b11` (which decodes to 587146268672). The implementation was always right (`xmr_block_assembly_kat`'s K0 pins `writeVarint` against the vendored `tools::write_varint`), but that KAT was never in a CI target list, so the bad constant went unseen. Fixed with the derivation in a comment.

---

### FOUND → FINALIZE re-proven on an ISOLATED private regtest

`monerod --regtest --offline --fixed-difficulty 20000` with its **own** data dir and **own** ports (18189 rpc / 18188 p2p / 18190 zmq), pool stratum on 18189-adjacent port 3433 — every endpoint on `127.0.0.1` of the build host, sharing nothing with any other job. **The live stagenet node was never touched** (verified still up and running, untouched, after the run). Real `xmrig` at `rx/0`, 4 threads, against the pool's own stratum port; RandomX runtime verify + the exact 128-bit network gate on the submit path.

```
found registered=30  settled=25  orphaned=2  refused=0  pending=3
gate calls=30 accept=30 reject=0 | submit calls=30 ok=30 rejected=0
```

Offline verification of every canonical block, by an **independent** pure-Python Keccak-256 + block-blob parser (not the code under test):

```
blocks=28 verified=28 failed=0  (block ids matching the pool's FOUND set: 28)
RESULT: GREEN
```

Per block it asserts:
1. **payees == the canonical K_fair set** — the owed payee first at its full `EffectiveOwed`, then the exact-sum residual sink; `sum(vout) == reward` on every block;
2. **section-13 root committed** — `tx_extra` `0x03` root `f091f3734c82c659…` equals `keccak256("c2pool-v37-xmr-mm-leaf-v1" || chain_id_le32 || state_root)` for the lane's section-13 root `6b84bf170e886678ae8c648783c4ca6346b2570ee8079600778cf399b2729d57`, recomputed independently;
3. **block id == the pool's FOUND bid** — all 28 canonical ids matched (the 2 non-canonical are same-height races, correctly ORPHANED pre-SETTLED per O3.5);
4. **FINALIZE at `H_b + D_conf`** — e.g. `h=1 SETTLED at bin_height=4`, `h=2 → 5`, with `D_conf=3`, for 26 settled blocks.

The previous option-B run's accept rate (19/19) is matched and exceeded: **30/30 submits accepted, 0 rejected.**

---

## R-7 reconciliation + the count-invariance guard (added 2026-09-10)

Two must-fix defects found by an adversarial survey of this branch. Both are **consumer/impl-tree only** — `src/sharechain/v37` and the W4 `owed_digest` fold have a **0-line diff vs master** (`git diff --stat origin/master -- src/sharechain/v37/ src/c2pool/v37/w4_settlement.hpp src/c2pool/v37/w5_coinbase.hpp` is empty).

### Defect R-7 — two disjoint ledgers, and a fictional FOUND record

The daemon held **two** `OwedLedger` objects. The coinbase was built from the provider's settlement ledger; FINALIZE booked into `XmrNode`'s store-backed one. On top of that, `FinalizeConnect` synthesised `credit == payout == { operator --payee-* identity : FULL block reward }` — a payee that under option B is **not a coinbase payee at all** (the coinbase pays the K_fair owed rows plus the residual sink).

Two empirically confirmed consequences:

**(a) `EffectiveOwed` went negative.** `EffectiveOwed = finalW − Σ_pending payout`, so with the reported run's own `pending=3` it is `−3·reward` — the observed `−105,547,781,241,725`.

**(b) No-double-pay violation** (the harder half, not stated in the original PR body). The settlement ledger was never decremented, so `propose_coinbase` re-proposed the **same owed row at its full `EffectiveOwed` on every block**. This PR's own offline verifier says exactly that — *"the owed payee first at its full EffectiveOwed"* on all 28 verified blocks. That is 28 payments of a balance the paper (§9 / OI-W4-5 credit-at-finality) extinguishes **once**.

### The fix

Against the canonical contract in `w4_settlement.hpp` (`:446-458` on_block_found, `:485-500` on_block_finalized, `:531-542` effective_owed) — **unmodified** here.

1. **ONE LEDGER.** `XmrOwedFixture` gains a non-owning ctor over an external `OwedLedger`; `main_v37_xmr.cpp` hands the provider `XmrNode::ledger()` — the durable, `RecoveryDriver`-replayed, `XmrFinalizeDriver`-driven one. Block *N+1*'s proposal now sees block *N*'s pending payout already deducted, so a row in flight cannot be re-proposed. The owning ctor is kept for the KATs.
2. **PAYOUT = THE EMITTED `Role::Owed` SET** — never the proposal, never the reward. Read from `AssembledTemplate::outputs()` filtered to `CoinbaseOutput::Role::Owed`, and captured at the **submit seam** (`submit::BlockCandidate::coinbase_owed` → `submit::FoundBlockEvent` → `o2::FoundBlockEvent`) so the map belongs to the very snapshot whose bytes went to monerod and cannot be lost to provider-ring eviction. `Role::Fixed` / `Role::Sink` are **excluded**: those identities were never credited, so booking them would drive their `finalW` permanently negative. Reading the *emitted* set rather than re-running `project_w4_owed` is load-bearing: X6's owed pass can legitimately emit **fewer** rows than W4 proposed (it BREAKs on a budget-truncated sub-`h_min` tail and the sink absorbs it, `xmr_coinbase.cpp:135-139`; W4 CARRYs and keeps scanning, `w4_settlement.hpp:586-587`) — booking the proposal would decrement owed for money the sink actually took.
3. **CREDIT = `{}`** until Track A2 S-1 emission exists. `E_b` is the fold over *b*'s burial-gated prefix; there is no XMR sharechain fold yet, so the empty map is the honest value. `FoundBlockEvent::credit` exists so S-1 becomes a fill, not a re-wiring. Option A books the degenerate `{}/{}`. The old `amounts_from()` helper is deleted.
4. **The maps survive the thread boundary and the restart.** The pending-FOUND sidecar moves to **schema v2**, persisting both maps. A v1 line still parses, is flagged legacy, and is re-driven with **empty** maps (the fail-safe direction — owed is never over-decremented) with a loud warning. Without this, a restart inside the `D_conf` window re-drove a *different* payout than the block paid.
5. **A fail-closed INV-1 precheck** refuses any FOUND whose payout exceeds this ledger's `EffectiveOwed` for that key — which can only mean the coinbase was built against a different ledger — instead of poisoning owed negative. Surfaced as `split_brain_refused` in the status line.

**Invariants restored**

* **INV-1** — `EffectiveOwed(k) ≥ 0` at every step, FOUND and FINALIZE alike.
* **INV-2 (OI-W4-5)** — `on_block_finalized` moves `EffectiveOwed(k)` by **exactly `+credit[k]`** (it removes the same payout from the pending term that it subtracts from `finalW`): invariant when `credit == {}`, never decreasing. `EffectiveOwed` legitimately **drops at FOUND** — that is the pending-payout deduction the coinbase drew on — so monotonicity is a property of the FINALIZE transition, not of the whole run. (Asserting otherwise makes the KAT wrong; the first draft of KF-1 got this wrong and failed.)

### Defect: count-invariance — two opposite conventions for `0`

W4 canon reads `slot_budget_C == 0` as **UNBOUNDED** output count (`w4_settlement.hpp:580-583`, symmetric with W5's `max_payout_bytes == 0`). X6 reads its own `cap_owed == 0` as **"no room"** (`xmr_coinbase.cpp:113-114` + `:135`). `project_w4_owed` computed X6's number and fed it to W4's parameter **without translating**, so at `total_output_cap == n_fixed + 1` W4 proposed *every* eligible key while X6 emitted *zero* owed outputs and the residual sink silently swallowed the whole owed budget — `same_set()` still converged on the sink-only shape and the block shipped paying nobody their owed.

**This is a reachable production path, not an edge case.** `weight_aware_output_cap` floors its result at 1 (*"always room for at least the residual sink"*, `xmr_coinbase.cpp:430-433`), so any block whose selected txs fill the penalty-free zone to within ~212 bytes resolves the cap to `1 == n_fixed + 1` under the default empty `fixed` set. It fires when the chain is **busiest**.

**Fix:** `project_w4_owed` **clears and succeeds** at `cap_owed == 0` — zero owed rows, every key CARRIED with its `first_eligible` age intact (the same disposition as W4's own sub-`h_min` CARRY). *Clear-and-succeed, not refuse*, for three independent reasons: (i) X6 itself treats `output_cap == fixed.size() + 1` as **legal** (`BuildError::CapTooSmall` is `< fixed.size() + 1`), so refusing would make the consumer reject a template X6 accepts; (ii) a refusal here is **permanently fatal**, not a retry signal — the weight-aware cap is a deterministic function of the same inputs, so the rebuild refuses identically until `max_passes` is spent and the node stops serving every time its block is full (fail-*stopped*, not fail-closed); (iii) zero owed rows means nothing over-paid and no payee short-changed — the owed is paid by the next block with cap room.

**The two fixes must land together.** With R-7 wired proposal-side and the guard missing, a cosmetic divergence would become real loss to payees.

Also corrects a design comment in `xmr_block_assembly.hpp` that asserted the opposite of the W4 contract (*"the FOUND payout map … over ALL outputs (owed / fixed / sink)"*), and pins why `in.h_min = 0` into X6 is load-bearing.

### New KATs

`v37_xmr_o2_settlement_kat` — **115 checks, was 62**:

| KAT | What it pins |
|---|---|
| **KF-1** | One-ledger INV-1/INV-2 over **three concurrent pending FOUNDs** (the exact `pending=3` shape). **Regression pin**: the split ledger drives `EffectiveOwed` to `−3·reward`. |
| **KF-2** | The ledger decrement is exactly the emitted `Role::Owed` set; fixed and sink identities are never ledger keys; and a non-zero X6 `h_min` is shown to make the executor emit a **strict prefix** of the proposal — the reason the FOUND map must be read from the emitted outputs. |
| **KF-3** | No double pay: a row in flight yields a **sink-only** next block. **Regression pin**: the shipped split wiring re-pays it at full `EffectiveOwed`. |
| **KC-5** | `cap_owed == 0` emits zero owed rows at `n_fixed` 0 and 2; the block is exactly `n_fixed + 1` outputs; carried ages undamaged. |
| **KC-6** | Reachability (`weight_aware_output_cap(300000, 299900, 2700) == 1`) and the two `0`-conventions asserted side by side — the guard against "simplifying" the new early return away. |
| **KC-7** | The guard is unreachable at the default cap (back-compat). |

`v37_xmr_o2_finalize_connect_selfcheck` — **28 checks, was 20**. FC3 / FC10 / FC16 used to **assert the buggy contract** (`effective_owed(payee) == -reward` while pending) and now assert the reconciled one; plus the split-brain precheck (FC3c), the sidecar-v2 round trip (FC8c/FC17), the v1 legacy path (FC18) and a truncated-line refusal (FC19).

### Build + KAT results (host `workstation-191`, gcc-15, `-DXMR_BUILD_RANDOMX=ON -DXMR_BUILD_OPTION_B_KATS=ON`)

```
c2pool-v37-xmr                              built, 0 errors
v37_xmr_o2_settlement_kat                   115 passed, 0 failed   RESULT: GREEN
v37_xmr_o2_finalize_connect_selfcheck        28/28 passed
v37_xmr_o2_live_submit_selfcheck             55 passed, 0 failed
v37_xmr_node_smoke                           11/11
xmr_block_assembly_kat                      192 passed, 0 failed   RESULT: GREEN
xmr_template_primitives_kat                 ALL PASS (0 failures)
```

### Re-proof on an isolated private regtest

`monerod --regtest --offline --fixed-difficulty`, own data dir, own ports `18189/18188/18190`, stratum `3433`, all on `127.0.0.1` of the build host; real `xmrig` `rx/0` → RandomX gate → `submit_block`. The live stagenet node was never touched. The run includes a **deliberate kill and restart inside the `D_conf` window**.

```
44 FOUND registered, 35 FINALIZED, 4 orphaned, pending=3 at the kill point
negative min_effective_owed lines across BOTH runs:  0
```

**min `EffectiveOwed` for the whole run = `0`** — never negative, including while blocks were pending. Sample lines:

```
FOUND registered 4ab23aa54b7e… h=1 reward=35184338534400 piconero owed_rows=1
  owed_paid=1000000000000 credit_rows=0 (operator payee=b9d5bd32d700…, informational)
FOUND registered 07eeab1088b9… h=2 ... owed_rows=0 owed_paid=0 credit_rows=0
FINALIZED 4ab23aa54b7e… h=1 SETTLED at bin_height=4 effective_owed(dc9eac09c13a…)=0
  min_effective_owed=0 ledger_seq=7
owed: min_effective_owed=0 booked=1000000000000 ledger_seq=51 split_brain_refused=0
```

The seeded owed was paid by **exactly one** block (`owed_rows=1`, `owed_paid=1000000000000`) and every later block was **sink-only** (`owed_rows=0`) — the emitted `Role::Owed` set **shrank to empty** instead of repeating the same payee at full `EffectiveOwed` on every block. The schema-v2 sidecar re-drove all 3 pending FOUNDs after the restart (`boot: re-drove pending FOUND … h=24/25/26`) and they finalized normally.

---

## F-1 — the startup owed seed could leave a NEGATIVE row across a restart (FIXED)

Found by the R-7 verify, closed here.

### The defect

`--owed-demo-amount` called `OwedLedger::on_block_found()` / `on_block_finalized()` **directly** on the ledger, bypassing the settle-store write-ahead log. R-7 had just made that ledger the node's own — durable and `RecoveryDriver`-replayed — so the seed's **credit** leg was in-memory only, while the coinbase **payout** that drew on it went through `XmrFinalizeDriver`'s WAL and *was* persisted.

A restart against the same data-dir therefore replayed the payout with **no matching credit**, and `finalW(k)` landed at `-(amount paid)`: a negative `effective_owed` row that **survives the restart** — a whitepaper §9 no-double-pay / OI-W4-5 violation. It was silent because this store's `RecoveryDriver` (unlike `w6_persistence`'s) has no `ledger_seq` cross-check to trip on the missing events. Re-passing the flag then re-credited on top of the recovered ledger, so the row was applied **once per boot** rather than once ever.

Live evidence at the previous head (`6a56fca2`), from the R-7 verify's own regtest logs — a restart without the flag:

```
recovery: hw_height=65 cursor=62 events=132 (resumed)
owed: min_effective_owed=-1000000000000 booked=0 ledger_seq=129 split_brain_refused=0
```

`-1000000000000` is exactly the seeded amount.

### Why the R-7 acceptance KAT masked it

Phases 1–3 of `finalize_connect_selfcheck` arm owed the **right** way — a FOUND carrying an S-1-shaped `E_b` credit, write-ahead-logged — so the credit leg replays exactly like the payout leg and `FC8a` / `FC10b` pass across their restart. The daemon's seed used a **different** path that no phase covered at all, so no assertion could fail on it. "The probe only checks the fresh-boot path" is precisely this: the restart assertions existed, but never over the seed the daemon actually used.

### The fix

| change | where |
|---|---|
| `XmrFinalizeDriver::seed_owed_durable(bid, credit, bin_height)` — writes the seed as two ordinary write-ahead events (`FOUND(credit)` / `FINALIZE`) under a stable bid, so `RecoveryDriver` replays it exactly like any other event | `xmr_finalize_driver.hpp` |
| restart-idempotence for free: the replay leaves the bid in the ledger's **SETTLED (terminal)** set, and `on_block_found` is already idempotent per bid, so a later call is a no-op — applied **exactly once** for the life of the data-dir. The seed bid is kept out of the driver's found/by-height maps: a seed is not a mined block and must never be walked by the `D_conf` cursor | `xmr_finalize_driver.hpp` |
| the daemon routes the flag through it and reports `applied` / `already-durable` (flag ignored, loudly) / `refused` | `main_v37_xmr.cpp` |
| the payee resolver is learned **unconditionally when serving** — the owed ROW is durable but the `key -> ScriptRef` map is in-memory; without this a recovered row would resolve to the RAW sentinel and be CARRIED forever once the flag was dropped | `main_v37_xmr.cpp` + fixture |
| `XmrOwedFixture` gains resolver-only `learn_pay()` and a `DurableLedger` ctor tag; `seed_owed()` **REFUSES** a direct write against a ledger declared store-backed, so the defect cannot walk back in. KATs driving a bare in-memory `OwedLedger` are unaffected | `xmr_o2_settlement_fixture.hpp` |
| every boot prints the post-recovery `min_effective_owed`, so a restart-negative is visible immediately instead of surfacing later as a coinbase that quietly pays nothing | `main_v37_xmr.cpp` |

### The corrected acceptance KAT

New phase 4, with `FC20` as a **negative control** that reproduces the OLD non-durable seed and **requires** the restart to come back negative — so `FC21`'s non-negative assertion cannot pass vacuously:

```
[PASS] FC20  NEGATIVE CONTROL: the OLD non-durable seed DOES leave a negative row
             across a restart  -- min_effective_owed=-250000000000 want=-250000000000
[PASS] FC21  after a RESTART against the SAME data-dir, min_effective_owed >= 0  -- 0
[PASS] FC21b the seeded row nets to EXACTLY zero (credited once, paid once)  -- eo=0
[PASS] FC21c the store replay restored the seed as SETTLED under its stable bid
[PASS] FC22  re-passing the seed applies NOTHING -- ledger_seq 4/4, WAL seq 4/4
[PASS] FC22b a DIFFERENT --owed-demo-amount on restart is equally ignored
[PASS] FC23  restart INSIDE the D_conf window also recovers non-negative  -- min=0 eo=0
```

### Re-proof on an isolated regtest

Own data-dir and ports (`monerod --regtest --offline`, RPC 18489 / ZMQ 18490 / stratum 3633), real `xmrig` rx/0 into the RandomX gate into `monerod submit_block`. Seed 500 XMR:

```
run A  (fresh)   owed-demo: seeded 500000000000000 piconero ... through the settle-store
                 write-ahead log (bid=v37-owed-demo-seed, bin_height=1) -- durable,
                 replayed on restart, applied exactly once
                 owed: post-recovery min_effective_owed=0 (INV-1 holds)
                 owed: min_effective_owed=0 booked=316624017924021

run B  RESTART against the SAME data-dir, and WITHOUT the flag
                 recovery: hw_height=63 cursor=60 events=17 (resumed)
                 owed: post-recovery min_effective_owed=0 (INV-1 holds)
                 owed: min_effective_owed=0 booked=183375982075979
```

`316624017924021 + 183375982075979 = 500000000000000` **exactly** — seeded once, paid once, across the restart, and the residual row was still payable with the flag dropped. A separate run restarted **with** the flag still set reported:

```
owed-demo: --owed-demo-amount IGNORED -- this data-dir already carries the durable seed
(bid=v37-owed-demo-seed) ... effective_owed(demo payee)=0
owed: post-recovery min_effective_owed=0 (INV-1 holds)
```

Run B additionally recovered a store whose previous process had been killed mid-flight, with no torn-store refusal.

### Build

`g++-15` 15.2.0 on the dedicated build host. `v37_xmr_o2_finalize_connect_selfcheck` **35/35**, daemon `--mock-smoke` **46/46**, `v37_xmr_o2_settlement_kat` **115/115** — all GREEN.

`src/sharechain/v37` canon and the W4 `owed_digest` fold remain untouched: the diff is 4 files, all under `src/c2pool/v37/xmr/*` plus `main_v37_xmr.cpp`.

### REQUIRED OPERATOR RULINGS (paper-silent — implemented fail-closed, not self-picked)

* **F-1-R1 — what must a node DO with a negative row it recovers?** §4.4 explicitly permits a negative `EffectiveOwed` as over-credit netted **forward** (never a clawback), so refusing to serve would contradict canon; but serving a K_fair coinbase off a ledger carrying an *unexplained* negative row is the §9 hazard F-1 produced. This ships the non-committal side — **warn loudly, keep serving** — and does **not** self-pick. The operator rules whether a recovered negative must be fail-closed (refuse to serve) or netted forward.
* **R-7a — is the FOUND `payout` term the `Role::Owed` subset only?** The paper defines payout as *"the coinbase outputs broadcast in b (K_fair over EffectiveOwed, from propose())"*, which reads as owed-only, and the arithmetic forces it: a Fixed/Sink identity was never credited, so booking it drives that key's `finalW` permanently negative. Implemented that way; the paper does not say it in so many words.
* **R-7c — two same-height blocks whose coinbases pay the SAME owed row.** Both are assembled from the same ledger state, so the second FOUND's payout exceeds the (already deducted) `EffectiveOwed`. This revision REFUSES the second, fail-closed toward INV-1, and counts it as `split_brain_refused`. That is safe for non-negativity but is **not free**: if the refused block turns out to be the canonical one and the first orphans, the owed it paid is never extinguished and a later block can pay it again. The alternatives — book it and go negative, or clamp to `EffectiveOwed` and mis-state what the block paid — are each wrong in a different way, and the paper does not cover concurrent FOUNDs drawing on the same row. **Not self-picked**; the regtest run above hit same-height races (4 orphaned, one `header=MISMATCH` at h=9) but with sink-only coinbases, so the case did not fire. A proper fix probably belongs with S-1 emission, when a FOUND carries its own `E_b`.

* **R-7b — is an option-A (monerod-template) block accounted as a v37 settlement at all?** Implemented as the degenerate `{}/{}` record: the ledger registers the block and `ledger_seq` bumps, but no key is created and no owed moves.

---

### Honest gaps / rulings owed

* **R-7: FIXED in this revision** (was: "carried, unchanged by this PR"). The finalize record is no longer option-A shaped and the daemon no longer holds two disjoint ledgers — see the **R-7 reconciliation** section above. Two paper-silent sub-choices are surfaced as **R-7a / R-7b** below rather than self-picked.
* **R-3: section-13 root evolution is lane-consensus-visible.** master's `StateCommitment` has ONLY the `V37S` summary and `V37E` balance leaves — `grep -rn "NRG1\|NRC1\|NRX1" src/` is empty at the base commit. Adding an NR* sub-block later CHANGES the root and therefore every coinbase; it needs its own activation gate. **Operator ruling owed** on that gate.
* **R-2: the block carries a hash-BINDING of the root, not the 32 root bytes.** `keccak(domain || chain_id || state_root)`. A verifier holding the same ledger re-derives and compares (that is KS-2), but a headers-only device cannot extract the root from the coinbase to run the section-13 O(log n) balance proof against the block alone. **Operator ruling owed** on whether a future tag should carry the root in the clear.
* **R-4: the `h_min` CARRY-vs-BREAK window is currently empty on XMR** (`XMR_DUST == 0`), so the divergence the KW-1 regression pin exercises needs an explicitly configured `--settle-h-min`. The wiring is correct either way; this is a note, not a defect.
* **`--owed-demo-amount`: FIXED in this revision as F-1** (was: "written into the node's real ledger but not through the write-ahead log, so it does not survive a restart and is re-applied on every boot" — that sentence WAS the defect; see the F-1 section above). The seed is now durable and applied exactly once per data-dir. It remains a proof knob, not lane state; real owed arrives with Track A2 S-1 emission.
* **F-2 (PRE-EXISTING, out of scope, newly characterised): the daemon segfaults on `SIGINT` shutdown while the stratum listener is still delivering shares.** Reproduced on the **unmodified** `6a56fca2` binary and on this revision, both `EXIT_CODE=139` under the identical load, so F-1 neither causes nor worsens it. It is a teardown ordering race, not an owed-path bug, and the settle store survived it (run B above recovered such a store cleanly, `(resumed)`, no torn-store refusal). Worth its own fix; not attempted here.
* `FinalizeConnect::min_effective_owed()` walks `effective_owed_all()` on every FINALIZE and every status tick — O(K). Fine at prototype key counts; it should become an incremental minimum if K grows.
* The proof ledger is still the deterministic `XmrOwedFixture` (one seeded K_fair payee + the sink), because there is no live XMR `OwedLedger` yet — Track A2 S-1 emission. The section-13 root is therefore constant across the run.
* An incidental finding: the ed25519 points from `monero-project tests/crypto/tests.txt` that the impl-side assembly KAT uses do **not** pass `xmr_ref_valid` (X6 requires only well-formedness of an owed ref; the consumer projection requires torsion validity, and silently CARRIES a key whose ref fails). The new KAT uses canon KAT identities and the regtest wallet keys instead. Worth a follow-up look at whether the impl-side KAT should use valid points.

### Base

Rebased onto `05983eeb` (master, #1552 — a README-only change; the tested code is identical to the `b3e415b6` base the build ran against). The branch already carries option B merged at `2013555e` (#1539).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -181,7 +181,7 @@ jobs:
         # RandomX JIT/asm + intentional unaligned reads do not survive
         # -fsanitize, so the ASAN leg deliberately leaves it OFF. Light mode:
         # the KAT never allocates the 2 GiB dataset (256 MiB cache only).
-        run: cmake -S . -B build_ci -DCMAKE_TOOLCHAIN_FILE=build_ci/conan_toolchain.cmake -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Release -DCOIN_DGB=ON -DXMR_BUILD_RANDOMX=ON
+        run: cmake -S . -B build_ci -DCMAKE_TOOLCHAIN_FILE=build_ci/conan_toolchain.cmake -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Release -DCOIN_DGB=ON -DXMR_BUILD_RANDOMX=ON -DXMR_BUILD_OPTION_B_KATS=ON
 
       # Boost header-integrity + version gate: a torn/mixed Boost tree or apt-Boost
       # fallback fails HERE (fast, one TU), not deep in the coin build.
@@ -234,6 +234,7 @@ jobs:
                     xmr_crypto_kats xmr_ed25519_derivation_kat xmr_difficulty_kat xmr_coinbase_kat xmr_goldens_kat xmr_e2e_kat xmr_miner_block_path_smoke c2pool-v37-xmr \
                     v37_xmr_o2_live_submit_selfcheck v37_xmr_o2_finalize_connect_selfcheck \
                     xmr_randomx_verify_kat v37_xmr_o2_randomx_kat \
+                    xmr_template_primitives_kat xmr_block_assembly_kat v37_xmr_o2_settlement_kat \
             -j"${BUILD_J:-8}" 2>&1 | tee "${GITHUB_WORKSPACE}/ci-logs/build-tests.log"
 
       - name: Run tests
@@ -601,6 +602,7 @@ jobs:
             -DCMAKE_BUILD_TYPE=Release \
             -DCOIN_DGB=ON \
             -DC2POOL_GOLDEN_RESOLVED_CONFIG_GATE=OFF \
+            -DXMR_BUILD_OPTION_B_KATS=ON \
             -DCMAKE_CXX_FLAGS="-fsanitize=address,undefined -fno-sanitize=vptr -fno-omit-frame-pointer -fno-sanitize-recover=undefined -g1" \
             -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address,undefined -fno-sanitize=vptr" \
             -DCMAKE_SHARED_LINKER_FLAGS="-fsanitize=address,undefined -fno-sanitize=vptr"
@@ -659,6 +661,7 @@ jobs:
                     v37_xmr_canon_p1_kat \
                     c2pool-v37-btc v37_btc_node_smoke v37_btc_digest_drift_kat f1_finalize_kat_btc \
                     v37_xmr_o2_live_submit_selfcheck v37_xmr_o2_finalize_connect_selfcheck \
+                    xmr_template_primitives_kat xmr_block_assembly_kat v37_xmr_o2_settlement_kat \
                     xmr_primitives_selfcheck xmr_x2_node_kat xmr_receipt_kat xmr_torsion_kat xmr_wire_dos_check xmr_stratum_selftest xmr_provenance_gate \
                     xmr_crypto_kats xmr_ed25519_derivation_kat xmr_difficulty_kat xmr_coinbase_kat xmr_goldens_kat xmr_e2e_kat xmr_miner_block_path_smoke \
             -j4  # ASan+UBSan: cc1plus RSS ~2-3GB each; -j8 wedged runner cgroup memory.high (throttle-hang). -j4 caps peak <12G.
--- a/src/c2pool/v37/main_v37_xmr.cpp
+++ b/src/c2pool/v37/main_v37_xmr.cpp
@@ -306,7 +306,15 @@ static int serve_and_run(const XmrNodeConfig& cfg, LiveMonerodTransport& transpo
             e.block_id_hex    = hex_of(s.block_id);
             e.prev_id_hex     = hex_of(s.prev_id);
             e.reward_piconero = s.reward;
-            e.payee           = payee_key;
+            e.payee           = payee_key;   // informational since R-7 (never booked)
+            // R-7 (2026-09-10): the FOUND record books what the COINBASE PAID —
+            // the emitted Role::Owed set, captured at the SUBMIT seam from the
+            // very snapshot whose bytes went to monerod (never re-looked-up here:
+            // the provider ring can evict a template between submit and tick).
+            // Empty under option A (monerod's template pays no v37 owed) and for
+            // a sink-only option-B block. `credit` (E_b) stays empty until Track
+            // A2 S-1 emission exists.
+            e.coinbase_owed.insert(s.coinbase_owed.begin(), s.coinbase_owed.end());
             e.template_id     = s.template_id;
             e.nonce           = s.nonce;
             e.extra_nonce     = s.extra_nonce;
@@ -358,6 +366,15 @@ static int serve_and_run(const XmrNodeConfig& cfg, LiveMonerodTransport& transpo
                     static_cast<unsigned long long>(fs.orphaned),
                     static_cast<unsigned long long>(fs.refused),
                     fc.pending().size());
+        // R-7 acceptance probe, printed every status tick: the ledger-wide
+        // minimum EffectiveOwed. It must be 0 (i.e. no negative row) for the
+        // whole run, INCLUDING while blocks are pending. The pre-R-7 split-ledger
+        // wiring drove this to -reward per pending FOUND.
+        std::printf("  owed: min_effective_owed=%lld booked=%lld ledger_seq=%llu "
+                    "split_brain_refused=%llu\n",
+                    fc.min_effective_owed(), fc.stats().owed_booked,
+                    static_cast<unsigned long long>(node.ledger().ledger_seq()),
+                    static_cast<unsigned long long>(fc.stats().split_brain_refused));
         std::printf("  %s\n", rx.describe().c_str());
         const std::string g = sink.last();
         if (!g.empty()) std::printf("  gate: last=%s\n", g.c_str());
@@ -470,11 +487,12 @@ static int run_live(const XmrNodeConfig& cfg) {
             node.stop();
             return 2;
         }
-        std::printf("payee: identity_key=%s… (amount-honest FOUND/FINALIZE records)\n",
+        std::printf("payee: identity_key=%s… (INFORMATIONAL since R-7: this key is logged with "
+                    "each FOUND but is NEVER booked into the ledger — the FOUND payout is the "
+                    "coinbase's own emitted K_fair owed set)\n",
                     hex_of(*payee_key).substr(0, 16).c_str());
     } else {
-        std::printf("payee: no --payee-spend-hex/--payee-view-hex -> FOUND records are valueless "
-                    "{}/{} (the ledger still registers the block)\n");
+        std::printf("payee: no --payee-spend-hex/--payee-view-hex (informational only since R-7)\n");
     }
 
     // ── wire 2: RandomX runtime verify (init on main, BEFORE the listener) ───
@@ -506,6 +524,14 @@ static int run_live(const XmrNodeConfig& cfg) {
         scfg.chain_id   = cfg.lane_chain;
         scfg.h_min      = cfg.settle_h_min;
         scfg.output_cap = cfg.settle_output_cap;
+        scfg.allow_nonruled_local_only = cfg.lane_commitment_local_only;
+        if (!o2::parse_lane_commitment_source(cfg.lane_commitment_source, scfg.lc_source)) {
+            std::printf("REFUSED: --lane-commitment must be one of "
+                        "state-root (RULED) | owed-digest | explicit; got \"%s\"\n",
+                        cfg.lane_commitment_source.c_str());
+            node.stop();
+            return 2;
+        }
         std::array<std::uint8_t, 32> sink_B{}, sink_A{};
         if (serving) {
             if (!o2::hex32(cfg.residual_sink_spend_hex, sink_B) ||
@@ -519,12 +545,32 @@ static int run_live(const XmrNodeConfig& cfg) {
             }
         }
 
-        // The proof ledger (no live S-1 emission yet). Empty => the whole reward
-        // flows to the residual sink (one v37 output). --owed-demo-amount seeds a
-        // distinct K_fair OWED payee (the sink material with spend/view swapped —
-        // still two valid ed25519 points, a different identity) so the assembled
-        // coinbase carries an OWED output alongside the sink (multi-output proof).
-        o2::XmrOwedFixture ledger(cfg.lane_chain);
+        // ── R-7 (2026-09-10): ONE LEDGER ────────────────────────────────────
+        // The coinbase MUST be built against the SAME OwedLedger the finalize
+        // driver writes — the node's: store-backed, RecoveryDriver-replayed,
+        // XmrFinalizeDriver-driven. Before this fix the daemon held TWO disjoint
+        // OwedLedger objects (a settlement one the coinbase paid from, a node one
+        // FINALIZE booked into). The settlement one was never decremented, so
+        // propose_coinbase re-proposed the SAME owed row at its full
+        // EffectiveOwed on EVERY block (a no-double-pay violation of whitepaper
+        // section 9 / OI-W4-5) while the node one went negative by one full
+        // reward per concurrent pending FOUND.
+        // With one ledger, block N+1's proposal already sees block N's pending
+        // payout deducted (effective_owed = finalW - SUM_pending payout), so a
+        // row that is already in flight cannot be re-proposed.
+        // `node` outlives `ledger`, the provider and every retained template.
+        //
+        // --owed-demo-amount seeds a distinct K_fair OWED payee (the sink
+        // material with spend/view swapped — still two valid ed25519 points, a
+        // different identity) so the assembled coinbase carries an OWED output
+        // alongside the sink (multi-output proof). With no seed and an empty
+        // ledger the whole reward flows to the residual sink (one v37 output).
+        // NOTE (honest scope): the demo seed is written straight into the
+        // in-memory ledger, NOT through the settle-store write-ahead log, so it
+        // does not survive a restart and IS re-applied on every boot that passes
+        // the flag. It is a proof knob, not lane state; real owed arrives with
+        // Track A2 S-1 emission.
+        o2::XmrOwedFixture ledger(node.ledger());
         if (serving && cfg.owed_demo_amount) {
             ledger.seed_owed_std(sink_A, sink_B, cfg.owed_demo_amount);
             std::printf("owed-demo: seeded %llu piconero owed to a distinct K_fair payee "
@@ -534,9 +580,20 @@ static int run_live(const XmrNodeConfig& cfg) {
 
         o2::XmrSettlementTemplateProvider provider(transport, ledger, scfg, cfg.stratum_share_diff);
         o2::SettlementStratumTemplateSource template_source(provider);
-        std::printf("coinbase: %s (lane_chain=%u, residual sink %s)\n",
-                    to_string(cfg.coinbase), static_cast<unsigned>(cfg.lane_chain),
-                    serving ? "SET (torsion-checked at build)" : "UNSET (observe-side only)");
+        // Serve banner: name the two RULED consensus choices and print the
+        // resolved 32-byte lane commitment, so a node's committed value is
+        // auditable straight from the log (RULED 2026-09-10).
+        {
+            const ::v37::bytes32 lc = o2::resolve_lane_commitment(scfg, ledger.ledger());
+            char lc_hex[65];
+            for (int i = 0; i < 32; ++i) std::snprintf(lc_hex + 2 * i, 3, "%02x", lc[static_cast<std::size_t>(i)]);
+            std::printf("coinbase: %s (lane_chain=%u, residual sink %s, k_fair=%s, "
+                        "lane_commitment=%s%s %s)\n",
+                        to_string(cfg.coinbase), static_cast<unsigned>(cfg.lane_chain),
+                        serving ? "SET (torsion-checked at build)" : "UNSET (observe-side only)",
+                        o2::to_string(scfg.kfair), o2::to_string(scfg.lc_source),
+                        cfg.lane_commitment_local_only ? " LOCAL-ONLY-OVERRIDE" : "", lc_hex);
+        }
 
         auto candidate = [&provider](std::uint32_t tid, std::uint32_t en, sub::BlockCandidate& out) {
             std::string w;
@@ -615,6 +672,8 @@ int main(int argc, char** argv) {
         else if (a == "--settle-output-cap") cfg.settle_output_cap =
                      static_cast<std::uint32_t>(std::stoul(next("0")));
         else if (a == "--owed-demo-amount") cfg.owed_demo_amount = std::stoull(next("0"));
+        else if (a == "--lane-commitment") cfg.lane_commitment_source = next("state-root");
+        else if (a == "--lane-commitment-local-only") cfg.lane_commitment_local_only = true;
         else if (a == "--lane-chain") cfg.lane_chain =
                      static_cast<::v37::ChainId>(std::stoul(next("0")));
         else if (a == "--d-conf") cfg.d_conf = std::stoull(next("60"));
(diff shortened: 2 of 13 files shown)


frstrtr:
Closing: master settled the lane commitment on owed_digest, not a section-13 state root. See #1697 (RECON: coinbase-authority payout + on-chain credit cut) and #1704 (K_fair arming finalized-only, owed_digest V37O->V37Q). The durable owed seed (F-1) was replaced by the R-C restart work in #1718. The per-reward re-projection helpers here (reproject_owed / project_w4_owed) and the StateRoot commitment source are not on master. The branch is kept, and this PR can be reopened if that direction is revisited.

PR #1563, opened 2026-09-10
xmr(native): WF-C2b -- the trust-anchor bundle (format, fail-closed loader, generator, real stagenet bundle)

frstrtr:
**DO NOT MERGE** — DRAFT. Wave 1 component **WF-C2b** of the native-minimal Monero node. Branched off `c2be1b7c` (the frozen Wave 0 contracts, PR #1558), base `master`.

## What this is

`contracts/anchor.hpp` — frozen in Wave 0 — owns the `AnchorBundle` struct, the exact window sizes and `anchor_self_check()`. This PR lands everything downstream of that: **how a bundle is spelled on disk, how it is loaded, how it is minted**, and a **real stagenet bundle** to load.

`R-ANCHOR` = release-pinned self-generated bundle. It is the same class of artefact as monerod's own compiled-in checkpoints, and it is the thing that makes a cold start cost `O(anchor age)` rather than `O(chain age)`.

## Files

| file | what it owns |
|---|---|
| `anchor/xmr_anchor_sha256.hpp` | a small SHA-256 — the digest is the one field `contracts/` deliberately cannot verify (no crypto dependency there), so the loader carries it |
| `anchor/xmr_anchor_codec.hpp` | the `.inc` format: canonical body, digest rule, writer, and a fail-closed reader that judges FORM only |
| `anchor/xmr_anchor_load.hpp` | `load_anchor()` — the four gates |
| `anchor/xmr_anchor_generate.hpp` | `generate_anchor()` — the minting rules behind an abstract read-only `MoneroDaemonRpc` port |
| `anchor/xmr_anchor_embedded.hpp` | the release-pinned bundles compiled into the binary, one per network |
| `anchor/xmr_chain_anchor_stagenet.inc` | **the real stagenet bundle**, 24 KB |
| `tools/xmr-anchor-gen/xmr_anchor_gen.py` | the transport-bound capture path (monerod JSON-RPC, read-only) |
| `test/xmr_anchor_self_check_kat.cpp` | the KAT, 153 checks |

## Design decisions worth reviewing

**Line-oriented text, not a blob.** The bundle is the node's trust root. It is read by a human before a release pins it and diffed between releases, so its serialisation has exactly two jobs: unambiguous to a parser, readable to the person signing off. A binary blob is better at neither. The shape follows monerod's own checkpoints file — one fact per line, keyword first — because that is the artefact this one is the analogue of.

**One artefact, not two.** The file opens with a raw-string frame (`R"ANCHOR(` … `)ANCHOR"`), so the *identical bytes* load from a path and from the binary via `#include`. There is no text copy and header copy that can drift. A raw string literal is one preprocessing token formed before directives are considered, which is what makes the `#` provenance comments inside the bundle safe to embed verbatim.

**The digest covers the body, not the file.** Comments are outside it on purpose, so a release can annotate a bundle — which daemon, which date, which tip — and still be the same bundle. What the digest protects is the DATA. It is explicitly **not a signature**: anyone who can rewrite the file can rewrite the digest line. It catches a truncated fetch, a merge conflict and an editor.

**Run-encoded long-term window.** `ltw` carries 100 000 entries and Monero's long-term weight moves slowly, so a real bundle is long runs of one value. The grammar admits `<count>x<value>` beside a bare value; both expand to the same numbers. The real stagenet capture is **24 KB instead of ~700 KB**, with no loss of information and nothing different in what the loader ends up holding.

**Fail-closed, and the gate this function cannot close.** `load_anchor()` runs SOURCE → FORM → MEANING and then says, in code, what it cannot do: `anchor_confirmed_by_network()` must be called by C2's boot with the hash of the block peers actually served at `height`, and the node must refuse to start unless it matches. Gates 2 and 3 make a corrupted bundle reject the **true** chain — a loud halt, never a cheap false chain. The id is the one datum whose wrongness is a release-compromise scenario, and gate 4 is its only defence. `anchor_boot_duty()` exists to be quoted in that review.

**The generator has no transport.** `MoneroDaemonRpc` is an abstract read-only view: four queries, no way to reach a submit, a wallet or the peer list through it. Keeping it abstract is what lets the KAT stage the cases a live daemon can never be asked to produce on purpose — a broken `prev_id` chain, a short response, cumulative difficulty dipping mid-window, coins that underflow. Its last act is `anchor_self_check()` on what it just built, so a generator bug fails at the tool in front of the person minting the bundle, not at some operator's boot three weeks later.

**One format, two implementations, one pin.** The Python capture tool and the C++ `generate_anchor()` both write the same canonical body. Rather than trust that, the KAT re-serialises the parsed **real** bundle with the C++ writer and asserts the result hashes to the digest line the Python tool wrote. A byte of drift between them is a red KAT here rather than a bundle that loads on the author's machine and nowhere else.

## The embedded stagenet bundle

Minted from a **synced monerod 0.18.5.1** over **read-only RPC** (`get_info`, `get_block_headers_range`, `get_miner_data`, `get_coinbase_tx_sum` — queries only; nothing submitted, nothing mined).

```
H_a        2204000            (daemon tip 2204848; anchor buried 848 blocks)
id         c55f08bce08d3514f520611d6262afb15247e506d58453a73fd48067813dcdf5
prev_id    08034ed1a2196cb678ccb828d88448700a4cb5230d1a00c1e00ef65b9e314e79
version    16                 (= hf_version_for_height(Stagenet, 2204000))
seed       2203648            (= rx_seedheight(H_a + 1), epoch-aligned)
windows    735 difficulty / 100 short-term / 100 000 long-term
digest     6ef031fc7ce4be63aebef4e79d1f270c694edb20c89ebf04a8e8bcda0dc669a8
```

`already_generated_coins` is walked back from `get_miner_data` at the tip by subtracting each block's coinbase, and **cross-checked against `get_coinbase_tx_sum` over the identical range** — the capture refuses if the two disagree.

It carries **no `checkpoint` rows**: monerod's compiled-in stagenet checkpoints all sit below this height, and the field only ever holds checkpoints at or above `H_a`.

## The KAT — 153 checks, 0 failures

1. **The hash.** SHA-256 against the published NIST vectors, including the million-'a' streaming case and the 55/56-byte padding boundary where hand-rolled implementations break.
2. **The format.** A synthetic bundle survives write → parse → write with every field and all 100 835 window values intact; each documented damage produces its **own** `AnchorParse` code. A loader that says "malformed" for a bundle that is merely for the wrong chain is a loader that gets debugged at 3am, so the two vocabularies are tested apart.
3. **The judgement.** Every `AnchorStatus` in `contracts/anchor.hpp` is **reached** by mutating one field of an otherwise valid bundle. A status no input can produce is either dead code or a check that silently does not fire; this block is what tells the two apart.
4. **The loader.** All four gates, including a tampered file written to disk, a bundle offered to the wrong network, a missing path, and a network with no embedded bundle. A refused load always leaves the caller holding nothing.
5. **The generator.** Honest daemon, then nine fault cases, each refused with a reason and with `out` untouched.
6. **The real bundle.** Loaded from the binary, self-checked, round-tripped, and the Python↔C++ digest cross-check described above.

### Two defects the KAT found while being written

* **`generate_anchor` published a partly-filled bundle on failure.** The windows were written straight into the caller's `AnchorBundle` and the coins gate returned `false` after them, so a caller that ignored the bool held 100 000 real weights and a zero height. Fixed: the bundle is built into a local and assigned to `out` only on success. *(product bug)*
* **`to_hex(const Hash&)` and `to_hex(const std::array<uint8_t,32>&)` were a redefinition, not an overload** — `Hash` **is** that array. Fixed to one function. *(build break, caught on first compile)*

## Verification

Built on VM500 (gcc 13.3), capped with `nice`/`ionice` so as not to starve the stagenet sync running on that host:

| leg | result |
|---|---|
| `-std=c++20 -O1 -Wall -Wextra` | 153 checks, 0 failures — no warnings |
| `-std=c++17 -O1 -Wall -Wextra` | 153 checks, 0 failures |
| `-std=c++20 -fsanitize=address,undefined -fno-sanitize-recover=all` | 153 checks, 0 failures |

Registered in **both** `--target` lists in `.github/workflows/build.yml` in this same change — the #1539 lesson: a target added with `add_test` but never built reports `***Not Run` and exits non-zero.

## Scope fence

`src/impl/xmr/native/` only, plus the tool and its `.gitignore` unignore (following the exact precedent of `tools/dash/gen_mn_checkpoint.py`, tracked for the same reason). **No `src/sharechain/v37` canon is touched**, no consensus is activated, and nothing outside this tree changes behaviour. Header-only, STL-only: no RandomX, no boost, no libzmq, no conan additions.

## What is deliberately left to a later component

* **C2a's boot must call `anchor_confirmed_by_network()`.** This PR cannot do it — there is no P2P client yet. `anchor_boot_duty()` is the string that review quotes.
* **The real HTTP-backed `MoneroDaemonRpc`** is WF-C6a's, wired to this same port.
* **Mainnet / testnet bundles** are `nullptr` until a release claims those networks. Adding one is deliberately a code change plus a review of the bytes, not a config knob.
* **The `checkpoint` rows** are a release-time hand-copy from monerod's `checkpoints.cpp`; monerod publishes no RPC for them. The generator takes them through the port so it cannot know where they came from.

Sub-tasks of the native-node plan: `/home/ubuntu/v37-work/xmr-native-node-plan.md` sections 4.10, 8 (WF-C2b) and 11 (R-ANCHOR).


frstrtr:
**Stacking note.** This branch is cut from `c2be1b7c` (Wave 0 contracts, #1558), which is not on `master` yet, so GitHub's diff-vs-master shows 39 files: the 12 that are mine plus the Wave 0 commits underneath. **The WF-C2b change is exactly these 12** — `git diff c2be1b7c..HEAD --name-only`:

```
.github/workflows/build.yml
.gitignore
src/impl/xmr/native/README.md
src/impl/xmr/native/anchor/xmr_anchor_codec.hpp
src/impl/xmr/native/anchor/xmr_anchor_embedded.hpp
src/impl/xmr/native/anchor/xmr_anchor_generate.hpp
src/impl/xmr/native/anchor/xmr_anchor_load.hpp
src/impl/xmr/native/anchor/xmr_anchor_sha256.hpp
src/impl/xmr/native/anchor/xmr_chain_anchor_stagenet.inc
src/impl/xmr/native/test/CMakeLists.txt
src/impl/xmr/native/test/xmr_anchor_self_check_kat.cpp
tools/xmr-anchor-gen/xmr_anchor_gen.py
```

Nothing under `src/sharechain/` appears in either list. Rebase onto whatever #1558 lands as before this leaves DRAFT.

PR #1564, opened 2026-09-10
xmr(native): Wave 1 C1a -- Monero levin codec, epee portable storage, typed P2P messages

frstrtr:
**DRAFT — DO NOT MERGE.** Wave 1, component C1a of the native-minimal Monero
embedded node. Authored against the frozen Wave 0 contracts (`#1558`, branch
`v37/xmr-native-w0` @ `c2be1b7c`), which this branch is cut from.

## What this is

Everything needed to turn bytes on the Monero P2P wire into typed values and
back, and nothing else. No socket, no thread, no connection state: the
transport and handshake (C1b) and the peer pool with its DoS budgets (C1c) are
separate components that build on these three headers.

| file | what |
|---|---|
| `src/impl/xmr/native/p2p/levin_codec.hpp` | the 33-byte levin bucket header, the command ids, the frame classification rules, the inbound per-command size caps |
| `src/impl/xmr/native/p2p/epee_storage.hpp` | the epee portable-storage encoder plus a bounded decoder: the 2-bit size-mark varint, the type tags, sections and arrays |
| `src/impl/xmr/native/p2p/levin_messages.hpp` | the typed messages 1001/1002/1003/1007 and 2001–2010, decoding straight into the W0 contract types |

The message decoders produce `BlockEntry`, `ChainEntry`, `TxBlobEntry` and
`PeerSyncData` from `contracts/types.hpp`, so C2 and C3 consume wire data with
no second conversion step. `MAX_OBJECT_REQUEST_IDS` from `contracts/fetcher.hpp`
is enforced on the encode side, so we can never build the 2003 request that
would get us dropped.

## Five things read out of monerod rather than inferred

Each is a place an independently written codec would have been wrong on the
wire. Sources read verbatim (monerod master, 2026-09-10): `levin_base.h`,
`levin_protocol_handler_async.h`, `portable_storage_{base,to_bin,from_bin}.h`,
`keyvalue_serialization{,_overloads}.h`, `p2p_protocol_defs.h`,
`cryptonote_protocol_defs.h`, `net_utils_base.h`, `net/enums.h`,
`connection_context.cpp`, `cryptonote_config.h`.

1. **Sections are emitted in sorted key order.** `epee::serialization::section`
   holds a `std::map<std::string, storage_entry>`, so monerod writes entries
   lexicographically, not in the order the KV map declares them. An encoder that
   preserved declaration order would produce frames a daemon still parses — and
   would never be byte-identical to a capture, which is exactly the comparison
   C6 exists to run.
2. **`KV_SERIALIZE_OPT` omits the field on store at its default.** `rpc_port` at
   zero, `pruned` at false, `dandelionpp_fluff` at true, `prune` at false: all
   simply absent. The omission is the format, not an optimisation.
3. **Empty containers are never written**, and the loader's failure to find one
   is discarded (`KV_SERIALIZE` ignores its return value), so an absent
   container means empty at both ends rather than a parse error at either.
4. **`cumulative_difficulty_top64` is unconditional on store** — monerod
   branches on `is_store` and only makes the field optional when loading.
   Likewise `first_block` in 2007 is a plain `KV_SERIALIZE`, written even empty.
5. **The size-mark varint is not the CryptoNote LEB128 varint** used inside
   block and transaction blobs. Two varints, two codecs; the other one is
   `consensus/xmr_blob_reader.hpp` and the two are never mixed.

Two corrections to the C1 design lens fell out of the same reading and are
worth recording, because the lens is the document the rest of the wave reads:
epee `SIGNATURE_B` is `0x01020101`, which is `01 01 02 01` on the wire (the lens
wrote `01 02 01 01`), and current `COMMAND_TIMED_SYNC::response` carries only
`payload_data` and `local_peerlist_new` — the legacy `local_peerlist` dance and
the `local_time` field named in older sources are gone from master.

## The decoder is the attack surface

This is where an unauthenticated peer's bytes first get to choose how much we
allocate. It mirrors epee's own guards exactly — array counts refused against
the bytes actually remaining, duplicate keys rejected, a bool byte above 1
rejected — and then tightens them:

* recursion depth **8** instead of epee's 100 (nothing in this protocol nests
  deeper than 4), enforced on the encode side too, so we never emit a frame we
  would refuse to read;
* hard ceilings on entries, objects, strings and array elements;
* the array reserve grows with arriving elements instead of trusting a declared
  count, so a legitimate six-figure count in a 4 MB frame cannot make us reserve
  for it up front;
* where epee **truncates** an integer that does not fit the requested width,
  this decoder **refuses** it. An honest peer writes the declared width, so the
  only sender that trips this is one trying to make our number differ from the
  one it sent.

Per-command inbound caps (R-CAPS) are below monerod's own wherever a pool never
legitimately needs the headroom: 2002 and 2004 at 4 MiB / 32 MiB against
monerod's 128 MB, 2003 at 4 KiB against monerod's 2 MB. Unknown commands
(1004–1006, 2005, anything invented) get a cap of zero, which closes the
connection. Pre-handshake only 1001 and 1007 are legal, under the 256 KiB
ceiling. Fragments are refused on clearnet by default: fragmentation exists
solely for the tor/i2p white-noise feature, and a fragment on a TCP link is
either a bug or an attempt to make us buffer unbounded state.

## KATs

Four new targets, registered with `add_test` **and** added to both `--target`
lists in `build.yml` in the same change — a target registered but never built
reports as `***Not Run` and exits non-zero (#1539).

| target | checks | what it pins |
|---|---:|---|
| `xmr_levin_codec_kat` | 88 | header layout goldens, the frame classification table (noise and fragments included), the cap table, every rejection path |
| `xmr_epee_storage_kat` | 172 | the varint at every size-mark boundary, the type-tag matrix, sorted key order, and every decoder bound |
| `xmr_levin_messages_kat` | 142 | goldens for the handshake, the chain request and the address encoding; a round trip of every message; both shapes of `block_complete_entry` |
| `xmr_levin_fuzz_kat` | 25 | the bounded deterministic fuzz pass |

The fuzz KAT's load-bearing assertion is that **canonical re-encoding is
idempotent**: any buffer that decodes must re-encode and re-decode to the same
value tree, and the second encoding must be a fixed point. That makes the
encoder and the decoder each other's oracle, and it is what catches a
sort-order or size-mark bug that a round trip through our own encoder alone
would hide. It also asserts coverage rather than assuming it — the mutation
pass fails if no mutation still decoded, or if every one did.

`xmr_levin_fuzz_kat --dump-corpus <dir>` exports the 16-seed corpus (one real
body per command) as `.bin` files for an external libFuzzer run. The corpus is
generated from the message encoders rather than committed as binaries, so it
cannot drift away from the codec it exercises.

Three bugs were found by running these, not by reading:

* a golden of mine for the `1073741824` varint boundary was arithmetically
  wrong — the code was right, the expectation was not;
* the structured fuzz generator produced arrays of **empty strings**, which the
  decoder refuses. That turned out to be correct and monerod-faithful: epee
  pins `ps_min_bytes<std::string>` at 2 while an empty element is one byte, so
  monerod refuses the same frame and it cannot travel the network. The
  behaviour is now pinned by name in the storage KAT and the generator says why
  it emits at least one payload byte;
* the header fuzz driver reached the accept path **zero** times in 20 000
  tries, so it was testing only the reject path. It now seeds in three tiers
  (raw noise / valid signature / plausible frame) and asserts it reached both.

## What this does NOT prove

Byte parity against a real captured monerod frame. That is hardest-unknown
**U5**, it needs a daemon, and it belongs to the C6 parity rig. The goldens here
are *derived* from the serialization rules, so they check the encoder rather
than photograph it — which is the strongest offline claim available, and it is
deliberately not dressed up as more.

Also owed, and outside C1a: **U8** (testnet/stagenet genesis ids, needed before
the chain locator can be KAT'd for those networks — C1c) and the R-CAPS
ratification of the numbers in the cap table.

## Scope and base

Scope fence unchanged: everything under `src/impl/xmr/`. This tree is a work
source for the pool, activates no v37 consensus, and does not touch
`src/sharechain/v37` — the diff confirms it.

Base is `master` per the wave-1 brief. Until #1558 lands, this PR's diff
includes the Wave 0 commits it is cut from; the C1a change itself is the 12
files above. Rebasing onto whichever base the integrator rules for wave 1 is a
no-op for those files.

## Build

VM500 (gcc 13.3, C++20), capped at `-j2` under `nice`/`ionice` so as not to
starve the stagenet sync running on the same host. Both legs green:

* ordinary Debug: 7/7 ctest passed (3 Wave 0 KATs + 4 C1a KATs)
* `-fsanitize=address,undefined -Wall -Wextra`: 7/7 ctest passed, **zero
  warnings**

427 C1a checks, 0 failures.


frstrtr:
## Independent verification — SHIP

Fresh clone of `v37/xmr-native-c1a` @ `195bc9bc` on VM500 (gcc 13.3, `-j2` capped, nice/ionice, stagenet monerod untouched). Rebuilt and re-ran everything from scratch; did not reuse the author's build.

**1. Compiles / KATs.** Both legs, `-Wall -Wextra`, zero warnings on all 7 targets.

| KAT | ordinary | ASan+UBSan |
|---|---|---|
| `xmr_levin_codec_kat` | 88 / 0 | 88 / 0 |
| `xmr_epee_storage_kat` | 172 / 0 | 172 / 0 |
| `xmr_levin_messages_kat` | 142 / 0 | 142 / 0 |
| `xmr_levin_fuzz_kat` | 25 / 0 | 25 / 0 |
| 3 W0 KATs | 392 / 6279 / 59, 0 failures | same |

No sanitizer or leak reports. Counts match the claim exactly (427 C1a checks).

**2. Scope / hygiene.** `git diff c2be1b7c..HEAD` = 12 files, +4479/-11, all under `src/impl/xmr/native/` plus 2 lines in `.github/workflows/build.yml`. `src/sharechain/`, `src/core/`, `src/c2pool/` untouched. Both `--target` lists carry all four new targets (#1539 rule honoured). Commit SSH-signed, verified by GitHub, author `frstrtr`, English only, no attribution markers anywhere in the diff.

**3. W0 contract interface.** `levin_messages.hpp` consumes `BlockEntry` / `ChainEntry` / `TxBlobEntry` / `PeerSyncData` from `contracts/types.hpp` directly — no shadow copies, no re-declaration, and no edit to `contracts/`. `block_complete_entry`, `CORE_SYNC_DATA`, `basic_node_data` and `NOTIFY_RESPONSE_CHAIN_ENTRY` each match monerod's KV map field-for-field, including which fields `KV_SERIALIZE_OPT` omits at their defaults and the unconditional-on-store `cumulative_difficulty_top64`. Fan-in should not need rework.

**4. Gap U5 is no longer open for the handshake path.**

The PR states byte parity against a captured monerod frame is unproven. I proved it. A throwaway probe (not committed) built only from these three headers spoke to the live stagenet monerod on the build host over TCP:

```
PROBE: frame cmd=1001 class=Response rc=1 cb=844
PROBE: OK peer_id=f99a9cd23f2cb9d1 their_height=435380 top_version=11 peerlist=9
PROBE: U5 BYTE PARITY vs captured monerod 1001 response: IDENTICAL (captured 844, re-encoded 844)
PROBE: captured 2006 NOTIFY cb=959 decode=1 ids=29 prune=1 parity=IDENTICAL
PROBE: captured 1002 RESPONSE cb=190 decode=1 height=435380 parity=IDENTICAL
```

monerod v0.18.5.0 **accepted** a C1a-encoded `COMMAND_HANDSHAKE` invoke and a C1a-encoded `COMMAND_TIMED_SYNC` invoke, and all three frames it sent back decode and **re-encode byte-for-byte identically** through this encoder. That is real captured-frame parity, not a derived golden, and it independently confirms the three lens corrections in the PR body: sorted key order (a 844-byte frame with a 9-entry peerlist cannot round-trip identically otherwise), `SIGNATURE_B` on the wire as `01 01 02 01`, and a TIMED_SYNC response carrying only `payload_data` + `local_peerlist_new`.

Additionally: 40 000 bit-flip/truncation mutants of the captured 844-byte frame pushed through the handshake, chain-entry and new-block decoders under ASan+UBSan — 6 754 accepted, 33 246 refused, no trap. Real coverage of the accept path, unlike the 0-of-20 000 header driver the PR body reports fixing.

**Residual (all pre-existing, none blocking):** U5 is still open for 2001/2004/2008 (needs a block push — C6); U8 testnet/stagenet genesis ids (C1c); R-CAPS numbers await operator ratification. The three lens corrections must be folded into `v37-xmr-native-node-c1-levin-p2p.md` before C1b/C1c are authored against it.

Verdict: **SHIP**. Stays DRAFT / DO-NOT-MERGE per the wave rule.

PR #1573, opened 2026-09-10
xmr(native): C4 native template source — rebind the option-B assembler to the native chain index + txpool (DRAFT, DO NOT MERGE)

frstrtr:
**DO NOT MERGE — DRAFT.** Wave 1, batch 3 of the native-minimal Monero embedded
node. Authored on top of `v37/xmr-native-w1i2` @ `3c2669d6` (PR #1570: W0
contracts + C1a codec + C2a consensus + C2b anchor + C1b transport + C2c index +
C3 txpool). Base is `master` per the standing fan-out rule, so the diff shown
here includes the W1-i2 stack; the C4 change is the last commit.

## What this is

C4 is the component the whole native node exists for: it is where a Monero block
template stops needing a daemon.

The option-B block/coinbase assembler is rebound from the monerod
`get_miner_data` RPC to the native sources — tip / prev_id / difficulty /
median weight / already-generated-coins / seed hash from C2c's chain index
(`IChainView`), and the transaction set from C3's txpool (`ITxpoolSnapshot`) —
through the single `IMinerDataSource` seam Wave 0 pinned for exactly this.

### Three new header-only files, `src/impl/xmr/native/template/`

| file | what |
|---|---|
| `xmr_native_miner_data.hpp` | `NativeMinerDataSource` — the daemonless arm, over `IChainView` + `ITxpoolSnapshot` / `ITxBlobSource` |
| `xmr_monerod_miner_data.hpp` | `MonerodMinerDataSource` — the existing `get_miner_data` path, wrapped and otherwise unchanged |
| `xmr_template_arm.hpp` | `ArmResolver` — which arm serves, which shadows, and a loud, counted fallback |

### It derives nothing

`NativeMinerDataSource` computes no consensus value. All seven miner-data fields
come out of `TemplateInputs`, which C2a/C2c already derive and which
`xmr_native_consensus_state_kat` already pins against monerod over 600 real
heights. Re-deriving any of them in `template/` would be a second implementation
of a consensus rule — the exact split-risk the `contracts/` family exists to
prevent, and the same reasoning that made `xmr_tx_weight.hpp` W0-owned rather
than C3-owned.

What C4 adds is the seam, the readiness gate, the epoch/rebuild rule, the body
pins, and three guards that only make sense at the template layer:

| guard | effect | why |
|---|---|---|
| difficulty sanity band `[tip/4, tip*4]` | **REFUSES** the template | the retarget cannot legitimately move that far in one block, so a value outside it means our own window is corrupt — and a template on a corrupt target asks miners to grind the wrong number |
| stale tip (default 30 min) | **FLAGS**, keeps serving | the tip we have is still the best chain we know; the flag is what the arm resolver falls back on |
| peer floor | **FLAGS**, keeps serving | on regtest zero peers is a correct configuration, and a pool that stopped paying miners over a peer count would be broken, not safe |

### The byte-stability rule

The assembler stamps a fresh header timestamp on every build, so re-assembling
under an unchanged tip changes bytes a miner is already grinding and its shares
then miss on re-hash ("Low diff"). The epoch therefore moves on a tip change
always, and on a backlog change only under `backlog_refresh_s > 0` — at most
once per interval, and only when the selectable set really changed. The default
`0` is tip-only, which is bit-for-bit the behaviour the provider had before the
seam existed.

## The rebind

`src/c2pool/v37/xmr/xmr_o2_settlement_provider.hpp` gains an
`IMinerDataSource&` constructor and **keeps** its `IMonerodTransport&` one,
which now owns a `MonerodMinerDataSource` and pumps it once per refresh — the
same single RPC per refresh it always did. `assemble()` is byte-for-byte the
function it was. `SettlementSnapshot` gains `source_name` and the
`MinerDataEpoch`, so a parity sample can always be attributed to an arm without
guessing.

The option-B K_fair settlement coinbase — owed payees, exact-sum residual sink,
`owed_digest` in `tx_extra` `0x03` — is therefore preserved by construction: the
code that builds it did not move. The second KAT proves that as a *property*
rather than as a file diff, so it survives someone later editing the assembler.

## KATs

Two new `ctest` targets, both **registered AND built in BOTH `--target` lists**
in `.github/workflows/build.yml` (the #1539 Not-Run rule: a registered target
that is never built reports `***Not Run` and exits CTest 8).

### `xmr_native_template_kat` — 87 checks

The centrepiece is the **P-TPL parity table** against a REAL `get_miner_data`
captured read-only from a synced stagenet monerod (`0.18.5.1-release`, height
`2 204 960`):

```
  [PASS] height 2204960 == monerod 2204960
  [PASS] prev_id a1a18db281797fb6c9651a95500725a22f51e293b814bb62606607c76877a035 == monerod
  [PASS] major_version 16 == monerod 16
  [PASS] seed_hash 5bd07b3c898f74aa3923d10f2749c9d4ee7b36461a55948f389e6044acacb321 == monerod
  [PASS] difficulty {lo=3789826, hi=0} == monerod {lo=3789826, hi=0}
  [PASS] median_weight 300000 == monerod 300000
  [PASS] already_generated_coins 18164488901536847281 == monerod 18164488901536847281
  [PASS] median_timestamp 1789074815 == median of monerod's own last-60 timestamps 1789074815
```

Not one of those seven EQUALITY values was computed by this repository, so a
formula error cannot pass by agreeing with itself. `median_timestamp` is the
eighth field and is **not exposed** by `get_miner_data` at all (p2pool derives it
too), so it is checked against monerod's own median RULE — epee's `get_mid`,
restated in the KAT rather than borrowed from the code under test — over
monerod's own captured timestamps.

There is a negative control: the same source driven one block back must produce a
*different* difficulty, otherwise the parity above would be proving nothing.

Suite B is the other half of the claim: the same captured body is replayed
through the **production** `MoneroDaemonRpc::parse_miner_data` over a fake
transport — hex-string `"difficulty":"0x39d402"` and all, the PR #1529 shape —
which is what makes "the monerod arm IS the old path" a statement about the real
decode path rather than about a re-typed struct. Both arms then agree on all
seven fields in one comparison.

The rest: the readiness matrix (K-C4-3, incl. "not synced" and "seed
unreachable" as distinct refusals, and thin-peers / stale-tip proven to be flags
rather than refusals), the difficulty band with both closed edges and a
band-disabled control (K-C4-4), the epoch rule under both refresh policies
(K-C4-5), backlog policy plumbing incl. the explicit-policy overload the shadow
arm needs, pin lifetime across the retained ring, and the arm-resolver
serve/shadow/fallback matrix.

### `xmr_native_template_coinbase_kat` — 31 checks

The coinbase shape survives the rebind. Built from the native derivation
(including the *native* emission function on the *natively* carried coin total)
through the same `settle/xmr_coinbase.cpp` executor the daemon-ful path uses:

* one output per owed payee, in canonical K_fair oldest-owed-first order, with
  distinct one-time keys;
* `Sum(vout) == base_reward + fees` **exactly** — Monero post-HF13 has no
  burn-the-remainder escape hatch, so this is consensus, not tidiness — with the
  residual sink absorbing precisely the remainder;
* `owed_digest` under the `tx_extra` `0x03` merge-mining tag at depth 0, and a
  different `owed_digest` moving the root, the tx secret `r` and the whole key
  set;
* byte-identical `prefix` / `tx_extra` / `r` across the two arms;
* the pre-CARROT fence still firing on `major_version 17` from the native arm;
* a sensitivity pass proving **every** rebound field is load-bearing — without
  it, "the arms agree" would be consistent with the coinbase ignoring the miner
  data entirely.

### Parity capture provenance

`tools/xmr-c4-parity/capture_miner_data.py` is read-only and paced (200-header
pages, 0.4 s apart) and is the only thing in this component that ever touches a
daemon. It ran against the synced stagenet monerod, not against the VM500/.44
sync. `src/impl/xmr/native/test/gen_c4_parity_golden.py` turns its output into
the checked-in golden, and asserts on the way that the id monerod reported for
`rx_seedheight(H)` equals the `seed_hash` monerod reported at `H` — so the seed
schedule is cross-checked by the generator, not assumed.

The golden's chain rows deliberately start at `xmr_c2a_golden.hpp`'s
`TEST_FIRST`, so the 100 000-entry long-term weight window is seeded from that
golden and rolled forward, rather than re-captured with 500 more paged calls
against someone's daemon.

`xmr_native_template_kat --parity-json <file>` replays a fresh capture and is
deliberately **not** what ctest runs: CI must not depend on a network, and a
golden that only ever ran against a live daemon is a golden nobody can
reproduce.

## Build + test evidence

Built on the capped build host (`nice -n 15 ionice -c3 -j4`, no new `-jN` beyond
the #1522 cap), configured as the CI non-sanitizer leg does
(`-DXMR_BUILD_RANDOMX=ON -DBUILD_TESTING=ON`, Release, gcc, ccache):

```
=== BUILD OK ===
100% tests passed, 0 tests failed out of 22
```

That 22 is the full native ladder (all 15 pre-existing Wave-0/1 KATs plus the 2
new ones) **plus** the consumer targets that include the rebound provider:
`c2pool-v37-xmr` (the daemon itself, which is the only thing that includes
`main_v37_xmr.cpp`), `v37_xmr_node_smoke`,
`v37_xmr_o2_live_submit_selfcheck`, `v37_xmr_o2_finalize_connect_selfcheck`,
`v37_xmr_o2_randomx_kat` and `v37_xmr_canon_p1_kat`. New-KAT counts:
`xmr_native_template_kat` 87/0, `xmr_native_template_coinbase_kat` 31/0.

## Scope fence

`src/impl/xmr/` plus the one provider rebind in `src/c2pool/v37/xmr/`, which is
the seam the plan assigns to C4 (risk R14). No consensus digest is defined,
`src/sharechain/v37` is not touched, and nothing here activates any v37 rule.

## What is NOT in this PR

* **The wiring in `main_v37_xmr.cpp`.** The provider accepts either arm, but the
  daemon still constructs it from the transport. Selecting the native arm at
  runtime (`--xmr-template-source` / `--xmr-template-shadow`) is a deployment
  decision and belongs with the operator, not with this component.
* **The live M2 soak.** The plan's M2 exit is serve=monerod / shadow=native for
  72 h with P-TPL CLEAN, then the flip and another 72 h. That needs C1c (the
  peer pool) to feed a live index, and it is WF-I3's job. What this PR
  establishes is the seam and a replayable single-height parity pin.
* **C6's comparator.** `SettlementSnapshot` now carries what the oracle needs;
  the oracle itself is C6.


frstrtr:
## Review round 1 — three MUST-FIXes addressed (head `daebf73a`, force-with-lease over `62df176c`)

The C4 review returned NO-SHIP on three defects. All three are fixed, each with a regression suite that **fails on the pre-fix code and passes with the fix**. Re-verified on a capped build host (`nice -n 15 ionice -c3`, `-j4`, `XMR_BUILD_RANDOMX=ON`, Release, `BUILD_TESTING=ON`).

### R-C4-1 (HIGH) — `tx_body(id)` returned another transaction's bytes
`NativeMinerDataSource::pin_bodies_locked_()` zipped the body source's `out` vector with `ids` by index. `ITxBlobSource` **compacts** `out` and names absent bodies in `missing`, and an id can go absent because `selectable_backlog()` and `get_blobs()` are two separate lock acquisitions (mined / key-image conflict / expiry in between). One hole shifted every later body onto the wrong id, and C5's ARM A would have relayed a block whose bytes do not hash to the ids beside them.

Fixed: the fill walks the **ids** and consumes `out` in order, skipping exactly what was reported missing. A reply that satisfies neither shape (`out.size() + missing.size() != ids.size()`) files **no** body and is counted — a wrong body is worse than a missing one.

### R-C4-2 (MED) — an unchanged-tip re-snapshot left the live template with no pin
The pin key *is* the tip id, so `pin()` followed by the dedupe block's `unpin()` of the same key erased the pin just installed. Fixed: the superseded generation is retired **before** the new pin is taken, the key being re-pinned is never unpinned, and ring eviction unpins only a key no live generation still names.

### R-C4-3 (MED-HIGH) — the rebind changed the default/production arm's rebuild rule
`MonerodMinerDataSource::epoch()` reported `backlog_seq` = the `tx_backlog` size, so on the legacy transport constructor (the one `main_v37_xmr` uses) every change in the daemon's txpool **size** under an unchanged tip forced a full reassemble — fresh header timestamp, new template id — at the 5 s poll cadence. That is the "Low diff" job-churn hazard the provider itself argues is load-bearing, and it churned the `RING = 6` retained ring down to ~30 s of job history.

Fixed: `backlog_seq` is an **opt-in** rebuild trigger and the daemon arm does not opt in — it reports `0`, so the provider's added conjunct is `0 == 0` on every refresh of the production path. The native arm's sequence stays **additive** and off by default (`backlog_refresh_s == 0`).

**Production-arm identity, checked rather than asserted:**
1. the RPC leg moved verbatim — same `body_get_miner_data()`, same production `parse_miner_data`, same four operator-visible error strings;
2. `assemble()` does not appear in the diff at all, so the K_fair settlement coinbase cannot have drifted;
3. the only change to the rebuild condition is `&& m_cur.epoch.backlog_seq == ep.backlog_seq`, which is `0 == 0` for this arm;
4. **Suite J** drives the real `MonerodMinerDataSource` through the production decoder over eight scripted responses (one tip, with a txpool that grows, holds, drains and empties, then a tip move) and asserts the observed decision equals the **pre-seam oracle** at every step — 2 rebuilds in 8 refreshes. Non-vacuity: the script really does move the pool under an unchanged tip, and the size-derived rule would have rebuilt **7** times instead of 2.

### Results

| | before the fix | after |
|---|---|---|
| `xmr_native_template_kat` | 157 checks, **28 failed** (suites H, I, J) | **157 checks, 0 failed** |
| `xmr_native_template_coinbase_kat` | 31 / 0 | **31 / 0** |
| full `ctest` (native + consumer) | 21/22 (`v37_xmr_canon_p1_kat` ***Not Run*) | **22/22, 100% passed** |

The 28-failure column is a real run: the two pre-fix headers were restored into the same build tree, the KAT rebuilt, and the failures recorded, before restoring the fixed ones. Parity is unchanged and still green: the native arm's seven `get_miner_data` fields still equal the values a synced stagenet monerod reported at height 2 204 960 (suites A and B, 28 checks).

`v37_xmr_canon_p1_kat` is now built as well as registered, so ctest is fully green rather than 21/22 with a Not-Run (the #1539 rule).

### Also in this round
- `tools/xmr-c4-parity/capture_miner_data.py` is now **tracked**. It was silently ignored by `.gitignore:tools/*`, which is why the golden named a capture step that was not in the tree; it gets the same exception `tools/xmr-anchor-gen/` and `tools/dash/` carry.
- README: the body rule, the pin rule and the opt-in `backlog_seq` semantics are written down where the next component will look for them.

Canon untouched: zero `src/sharechain/` hits in the whole branch diff, and `v37_xmr_canon_p1_kat` passes. Still DRAFT.

PR #1577, opened 2026-09-11
xmr(native): W1-i4 C5 -- dual-arm found-block relay (self-contained 2008 fluffy push + on-demand monerod arm) [DO-NOT-MERGE]

frstrtr:
**DO NOT MERGE.** Wave 1 batch 4 component of the native-minimal Monero node. Branched off `v37/xmr-native-w1i3` @ `3dff0fda` (PR #1576, the 11 prior components), targeted at `master` so the diff is reviewable on its own; the integration branch is where it actually lands.

## What C5 is

Where a found block leaves the pool. It is the one component whose mistakes are paid for by somebody else: an invalid block does not bounce, it bans our P2P identity at every peer that received it, and a block that quietly fails to go out is a found block thrown away along with the share that paid for it. So most of what it does is refuse.

## The two arms

* **ARM A — levin P2P.** One `NOTIFY_NEW_FLUFFY_BLOCK` (2008) written to **every** handshaked peer in `state_normal` through C1c's `IBroadcastPort` (the DASH `submit_block_p2p_raw` redundancy policy: write to everyone, duplicates are a non-event, a miss is a loss). This is the **daemonless-primary** arm.
* **ARM B — monerod `submit_block`**, on demand: it fires only when a daemon is actually configured. The backup, not the gate.

Default order is `Parallel` with P2P first. `DaemonFirst` (the plan's M0–M4 posture, where a daemon rejection suppresses the P2P push entirely) and `P2pOnly` are configurations; all three are pinned by the KAT. An unset daemon sink reports `daemon_armed == false`, never `daemon_rejected` — C6 counts a daemon that REJECTED our block as a parity failure and no daemon at all as a void sample.

## The frame is self-contained

monerod relays an accepted block with `txs` EMPTY and lets peers fill the gaps with a 2009 round trip. We send the block blob **and every transaction body**, so a receiving peer connects the block without asking us anything. The bytes are cheap (single-digit KB); the round trip is not, because it is a round trip on the one message that must not be missed, against a peer that has never heard of us. `include_all_tx_bodies=false` mirrors monerod exactly for the parity run, and the 2009 responder stays wired either way. `current_blockchain_height` is H+1, so a peer that cannot connect the block asks for our chain instead of dropping us.

## Only a verified block gets out

`contracts/relay.hpp` states the exact 128-bit RandomX check as a caller precondition. C5 **also** enforces it, fail-closed: no gate installed means nothing is ever relayed, and any verdict but `Accept` stops both arms. The found-block path installs an **attestation** naming the id the O-2 verifier already accepted, rather than paying for a second ~25 ms hash on the hottest path — and that id is compared against the id of the bytes about to go out, so a wiring bug cannot relay a different block than the one that was verified. Everything cheap runs first: the blob must parse, hash to the claimed id, commit to the claimed tx list, carry the winning nonce, and agree about the coinbase height.

## Fail loud, never partial, never silent

A transaction with no body — or one that does not re-hash to the id the block commits to — **suppresses ARM A**. An incomplete self-contained frame is worse than no frame. ARM B still fires (monerod has its own pool). A verdict that reached nobody carries `why`, bumps a counter and is logged at error level.

## The c2pool-side redundant broadcast

Ports the v36/DASH pattern: every peer that can rebuild the winning block broadcasts it. The `xmr_found_block` carrier is the block blob plus attribution (a Monero block blob is already the compact form), and `FoundBlockIngress` is the receiving half, ordered cheap → heavy with **RandomX LAST**: per-peer budget, dedup, structure and id, the tip relation, the canonical-coinbase check, then the verify. A forged id, a wrong height or a non-canonical coinbase is a **ban**; a foreign branch, or a RandomX verdict that blames our own verifier, is a **defer**. The coinbase check gates RELAY only — lane admissibility is #1551's still-open question and is not answered here.

## Files

| file | what |
|---|---|
| `src/impl/xmr/native/relay/xmr_block_relay.hpp` | `LevinBlockRelay`: the arms, the 2008 frame, the 2009 responder, the retained-block book |
| `src/impl/xmr/native/relay/xmr_found_block_carrier.hpp` | the carrier codec and the admission ladder |
| `src/impl/xmr/native/relay/xmr_block_relay_submit_bridge.hpp` | the only file that includes the live-submit surface: `BlockCandidate` in, `LiveBlockSubmitter` as ARM B, and `CandidateBlockRelay` — the plan's `relay(candidate, nonce, extra_nonce)` |
| `src/impl/xmr/native/test/xmr_native_block_relay_kat.cpp` | the KAT |

All header-only, under the standing `src/impl/xmr/` scope fence. No consensus digest; `src/sharechain/v37` is not touched.

## KAT — `xmr_native_block_relay_kat`, 97 checks, 0 failures

Every block and every transaction body is a real stagenet capture with the identity monerod itself reported (`golden_c2a::BLOCKS`, `golden::FULL_TXS`), and suite Z re-derives both with this repository's own code before any other suite uses them. The single synthetic step is spelled out where it happens: a real 0-transaction block has its trailing `varint(0)` replaced by `varint(3)` plus three real transaction ids — the only shape in which a self-contained frame can be checked end to end, since the capture format carries block blobs and transaction blobs but no block whose transactions were also captured in full. Frames are encoded here and decoded back with the **C1a** codec, the way a peer would read them.

| suite | what it pins |
|---|---|
| Z | the fixtures are what they claim to be (ids re-derived, 12/12 bodies hash to monerod's ids) |
| A | the 2008 frame: command, Notify class, blob byte-identity, bodies in block order, H+1, and the monerod-mirror shape |
| B | the arm matrix across all three orders, including daemonless |
| C | never a silent drop: `reached_network()`, `why`, the error log line, the counter |
| D | the gate is the only door: no gate, every bad verdict, the hashing blob it is shown, and the attestation for a different block |
| E | structural refusals — empty, garbage, id, tx list, nonce, height — each with no arm fired, and RandomX not spent on a non-block |
| F | the partial-block refusal, with the `verify_tx_bodies=false` non-vacuity control showing the re-hash is what caught it |
| G | the 2009 responder: exact bodies in the requested order, out-of-range declined, retention boundary and expiry |
| H | the carrier codec and the admission ladder, with `pow_calls()` asserting RandomX never runs behind a cheaper rejection |
| I | PREFER-OWN: the block goes to our own index, and an index that refuses it does not un-relay it |
| J | the `BlockCandidate` bridge against the native id path |

Registered in **both** `build.yml` `--target` lists (the #1539 Not-Run rule). Built and run on workstation-191, capped: `xmr_native_block_relay_kat` 97/0, and the whole `xmr_native|xmr_levin|xmr_epee` set 22/22 with no regressions.

## Owed

* A **live** submit against an isolated regtest monerod (own datadir/ports) over ARM A is the one proof this PR does not carry; it needs C1c's dialer pointed at a local daemon and is tracked with the C6 regtest harness.
* `ArmOrder::Parallel` as the default is the brief's daemonless-primary posture; the plan's §2.2 rule 2 makes `DaemonFirst` the M0–M4 default. Both are implemented and both are KAT-pinned — which one ships is an operator call at the M-ladder, not a code change.


frstrtr:
## Live proof landed: a real monerod accepts our 2008 push

The "owed" item in the description is now discharged. `8c904780` adds `src/impl/xmr/native/relay/tools/xmr_c5_regtest_push.cpp`, which drives the **real** `LevinBlockRelay` and the **real** C1a codec over a raw socket against a real daemon.

**Rig:** monerod 0.18.5.1-release, two isolated regtest daemons on loopback (own data dirs, own ports 18188/18189 and 18288/18289), peered only with each other, `--regtest --fixed-difficulty 1 --no-igd --no-zmq`. Nothing on .40 or the .44 stagenet sync is touched.

**Run:** both nodes at height 11; template taken from node A, block pushed to node B, **daemon arm not configured**.

```
[live] block id aa6a3fa30d170b2b41ba6663aa29c74a10c5f266c24cda6177cdaf50fd229e98 at height 11, 0 txs, nonce 7
[live] <- command 1001 (HANDSHAKE), 262 bytes, class Response
[live] HANDSHAKED
[live] writing 217 bytes, command 2008
[relay:info] relay: block aa6a3fa3... h=11 peers=1 daemon=none first=p2p
[live] verdict: peers=1 reached=1 why=''
```

monerod's own log on node B:

```
[127.0.0.2:40747 INC] 184 bytes received for category command-2008 initiated by peer
[127.0.0.2:40747 INC] Received NOTIFY_NEW_FLUFFY_BLOCK <aa6a3fa3...> (height 12, 0 txes)
+++++ BLOCK SUCCESSFULLY ADDED
```

Both daemons then reported height 12, and `get_block_header_by_height(11)` on **both** returned `aa6a3fa3...` with nonce 7 — node B accepted our block and relayed it to node A over Monero's own P2P. A second push, this time through the CMake-built target, took both nodes to 13. `daemon=none` throughout: **the block reached the Monero network over levin alone**, which is the daemonless claim C5 exists to make.

**Build/run posture:** the tool is built by CI (both `--target` lists) and run by nobody — it needs a daemon and CI must not depend on one. It registers no ctest test, so the #1539 Not-Run rule does not apply to it; being in the target lists is what stops it bit-rotting.

`relay/tools/README.md` carries the recipe, including the two things that cost the most time to find:

* with `--offline` monerod never binds its P2P listener at all, and *without* `--offline` a daemon with zero peers never declares itself synchronized — so it refuses both mining RPCs and relayed blocks. Two daemons that handshake each other clear both within seconds, and the second node is also what proves onward relay.
* `--no-zmq` on both (the ZMQ RPC port has no per-instance default, so the second daemon dies on the collision), and a distinct source IP per client (monerod allows one connection per remote address, and the peer daemon already holds 127.0.0.1's slot).

**Still owed there:** the pushed block carries no transactions, so the self-contained body path is not yet exercised against a real daemon. That needs a regtest wallet filling the pool first and belongs to the C6 harness rather than to this tool.

PR #1580, opened 2026-09-11
xmr(native): M0 — assemble the node and follow a live tip over levin

frstrtr:
**DO NOT MERGE — DRAFT.** Stacked on #1579 (`v37/xmr-native-w1i4`, the complete
13-component native node): this PR's base is that branch, and it should be read
after it.

## What this is

M0 of the native-minimal Monero node plan: the milestone where the components
stop being a library and start being a node that follows a live chain.

Thirteen components landed in waves 0 and 1, each authored against the pinned
contracts and each proven on its own. None of them is a node. Every one of them
states a threading contract it does not itself provide, nothing in the tree
sends the first `NOTIFY_REQUEST_CHAIN`, and nothing seeds row zero. This PR is
that assembly plus the run that judges it.

## The proof

Two isolated `monerod 0.18.5.1 --regtest --fixed-difficulty 1` daemons on
loopback (own data dirs, own ports, peered only with each other), and the native
node dialing one of them over levin:

```
=== summary ===
tip            : height=24 verified_frontier=24 rows=25 synced=1 alt=0 orphans=0 reorgs=0
boot           : genesis booted=1 inspected=1 refusals=0 dropped=0 (genesis row seeded from the levin-delivered blob)
chain-entry    : start=0 total=21 ids=21 first=418015bb9ae9... last=66979fcc5a85...
peers          : handshaked=1 total=1 netgroups=1 silent=0 handshakes=1 dials=2/0 last_close= ()
levin in       : blocks=4 chain_entries=1 objects=2 txs=0 frames=7 dos_drops=0
randomx        : mode=1 hashes=24 prefetches=1 rekeys=1 verified=24 failed=0 foreign=0
monerod rpc    : calls=10 (parity/bootstrap only; never on the tip-follow path)
parity         : clean=5 fail=0 served_mismatch=0 void=0
M0-VERDICT: PASS heights=24 parity_clean=5 parity_fail=0 randomx_foreign=0
```

* **Heights 1–20 backfilled over levin**: one chain request from a
  genesis-terminated locator, one `RESPONSE_CHAIN_ENTRY` with 21 ids, and the
  index's own object requests for what it lacked.
* **Heights 21–24 followed live** as they were mined: four inbound
  `NOTIFY_NEW_FLUFFY_BLOCK` pushes, `chain_requests` flat at 1 — the driver goes
  quiet while the height keeps moving, which is the observable difference
  between catching up and following.
* **C6 P-TIP CLEAN at every sampled height**, all nine fields (`id`, `prev_id`,
  `cumulative_difficulty`, `difficulty`, `timestamp`, `reward`, `block_weight`,
  `long_term_weight`, `major_version`) against monerod's own `get_info` +
  `get_last_block_header`. The oracle coalesces a burst of tips to its newest by
  design, so the twenty backfilled heights produced one sample and the four live
  heights one each.
* **RandomX ran 24 times, all on the verify thread, `foreign=0`** — a counter
  rather than a claim; see below.
* **Zero monerod RPC on the tip-follow path.** The ten calls are exactly two per
  parity sample; `--probe-only` and `--no-parity` runs report `calls=0` while the
  index still follows.

Plus a **read-only live probe** against the stagenet daemon on `192.168.86.44`,
which is mid-sync and was not disturbed: one connection, one question, no blocks
fetched. It answered a genesis-terminated locator with `start=0 total=1067456
ids=10000`, `ids[0] == STAGENET_GENESIS` — the daemon's own proof that it
accepted our network id, our handshake and our locator terminus.

`node/README.md` carries both runs in full, with the rig recipe.

## The io-thread fence is a number

`ThreadWitnessPowSource` records the thread every RandomX evaluation actually ran
on and counts the ones that ran on a thread the node declared forbidden (the io
threads). The node prints it, and the KAT asserts the counter itself works — a
witness that could not catch a forbidden call would be worth nothing.

## Two defects this assembly found

Both were invisible to every component KAT, because both are about the parts
being wired to each other and to a real daemon.

1. **A parked block was never revisited when its parent connected.** C2c's
   branch path resolves descendants; the fast path (a block that extends the
   tip) did not. monerod sends its top block to any peer it sees as behind, so
   the first block a cold node receives is normally one it cannot connect yet —
   every cold start, not a corner case. The live symptom was a node that
   backfilled to height 8, sat there, and let four already-delivered blocks rot
   in the alt pool while its peer climbed, with no error anywhere because
   nothing had failed. Fixed by draining the tip's parked children as ordinary
   Extends (routing them through the fork-choice switch would announce a Reorg
   of depth 0 for what is simply the next block) and pinned by a regression that
   fails 2 checks on the code as it stood.
2. **The first chain request waited out the boot fetch's timeout** — 20 s of
   "connected, synced=0, nothing happening" on every cold start, which is the
   exact shape of failure this layer exists to make unmisreadable.

A third, recorded rather than fixed here: `ChainIndex::refetch_wanted()` is a
STANDING want list and is not cleared when a block arrives, so a driver that
asked for all of it every tick asked twice a second forever (observed: 361
`RESPONSE_GET_OBJECTS` frames for a 12-block backfill). The driver now
rate-limits per id and skips ids the index already has.

## Diff shape

| | |
|---|---|
| new, `src/impl/xmr/native/node/` | the assembly: worker loops, chain boot, sync driver, the parity/bootstrap HTTP arm, `NativeNode`, and `tools/xmr_native_node_main.cpp` |
| new, `test/xmr_native_node_kat.cpp` | 88 checks; registered in ctest and in both `build.yml` `--target` lists |
| changed, `chain/xmr_chain_index.hpp` | defect 1 above |
| changed, `p2p/xmr_peer_pool.hpp` | `Config::bind_ip` — the source address for outbound dials. monerod enforces one connection per remote IP and does not source-bind its own outbound connections, so on a loopback rig the two daemons occupy each other's 127.0.0.1 slot and a third participant is refused before the handshake. Same knob and reason as the C5 live-push tool's `C5_SRC_IP` |
| changed, `CMakeLists.txt` ×2, `build.yml` | `xmr_native_node` (built by both legs, run by neither — it needs a daemon) and `xmr_native_node_kat` (registered and run) |

Scope fence unchanged: everything under `src/impl/xmr/`, no consensus digest,
`src/sharechain/v37` untouched. 24/24 native KATs green on workstation-191
(Release, `-DXMR_BUILD_RANDOMX=ON`), zero warnings.

## Wired and deliberately not driven

C5's relay is constructed with its 2009 responder armed but nothing calls
`relay()` (M3). C4's arms are constructed and reported (`tmpl=native`) but the
option-B settlement provider is not rebound through them (M2). C3's txpool is
fed from levin and its chain-event subscription is wired, but per-transaction
equality against monerod's pool is M1. Anchor boot is wired and is the
production path; M0 ran on the genesis path because a fresh regtest chain is
younger than the anchor format can describe.

## Owed operator calls

* **R-ARMORDER — the found-block arm order.** `contracts/relay.hpp` defaults
  `ArmOrder::DaemonFirst` (the plan's M0–M4 posture); C5's own
  `self_contained_policy()` defaults to `Parallel` with the P2P arm first (the
  daemonless posture the track exists to reach). The node takes `DaemonFirst` as
  its default and exposes `--relay-order`; which one a production pool ships
  with is a ruling, and it is the last knob before M3 can claim a daemonless
  find.
* **The sustained P-TIP soak.** M0's exit criterion in the plan is 48–72 h with
  ≥1 EpochEdge and zero field diffs over ≥2200 blocks. This run is a SHORT proof
  (24 heights, 5 samples, ~22 s of wall clock) and the graduation ledger says so
  itself (`state: Observing`). Which network it soaks against — stagenet from a
  host that is not .44's syncing daemon — and for how long is the operator's
  call.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -244,7 +244,7 @@ jobs:
                     xmr_native_p2p_dos_kat xmr_native_peer_pool_kat \
                     xmr_native_template_kat xmr_native_template_coinbase_kat \
                     xmr_native_block_relay_kat xmr_c5_regtest_push \
-                    xmr_native_parity_kat \
+                    xmr_native_parity_kat xmr_native_node_kat xmr_native_node \
             -j"${BUILD_J:-8}" 2>&1 | tee "${GITHUB_WORKSPACE}/ci-logs/build-tests.log"
 
       - name: Run tests
@@ -682,7 +682,7 @@ jobs:
                     xmr_native_p2p_dos_kat xmr_native_peer_pool_kat \
                     xmr_native_template_kat xmr_native_template_coinbase_kat \
                     xmr_native_block_relay_kat xmr_c5_regtest_push \
-                    xmr_native_parity_kat \
+                    xmr_native_parity_kat xmr_native_node_kat xmr_native_node \
             -j4  # ASan+UBSan: cc1plus RSS ~2-3GB each; -j8 wedged runner cgroup memory.high (throttle-hang). -j4 caps peak <12G.
                  # xmr_provenance_gate: previously omitted here because
                  # -fsanitize=undefined rejected its constexpr pointer comparison
--- a/src/impl/xmr/native/CMakeLists.txt
+++ b/src/impl/xmr/native/CMakeLists.txt
@@ -116,6 +116,48 @@ if (BUILD_TESTING AND UNIX)
     target_compile_features(xmr_c5_regtest_push PRIVATE cxx_std_20)
 endif()
 
+# ─────────────────────────────────────────────────────────────────────────────
+# M0 (node/): THE ASSEMBLY. The wave-1 components are header-only and each one
+# states a threading contract it does not itself provide; node/ builds the
+# threads, the boot, the sync schedule and the status surface that turn them
+# into a running node, and node/tools/ is the entrypoint that runs it.
+#
+# Same discipline as xmr_c5_regtest_push above: BUILT in both build.yml
+# --target lists so it cannot bit-rot, RUN by nobody in CI (it needs a daemon
+# and a network), and registering no ctest test -- so the #1539 Not-Run rule
+# does not apply to it. What IS registered is test/xmr_native_node_kat, which
+# covers the parts of the assembly that can be judged without a daemon: the
+# queue adapters, the io-thread fence, the boot's derivation of the genesis
+# row, and the sync driver's schedule.
+#
+# Links boost::asio (header-only in the pinned Boost 1.90), a thread library,
+# and xmr_native_txpool (which brings xmr_coin for keccak, the tree hash and
+# the vendored difficulty retarget). It links `randomx` when
+# XMR_BUILD_RANDOMX is on, and defines XMR_NATIVE_NODE_HAVE_RANDOMX so the
+# node compiles the real LightVerifier in; without it the binary still builds
+# (the sanitizer leg has no RandomX) and refuses to call any block verified.
+# ─────────────────────────────────────────────────────────────────────────────
+if (BUILD_TESTING AND UNIX)
+    add_executable(xmr_native_node node/tools/xmr_native_node_main.cpp)
+    target_include_directories(xmr_native_node PRIVATE ${CMAKE_SOURCE_DIR}/src)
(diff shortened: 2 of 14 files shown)


frstrtr:

PR #1581, opened 2026-09-11
v37.1: activation prep (gate-OFF inert, ready-to-activate; golden + shipped()-flip owed to operator)

frstrtr:
**DO NOT MERGE.** This is V37.1 activation *prep*. It takes the release to the edge of the consensus flip and stops there. The remaining steps are the operator's hand, and they are written out in full at the bottom of this description.

## What this branch is

The five ordered V37.1 canon patches, applied to master @ `05983eeb`:

| order | patch | what it lands |
|---|---|---|
| 1 | `p1-rebase` | MRR retrofit rebased onto the shipped lane (seam S1) |
| 2 | `p2-window` | S6-S7 time-denominated window gate (`WinGate`) |
| 3 | `p3-ridge` | S12 native-ridge ladder (`nr_ladder.hpp`, `NrGate`) |
| 4 | `p1b-provgate` | RULING-B: position-gate the `V37V` provenance emission |
| 5 | `p4-s8` | S8 `owed_digest` fold + S-1 settlement emission (consumer tree) |

All five applied `--check`-clean in that order, 5/5, on a fresh clone. `p1b` is a diff against the p1+p2+p3 lane and refuses bare master by construction; `p4` is a consumer-tree diff that applies to bare master but does not *compile* without p1+p2+p3.

Files touched: `src/sharechain/v37/{v37_lane.hpp, v37_fixed.hpp, nr_ladder.hpp}` and `src/c2pool/v37/{w4_settlement.hpp, w5_coinbase.hpp}`. Nothing else.

## Gate-OFF anchors — both unmoved

Built on workstation-191, g++-15 (15.2.0), C++20, `-O2`, capped.

```
KAT-0  lane byte-identity, canon master vs the applied fork
       2479d5b61933cbe62d8506ebfcd09483954b8b4cc4698e593f73c16c83fda4fd
       109161 checks, 0 failures

owed_digest, gate OFF  (== sha256d("V37O"), the empty value)
       b4db1ded95a73f939975a259f9b48a1d182109f44397ed77e35d624f1a5cf339
       kat_s8_s1: 7143 checks, 0 failures
```

The KAT-0 fork side is a namespace-only fork of *this branch's* headers, reconstructed from the applied tree and then diffed against the fork the anchor was originally measured on: the only differences are two additive TEST-ONLY shims and a set of exception-message string renames (`v37(...)` → `v37retro(...)`). No canon line is modified. Zero removed lines fall outside those diagnostic strings.

## Activation-position audit — nothing finite is baked anywhere

26 factory configurations, **355 checks, 0 failures**. For `LaneParams{}`, `for_version(0)`, `for_version(1)`, `v37_0()`, `v37_1()`, `shipped()`, and `for_version(v, lane)` / `v37_1(lane)` across every `LaneKind`:

```
mrr.activation_pos == win.win_activation_pos == nr.nr_activation_pos == UINT64_MAX
```

and, on a live lane, `mrr_active()` / `win_active()` / `nr_active()` all false, with the lane digest equal to the V37.0 base lane's. The ungated geometry (`window`, `c0`, `rollup`, `half_life`, `level_caps`, `journal_depth`) is identical across every factory — there is no separate geometry gate in this header, so geometry inertness is an equality, not a sentinel.

A textual sweep confirms no numeric literal is assigned to any activation position anywhere outside test code.

One nuance worth stating plainly: the 2-argument `for_version(v, lane)` **does** declare `nr_version 1` / `win_version 1` and fold the ruled ND-R6 constant table. That is the header's own documented contract — *declaring the ruled table is not the flip*. The audit therefore pins the invariant that actually matters: a non-zero version marker is only ever seen beside an OFF position.

## Regression on the applied tree

| test | result |
|---|---|
| `v37_test` | 100576 / 0 |
| `v37_executor_test` | 10551 / 0 |
| `kat_provgate` (RULING-B canary) | 7 / 0 |
| `kat_nr_lane` | 94 / 0 |
| `kat_win_lane` | 267 / 0 |
| `v37_1_activation_test` | 8 / 0 |
| `v37_w4_settlement_test` | 899 / 0 |
| `v37_w5_coinbase_test` | 96 / 0 |
| `v37_w4_owed_incremental_test` | 14657 / 0 |
| `v37_w4_prereq_test` | 798 / 0 |
| `v37_w4_estimator_wiring_test` | 8 / 0 |
| `v37_r1_coinbase_budget_test` | 19 / 0 |
| `v37_subthreshold_estimator_test` | 11 / 0 |
| `v37_scaffold_test` | 6606 / 0 |

## Base note

Master has since advanced to `1184ace4`. That delta is LTC/DASH/CI only — `.github/workflows/*`, `src/core/web_server.cpp`, `src/impl/{ltc,dash}/*`, `test/test_dash_mn_checkpoint.cpp` — and touches **no** v37 file. The branch is cut from `05983eeb`, the exact base the patch set was verified against, so the apply is guaranteed clean; rebasing onto current master is a no-conflict operation whenever the operator wants it.

---

# REMAINDER RUNBOOK — the operator's hand

Four steps. Each one is a consensus act; none of them is in this branch.

### A finding that changes step (a)

The natural reading of "set the position on mrr + nr + win, co-located" is **wrong for the shipped V37.1 shape**, and the constructor will refuse it. Measured on this branch's header at `P = 40960`:

| shape | result |
|---|---|
| `mrr=P, nr=P, win=OFF` (nr_version 1, S12 native ridge) | **ACCEPTED** |
| `mrr=P, nr=P, win=P` | **REFUSED** — `ND-R9 — the S3 bin-band pyramid is RETIRED; S12 supersedes it (a finite window gate beside nr_version 1 is refused)` |
| `mrr=P, win=P, nr=OFF/v0` (S6-S7 window path) | ACCEPTED |
| `mrr=P alone` (MRR retrofit only) | ACCEPTED |
| `mrr=nr=P+1` (unaligned) | REFUSED — must be epoch-aligned |
| `nr=P, mrr=OFF` | REFUSED — the ridge sits above the MRR retrofit (S10) |
| `mrr=P, nr=P+4096` | REFUSED — ruling M / ND-R9 co-location |

Under `nr_version 1` the window **is** the native ridge's window (`m_W_bins`, driven by the S4 law). It has no separate gate, and a finite one beside the ridge is an attributable throw. So the V37.1 flip sets **two** positions, not three.

### (a) Pick `V371_ACTIVATION_POS` and set it

The symbol does not exist in the tree yet — the operator introduces it. Constraints the constructor enforces, uniformly, on every node:

- `V371_ACTIVATION_POS % 4096 == 0` (`epoch_len() == c0 == 4096`; the activation share must be a rebuild / checkpoint point)
- `mrr.activation_pos == nr.nr_activation_pos` exactly (ruling M / ND-R9 co-location)
- `mrr.activation_pos` finite is a precondition for a finite `nr.nr_activation_pos` (S10)
- `win.win_activation_pos` stays `UINT64_MAX` (ND-R9, above)
- it must sit far enough ahead of the live head that every node upgrades first

In `src/sharechain/v37/v37_lane.hpp`, inside `for_version(std::uint32_t v, LaneKind lane)`, in the `if (v == 1)` block that already folds the ridge dimensions:

```cpp
static constexpr u64 V371_ACTIVATION_POS = /* the ruled, epoch-aligned position */;
...
p.mrr.activation_pos   = V371_ACTIVATION_POS;
p.nr.nr_activation_pos = V371_ACTIVATION_POS;
// p.win.win_activation_pos STAYS UINT64_MAX — ND-R9 refuses a finite window
// gate beside nr_version 1.
```

Note that `p.nr.coverage_blocks` is currently pinned to `d.w_default_bins` and is flagged **OWED (ND-R6)** in the header — it is economic policy, it is digest-committed in `NRG1`, and the header says in terms that *the gate-ON golden must not be minted over it*. Rule it before step (c), or the golden minted in (c) is minted over an unruled constant.

The RDWR-OQ2 estimator gate (`SubthresholdGate`) carries **no position of its own** — it is a pure `{enabled, K, mode, version}` record, already ON in `for_version(1)`. There is nothing to co-locate there.

### (b) Flip `shipped()` to the live value

`SHIPPED_CONSENSUS_VERSION` is already `1`, and `shipped()` already returns `for_version(1)` — but the **1-argument** `for_version(v)`, which names no `LaneKind` and therefore leaves both the window and the ridge on their OFF defaults. So the flip in (b) is not a version bump; it is giving `shipped()` a lane:

```cpp
static LaneParams shipped() { return for_version(SHIPPED_CONSENSUS_VERSION, /* LaneKind */); }
```

This is exactly the open **WIN-R2** ruling the header names: whether `shipped()` should carry a default `LaneKind`, or whether every V37.1 lane must be constructed through `for_version(v, LaneKind)` at the call site. Nothing self-picks it. Decide WIN-R2 first; (b) is the mechanical consequence.

After (a) + (b), the gate-OFF anchors **will** move for `shipped()` — that is the point. `for_version(0)` / `v37_0()` must still produce `2479d5b6…` and `b4db1ded…`; if they do not, the flip has leaked into the base version and must be backed out.

### (c) Mint the gate-ON `owed_digest` golden

1. Build the release binary from the tree as it stands after (a) + (b) — g++-15, C++20, the normal release configuration, not a KAT harness. The golden must come from the shipped binary, not from a probe.
2. Drive the **pinned activation-epoch stream** through it: the share sequence that crosses `V371_ACTIVATION_POS`, starting far enough before the boundary that the lane is warm and the ring has folded at least once.
3. Take `owed_digest()` **at the boundary** — at the first position where `mrr_active()` and `nr_active()` are both true, i.e. `next_pos > V371_ACTIVATION_POS`. That value is the gate-ON golden.
4. Commit it into `src/c2pool/v37/test/v371_activation_golden_v1.json` beside the existing `OWED_DIGEST_GATE_OFF_HEX`, and regenerate the `.sha256` stamp next to it. `v37_1_activation_test` case 4 reads `OWED_DIGEST_GATE_ON_HEX` from that file and checks both that it matches and that it *differs* from the gate-OFF value.
5. Record alongside it the exact stream, the binary's commit hash, and the ruled `coverage_blocks` — the golden is only reproducible against all three.

Do not mint this from the current branch. Minting it before (a) is complete produces a digest over an OFF lane, which is the gate-OFF value again, and it would be indistinguishable in the file from a real gate-ON golden.

### (d) Re-run KAT-0

KAT-0 compares canon master's `v37::Lane` against the forked applied lane over 109161 checks. After (a)+(b) the fork's `shipped()` is no longer gate-OFF, so KAT-0 must be run against `for_version(0)` / `v37_0()` on the fork side, and it must still produce:

```
2479d5b61933cbe62d8506ebfcd09483954b8b4cc4698e593f73c16c83fda4fd
109161 checks, 0 failures
```

That identity is what guarantees a node replaying pre-activation history reaches byte-identical state after the upgrade. If it moves, the activation is not backwards-compatible and must not ship.

Then re-run the full set on this PR's table, plus `kat_provgate` — RULING-B's position gate is what makes the KAT-0 identity hold for arbitrarily-tagged streams, not just all-V37 ones, and it is the check that would catch a provenance byte moving the digest before the boundary.

---

### Verification artifacts

The proof tree, the reconstructed KAT-0 fork, the activation-position audit (`kat_actpos.cpp`), the flip-shape probe (`probe_flip.cpp`), and the patch set all live on workstation-191 under `/mnt/ci/scratch/v371-mint/` (proof clone at `mint/c2pool`). Nothing in this branch depends on them.

### DIFF
--- a/src/c2pool/v37/w4_settlement.hpp
+++ b/src/c2pool/v37/w4_settlement.hpp
@@ -39,7 +39,13 @@
 //                              settlement boundary. Its final form (Lane-
 //                              boundary check vs canonical flag) awaits the
 //                              integrator D-B ruling (OI-W4-8); this does NOT
-//                              hard-decide D-B and does not block.
+//                              hard-decide D-B and does not block. EXTENDED
+//                              for V37.1 (S8): the native-ridge dimension set.
+//   (7) S8 / S-1              — the ONE owed commitment, in both directions:
+//                              S8 is the fold that puts value into the ledger,
+//                              S-1 is the template-build emission of
+//                              ledger.owed_digest() of that SAME ledger. The
+//                              emission NEVER recomputes; see §S8 below.
 
 #include <algorithm>
 #include <array>
@@ -215,6 +221,24 @@ inline std::vector<WeightedPayee> project(const View& v,
 }
 
 // E_b = split(reward, project(view)) as a key→amount map (per-block entitlement).
+//
+// ★★ S8 — THE E_b SINGLE-SOURCE RULE (V37.0 / V37.1). E_b is a pure function
+// of (reward, v.payout, v.identities) and of NOTHING else. This fold does not
+// read v.params, v.next_pos, v.digest or any lane gate, and it must never grow
+// a live-vs-ridge branch of its own. The V37.0/V37.1 choice of WHICH map
+// v.payout carries is made ONCE, at the engine's single view-build site —
+// LaneExecutor's `s->payout = l.payout_map()`, and Lane::payout_map()
+// dispatches on its own nr_active():
+//     gate OFF  →  payout_map_positional()   (live only)          ≡ master
+//     gate ON   →  nr_payout_map()           (live + carry, CARRY-ARITH:
+//                                             two SEPARATELY truncated
+//                                             products summed, never
+//                                             (live+carry).mul_q)
+// — with ND-R11 RULED (the carried ledger is PAID) selecting the live+carry
+// form at the flip. Keeping the decision at that one site is what makes S8 and
+// S-1 auto-consistent: every node at the same cut credits the ledger from the
+// identical map and therefore emits the identical owed_digest. A second branch
+// HERE would be a fork surface, not an optimisation.
 template <class View>
 inline std::map<bytes32, u64> settle_block(u64 reward, const View& v) {
     std::vector<WeightedPayee> payees = project(v);
@@ -223,6 +247,31 @@ inline std::map<bytes32, u64> settle_block(u64 reward, const View& v) {
     for (std::size_t i = 0; i < payees.size(); ++i)
         if (amt[i] > 0) e[payees[i].key] += amt[i];
     return e;
+}
+
+// Which map the view-build site put in v.payout. DIAGNOSTIC ONLY — nothing in
+// the fold, the ledger or the commitment branches on it; it exists so a node
+// can LOG and a KAT can ASSERT which E_b source produced a given credit, and
+// so an operator reading two nodes' logs at one cut can see a mis-set gate
+// instead of inferring it from a diverged digest. It re-derives the lane's own
+// predicate from the view's own frozen fields (Lane::nr_active() is
+// `nr_version == 1 && next_pos > nr_activation_pos && mrr_active()`, and
+// Lane::mrr_active() is `next_pos > mrr.activation_pos`), so it cannot drift
+// from the site it describes.
+enum class EbSource : std::uint8_t { LiveOnly = 0, LiveAndCarry = 1 };
+
+inline const char* eb_source_name(EbSource s) {
+    return s == EbSource::LiveAndCarry ? "live+carry (V37.1 native ridge)"
+                                       : "live-only (V37.0 positional)";
+}
+
+template <class View>
+inline EbSource eb_source_of(const View& v) {
+    const ::v37::LaneParams& p = v.params;
+    const bool mrr_on = v.next_pos > p.mrr.activation_pos;
+    const bool nr_on  = p.nr.nr_version == 1 &&
+                        v.next_pos > p.nr.nr_activation_pos && mrr_on;
+    return nr_on ? EbSource::LiveAndCarry : EbSource::LiveOnly;
 }
 
 // ─────────────────────────────────────────────────────────────────────────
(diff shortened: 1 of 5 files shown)


frstrtr:
Superseded by c2pool#1607 (v37/v371-activate-flip @ 6ba47cb8) — the activated V37.1 (Shape A) flip: activation_pos=4096 baked, gate-ON golden 87c5249a minted, KAT-0 2479d5b6 unmoved. This branch was the gate-OFF-inert prep.

PR #1583, opened 2026-09-11
xmr(native): M1 — txpool parity per transaction against monerod [DO NOT MERGE]

frstrtr:
**DO NOT MERGE** — DRAFT. Stacked on `v37/xmr-native-m0` (PR #1580, the assembled node + M0 tip-follow). Base moves to `v37-dev` when M0 lands.

M0 proved the node follows a tip. **M1 proves the thing a block template is actually built out of**: for every transaction both pools hold, `{id, weight, fee, blob_size}` is the same number on both sides — our side computed by C3 from bytes that arrived over levin `NOTIFY_NEW_TRANSACTIONS`, monerod's side the daemon's own `get_transaction_pool`. No part of our number came from the daemon.

## Parity result (isolated regtest on workstation-191)

```
samples        : judged=582 unjudged=14 aligned=471 no-samples=452
transactions   : distinct_compared=67 daemon_ids_seen=67 never_compared=0
per-tx verdicts: clean=1368 fail=0 served_mismatch=0 void=0
fields         : compared=4104 equal=4104 differed=0 absent=0
evictions      : mined=67 key_image_conflict=0
dos            : frames_in=45 dropped=4
M1-VERDICT: PASS txs_compared=67 all_equal=1 fields=4104/4104 never_compared=0
M0-VERDICT: PASS heights=171 parity_clean=9 parity_fail=0 randomx_foreign=0
```

**67 distinct transactions, 4104 field comparisons, 0 differences, 0 absences.** Real wallet transactions from a `monero-wallet-rpc` against the daemon, signed and relayed the ordinary way — never handed to the node.

`never_compared=0` is the line that stops the seam being vacuous. The per-transaction comparison runs over the *intersection* (two pools are never identical at an instant), so a probe could report "all clean" over an intersection that was always empty. The tally therefore remembers every id the **daemon's** pool ever held and every id actually compared; the difference is a gate. Every transaction the daemon ever pooled was compared, field by field.

`dropped=4` — the C1 token buckets fired during the run, on a daemon re-relaying transactions we already held, and cost the claim nothing: all 67 ids were still compared. The buckets held.

## Both evictions, live

**Mined** — four `[M1-EVICT]` lines, one per block, each naming what left the pool and matching the pool's own `evicted_mined` counter. A counter with no id behind it names nothing; an id that vanished with no counter behind it could have gone for size or age. The proof needs both and has both.

**Key-image conflict** — a daemon will not make a double spend for you, so M1 builds one out of a transaction a real wallet really signed: the wallet builds it with `do_not_relay`; the harness injects its key-image **twin** (one `tx_extra` byte flipped — same key images, same commitments, same range proof, new id, and the flip is done with C3's own decoder because only a parse knows where `tx_extra` begins); the **original** then goes to the daemon, which relays it to us and gets a no-drop `KeyImageConflict` refusal (first-seen-wins); and the block that mines the original evicts the twin **by key image**, because the twin's id is nowhere in it.

```
[M1-INJECT] conflict.twin bytes=2166 verdict=Accepted drop_offense=0 id=33a99d6b796fad97...
[M1-EVICT]  h=174->175 left_pool=1 mined=0 conflict=1 ids=33a99d6b796fad97
evictions   : mined=0 key_image_conflict=1
conflicts   : rejected_key_image=1 (of which injected=0) | injected=1 accepted=1
M1-VERDICT: PASS conflict_evictions=1 never_compared=1
```

`rejected_key_image=1 (of which injected=0)` says the refusal happened to a transaction that arrived over **levin**, not one the harness handed in. `mined=0 conflict=1` says the twin did not leave because it was mined — its id is nowhere in block 175.

That window is a deliberate divergence, and the run declares exactly its size in its own arguments: `--m1-allow-uncompared 1`. It is also a **scenario** run (`--m1-min-txs 0`), which cannot claim parity coverage at all: the verdict rests on `no_disagreement()` plus the eviction it explicitly required, and a scenario that required no eviction FAILS rather than passing for free. Every parity run passes zero allowance.

## The defect this milestone found

**C3's relay gate was never opened.** `RelayedTxPool` refuses every transaction with `NotSynced` until told the index reached the tip — monerod's `is_synchronized()` rule, fail-closed so a node catching up cannot build a template on a pool it could not have populated correctly. M0 wired C2 → C3 for *block* events and left this seam open: nothing in the tree called `set_synced()`, so the gate stayed shut for the life of the process and the pool refused every transaction that ever reached it.

It is invisible from every angle M0 looked from. The tip follows. The peer is healthy. The status line's `txs=` counts frames coming *in*, and they do. The only symptom is a pool that stays empty, which on a quiet chain is also what success looks like. M1 is the first milestone that asks the pool what it **holds**, which is why M1 is where it surfaced. Fixed by `publish_tx_gate_()`, published from the verify thread (which owns the sync state) to the pool thread (which owns the gate) on a change only; the status line now carries `txpool: gate=` beside the counts.

## What lands

| file | what |
|---|---|
| `parity/xmr_txpool_parity.hpp` | the P-POOL seam: presence-preserving parse, per-tx judgement through the **same** `compare_seam` that judges P-TIP/P-TPL, and `TxpoolParityTally` (the coverage claim) |
| `contracts/parity.hpp`, `parity/xmr_parity_types.hpp` | `ProbeKind::Pool` + the frozen `POOL_FIELDS` table; `COMPARATOR_VERSION` **1 → 2** |
| `txpool/xmr_relayed_txpool.*` | `facts()` (the whole pool, under one lock, not the policy-filtered selection) and `rejected_key_image_conflict` |
| `node/xmr_native_node.hpp` | `txpool_parity_sample()`, `inject_relayed()`, `publish_tx_gate_()` |
| `node/tools/xmr_native_node_main.cpp` | `--txpool-parity`, `--inject-dir` (`.twin` = inject the key-image twin), the M1 gate flags, `M1-VERDICT` |
| `test/xmr_txpool_parity_kat.cpp` + `xmr_txpool_parity_golden.hpp` | the offline replay |

**The transaction id is the alignment key, not a compared field**, and that is the strength of the seam rather than a gap in it: a Monero v2 id is the triple Keccak over the three measured byte spans that make up the blob, so two pools reporting the same id are holding the same bytes — which turns "weight, fee and blob_size agree" from three coincidences into three independent derivations from one agreed input. `weight` and `fee` are revocation sentinels: they are what the assembler selects and pays on.

`COMPARATOR_VERSION` moves because the frozen table set changed: a node graduated by a comparator that never looked at the transaction pool must not inherit that clean streak into one that does.

## The KAT

`xmr_native_txpool_parity_kat` — **210 checks, 0 failures** — is the offline replay of the live run: a real daemon's `get_transaction_pool` answer is checked in, the KAT parses it with the same `MonerodTxpoolRpc`, feeds each entry's `tx_blob` into a real `RelayedTxPool` through `IRelayedTxSink`, and runs the same `compare_txpools`. Every native number is recomputed at test time from the transaction bytes; every monerod number is one the daemon printed. It also pins the judgement (a perturbed weight/fee/size FAILS, an unanswered required field FAILS rather than being skipped, disjoint pools produce NO samples rather than agreement), the coverage tally, both evictions, and the DoS buckets absorbing a relay flood without the pool admitting a frame the guard refused.

Registered in **both** `build.yml` target lists (the #1539 "***Not Run" lesson). Local `ctest -R "xmr_native|xmr_levin|xmr_epee"`: **25/25 passed**.

## Scope and what this run is not

Everything under `src/impl/xmr/native/` plus both build.yml target lists. **No `src/sharechain/v37` touch**, no consensus digest, no descriptor change.

The plan's M1 exit criterion is 100 % per-transaction equality over **≥10 000** relayed transactions on stagenet, plus a ≥720-block `already_generated_coins` / `median_weight` streak. This is the **regtest** proof of the same claim at two orders of magnitude less volume: the mechanism, the comparator, the coverage gate and both evictions, on a chain we control end to end. The volume and the streak need stagenet, a host and days — the same operator call M0's P-TIP soak is waiting on.


frstrtr:

PR #1584, opened 2026-09-11
ltc(pool): bound the inbound-share path so RSS stops climbing to the self-abort

frstrtr:
Produced by a design → implement → adversarial-review chain. **Draft: one review item is unaddressed (below) and two constants want a live measurement before they are trusted.**

## The defect, read from production

The LTC pool node on contabo self-aborts: `RSS LIMIT EXCEEDED === RSS: 4047 MB (limit: 4000 MB) === ABORTING (clean exit before OOM killer)`. It is a deliberate self-abort on `rss_limit_mb`, not a segfault and not an OOM kill. Aborts came at irregular intervals plus one burst of **eight in 75 seconds**. The limit was raised 4000 → 5000 as a stopgap; that masks the symptom.

**Not the cause:** #1562 fixed a use-after-free on the download-stop path. The claim that it cured these crashes was made and retracted after the abort reason was read.

## What this branch does

Bounds the inbound-share path: an admission budget (shares and bytes) taken before any work is posted; ownership of the `sharereply` payload made explicit so unadopted shares are freed rather than leaked; the raw-share cache wired to the chain's `on_removed` hook; the ancestor walk given an explicit depth predicate; and the per-peer remembered-tx byte cap restored to p2pool's behaviour.

## Review outcome

**APPROVE_WITH_FIXES. No blocking issues.** Reward safety traced rather than asserted: phase-2 acceptance logic is unchanged, the post-add release is a pointer-identity check against the object the tracker actually stored, a refused batch never touched the tracker or the chain, and the depth cap equals the pruning-zone rule the node already applies — so nothing beyond it could have entered a payout.

**Tests bite, shown by mutation:** reverting the ownership fix fails 5 cases; removing the adoption release produces a **double free** that aborts the suite; removing the cache-erase hook fails 2; removing the budget release fails 2. Unmutated: 165/165. Folded into the existing allowlisted `share_test` target — no new executable, so the "new target silently never builds" trap does not apply.

## What is NOT done

1. **The admission refusal defeats the existing drop-OLDEST policy.** Under the wedged shape this was written for, the budget is consumed by a few large deferred batches, so `m_pending_adds` never reaches its cap and the drop-oldest branch becomes unreachable — after which the *newest*, tip-extending batches are refused while stale ones sit queued. Not a reward change, but a convergence regression against the fix that preceded it. This wants the suggested LTC-lane remedy before merge.
2. **Two constants want a live measurement.** Whether 8192 shares / 128 MiB inflight fits under the live limit was not measured. The branch adds `ingest=` and `rawcache=` to the heartbeat next to `rss=` precisely so this can be read on the running node first.
3. **Still unbounded, reported not fixed:** a core-lane defect in `sharechain/share.hpp` (a throw during unserialize loses the pointer — every malformed share leaks), and `m_rejected_share_hashes`, an insert-only set with no erase path.
4. `main_ltc.cpp` still carries `rss_limit_mb = 4000`; the 5000 stopgap lives only in the deployed config.

No self-merge. Review and merge belong to the designated merger.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01UmtVGYzu1iRk1taczpz1gB

### DIFF
--- a/src/impl/ltc/ingest_budget.hpp
+++ b/src/impl/ltc/ingest_budget.hpp
@@ -1,0 +1,105 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+//
+// LTC inbound-share ADMISSION BOUNDS (memory, not consensus).
+//
+// Why this exists: the LTC pool node self-aborted on rss_limit_mb under an
+// ingest storm (contabo 2026-09-09..11). The only bound on the ingest path was
+// NodeImpl::MAX_PENDING_ADDS, which caps the number of DEFERRED BATCHES waiting
+// for the tracker lock. That cap is (a) downstream of the phase-1 verify-pool
+// backlog, which had no bound at all, and (b) a count of batches, not of shares
+// or bytes — one batch is one sharereply (up to 1000 shares) or one unsolicited
+// `shares` message (bounded only by the 32 MiB socket cap). Both holes let a
+// single peer push hundreds of MB of deserialised shares into the process
+// before MAX_PENDING_ADDS was ever consulted.
+//
+// IngestBudget bounds the ingest pipeline in SHARES and BYTES at the point of
+// admission (NodeImpl::processing_shares), i.e. BEFORE any scrypt job is posted.
+// The reservation is held by the owning HandleSharesData and released by its
+// destructor, so it covers the whole life of a batch: verify-pool backlog,
+// m_pending_adds queue, and phase 2.
+//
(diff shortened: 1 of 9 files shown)


frstrtr:
SUPERSEDED BY #1586 (bound the inbound-share path, keeping drop-OLDEST reachable). #1584 was CONFLICTING and ran zero CI; closing so the lane surface shows one live slice.

PR #1544, opened 2026-09-08
v37/xmr: flip XMR_BUILD_OPTION_B_KATS ON — run-coverage for the two option-B KATs (#1543)

frstrtr:
## What

Closes #1543. Flip `XMR_BUILD_OPTION_B_KATS` **ON** in CI so the two X9 option-B KATs get **run-coverage**, not the compile-only coverage #1539 shipped.

- `xmr_template_primitives_kat`
- `xmr_block_assembly_kat`

## How (deliberately tight)

1. **Configure** the non-sanitizer Linux leg with `-DXMR_BUILD_OPTION_B_KATS=ON` (added to the existing `Configure` step alongside `-DXMR_BUILD_RANDOMX=ON`).
2. **Build** both targets by appending them to that leg's `cmake --build --target` list. That step already runs under `-j"${BUILD_J:-8}"`, so the **#1522 per-runner `BUILD_J` cap is honored automatically** — no new `-jN` is introduced. These are logic / byte-identity KATs (no RandomX JIT), so the non-san leg is the right home; the ASan leg deliberately leaves the option OFF.
3. **Allowlist**: drop the two `# ci-allowlist-exempt:` lines in `src/impl/xmr/test/CMakeLists.txt`. The targets are now in the build.yml `--target` allowlist, so the drift-guard sees them registered (no longer exempt). Guard passes: `297 test target(s) across 11 test tree(s) all present`.

## Finding — run-coverage caught a wrong test golden (test-only fix)

The moment the primitives KAT actually ran it RED'd on a hardcoded oracle:

```
[FAIL] varint(600000000000) = 80e0a596bb11 (canon 80a094a58b11)
```

The hardcoded `"80a094a58b11"` decodes to `587146268672` — **not** 0.6 XMR. The library's four independent varint implementations (`writeVarint` vector + byte-sink, `tools::get_varint_data`, `BlobWriter`) all agree on `"80e0a596bb11"`, which decodes back to exactly `600000000000`. The block-assembly KAT's byte-identical monerod-shaped coinbase prefix independently confirms the **product** varint is correct. Fix is the **test literal only** — no product or consensus bytes change. Compile-only coverage had masked this latent wrong-golden; this is exactly what the run-coverage flip is for.

## Evidence — both KATs registered + passing (local ctest)

```
Test project /tmp/c2pool-1543/build_ci
    Start 1507: xmr_template_primitives_kat
1/2 Test #1507: xmr_template_primitives_kat ......   Passed    0.02 sec
[...]

PR #1585, opened 2026-09-11
xmr(native): M2 -- the K_fair settlement coinbase, built without a daemon [DO NOT MERGE]

frstrtr:
[DO NOT MERGE] DRAFT — M2 of the native-minimal Monero node. Stacked on `v37/xmr-native-m0` (PR #1580). Base commit d0ea24a0.

## What this is

M0 assembled the native node and proved it follows a live tip over levin. C4 landed the `IMinerDataSource` seam and both of its arms. Nothing connected them: the option-B settlement provider was still fed by monerod's `get_miner_data`, so the pool's blocks were assembled from a daemon RPC no matter what the native node knew.

This PR binds them and makes the two claims that matter falsifiable rather than asserted.

## The template path is native

`--xmr-template-source native` starts the embedded node (levin P2P + chain-state index + relayed txpool) and hands its serving arm to the **unchanged** `XmrSettlementTemplateProvider`. Height, prev_id, seed_hash, difficulty, median weight, already_generated_coins and the major version come out of the C2 index; the backlog comes out of the C3 pool.

monerod stays configured, because M2 deliberately keeps it as the C6 parity judge and as the block submit arm — but neither of those is the template path. The status line therefore reports the template path's **own** `get_miner_data` count next to the node's total RPC count, so "no daemon call on the template path" is a number that can be read and disbelieved. `--native-template-fallback off` removes the second answer entirely.

## The coinbase shape is gated, not assumed

A rebind cannot be trusted on the grounds that the assembler file did not move; what has to hold is a claim about the bytes. `xmr_settlement_coinbase_shape.hpp` re-parses the `miner_tx` out of the assembled block with `parse_coinbase_prefix()` — the same parser a peer uses on a block it did not build — and judges five properties:

1. it parses as a v2 coinbase for this height (`unlock == height + 60`);
2. `canonical_coinbase_matches(final inputs, parsed)` — the W3 every-node ACCEPT check, which byte-compares R, every amount, every one-time key, every view tag and the whole tx_extra;
3. the amounts **as parsed** sum to `base_reward + fees` to the piconero;
4. the owed outputs are in K_fair oldest-owed-first order and the mandated residual sink is last;
5. the tx_extra `0x03` merge-mining tag carries `mm_commitment_root(chain_id, owed_digest)` for the digest of the ledger the caller says it is settling.

The gate runs **before** a template is given an id, so a refusal parks miners on the last good block rather than handing them one that pays the wrong set. It is installed for the monerod arm too — the rebind is what made the check necessary, but the property it pins was always the one that mattered.

## One defect found and fixed here

The daemon arm's readiness is a property of its cache: `MonerodMinerDataSource` has nothing to report until `poll()` has filled it once. So the obvious rule — resolve the arm, then pump whatever was resolved to — can never reach it: the resolver finds it unready every time, forever, and the fallback is dead code. `ResolvedMinerDataSource::resolve()` therefore resolves the free (native) arm first and polls the daemon **only when it might actually serve**, which keeps the native-only pump count at zero while leaving the fallback reachable. Suite F of the KAT pins all three cases.

## Evidence

Isolated regtest on workstation-191: two peered `monerod 0.18.5.1` fakechain daemons on their own datadirs and ports (45188/45189, 45288/45289), the pool dialling levin from its own source address, chain grown past the 60-block timestamp window.

```
template source: NATIVE — the embedded Monero node (levin, 1 pinned peer(s)) feeds the
option-B assembler; monerod stays the parity judge and the submit arm, and is NOT on the
template path
  native: template arm READY, fallback OFF (native-only: no template is served if the
  native arm is not ready)

coinbase: outputs=2 (owed=1 fixed=0 sink=1) sum=35177628287959 budget=35177628287959
          (base=35177628287959+fees=0) K_FAIR-SHAPE=OK (gate ok=6 refused=0)
native:   arm=native resolves=121 template-path get_miner_data=0 | verified_frontier=100
          peers=1 txpool_accepted=0 | node rpc(parity+submit)=21
parity:   seam=TIP      h=100 verdict=CLEAN compared=9/9 [P-TIP first=native]
parity:   seam=TEMPLATE h=101 verdict=CLEAN compared=6/6 [P-TPL served_by=native]
```

6 templates at heights 97..101, every one shape-OK; **template-path `get_miner_data` = 0** across 121 arm resolutions with the fallback off; P-TIP CLEAN 9/9 and P-TPL CLEAN 6/6 at every height — the native arm served and the monerod shadow arm agreed on all six EQUALITY fields.

An earlier run on a 17-block chain is worth recording because the oracle caught it: P-TPL returned FAIL on the `median_timestamp_present` CONSTRAINT, because below height 60 the native arm's 60-block timestamp window does not exist yet. That is the constraint doing its job on a chain too short to satisfy it, not a template defect — the run above was repeated on a mature chain rather than the constraint being weakened.

## KAT

`v37_xmr_m2_native_settlement_kat` — **85 checks / 0 failures**, six suites:

| suite | what it pins |
|---|---|
| A | the native arm end to end: the coinbase's owed order, exact sum and owed_digest read back out of the block, plus the assembler's own structural battery at three extra nonces (45/45) |
| B | the two arms build the same block: coinbase prefix, MM root, tree root and `full_blob` from the miner_tx onward are byte-identical (the header carries a per-build timestamp and is excluded) |
| C | the native path never touches the transport (0 calls over 8 refreshes), with the daemon arm's 8 calls on the same script as the control that makes the zero mean something |
| D | readiness is fail-closed: no chain state ⇒ no template id is ever issued |
| E | the shape gate refuses a wrong owed_digest, is counted as a failure, and **no template id advances** |
| F | the resolver: native-only serves native with zero pumps; native-only with a dead arm serves nothing and does not quietly ask the daemon; with fallback on it serves monerod, counts the fallback, explains it, and the fallback template is still the K_fair shape |

Registered with `add_test` **and** added to BOTH `build.yml --target` lists (the #1539 lesson: a registered-but-unbuilt target reports `***Not Run` and fails ctest with exit 8).

`xmr_native_template_kat` still reports 157 checks / 0 failures; its chain-state fixture moved into `xmr_c4_state_builder.hpp` so the new KAT drives the same rolled-forward state instead of a second, independently rolled copy of a 100 000-entry window.

## Two pre-existing defects, disclosed and NOT fixed here

1. **Shutdown segfault on the option-B daemon.** `c2pool-v37-xmr --coinbase v37` segfaults during teardown, in `serve_and_run`'s final `status()` after `node.stop()`. **Reproduced on a pristine build of the base commit d0ea24a0 with none of this PR's code**, so it is not an M2 regression; it happens after "stopping…" and does not affect serving. It wants its own change.
2. **`xmr_template_primitives_kat` fails on this base** with `varint(600000000000) = 80e0a596bb11 (canon 80a094a58b11)` — a wrong test literal, product correct, already fixed on `v37-dev` by PR #1544. It only shows up locally because this verification ran with `-DXMR_BUILD_OPTION_B_KATS=ON`; the option is OFF by default and the target is not in CI on this base.

## Scope

14 files. `src/impl/xmr/native/` plus the consumer tree (`src/c2pool/v37/xmr/`, `src/c2pool/v37/test/`) plus `build.yml` and two CMakeLists. **Zero files under `src/sharechain/`.** No v37 consensus digest is defined; no descriptor kind-byte is added.

## Not claimed

The plan's M2 exit criterion (§6) is a 72-hour serve=monerod / shadow=native stagenet soak followed by a 72-hour flip, both P-TPL CLEAN. This is regtest, minutes, six templates. What is proven is the mechanism, the shape gate, the zero-daemon-call measurement and P-TPL agreement; **M2 as a milestone is not met** and this PR should not be read as M2 graduation.


frstrtr:
Superseded by c2pool#1588 (M2h — the same K_fair settlement coinbase restacked onto M1 @ v37/xmr-native-m1 so the native C3 txpool carries a real levin-relayed backlog, closing the empty-backlog finding). Closing this branch. The live native-node tower is M0 (c2pool#1580) -> M1 (c2pool#1583) -> M2h (c2pool#1588) -> M3 (c2pool#1591).

PR #1588, opened 2026-09-11
xmr(native): M2h — K_fair settlement coinbase with a real native backlog [DO NOT MERGE]

frstrtr:
**Stacked on `v37/xmr-native-m1` (c2pool#1583).** Do not merge before it.

M2 (c2pool#1585) was cut from M0 (`d0ea24a0`), not from M1 (`4d5cf22c`). That is
the whole of its HIGH finding **F-1**.

M1's real defect was that C3's relay gate, `RelayedTxPool::set_synced()`, had no
production caller — so the native pool refused every levin-relayed transaction
for the life of the process. M1 fixed it with `publish_tx_gate_()` and proved 67
transactions ingested over `NOTIFY_NEW_TRANSACTIONS`. M2, branched earlier, did
not carry that fix, so its native templates were honest about their own
construction and empty of work: `n_tx=0`, `fees=0`, at every height it served.
The M2 verifier's control test made that concrete — a wallet transaction reached
a control monerod peer in 5 s and the native pool held 0 of them over 75 s.

This branch replays M2's thirteen-file change on top of M1 and folds the review
findings.

---

## 1. The restack (`535f50f7`)

One conflict, `.github/workflows/build.yml`, where M1 and M2 each appended to the
same `--target` list; resolved as the union, both legs.
`src/impl/xmr/native/node/xmr_native_node.hpp` merged cleanly — M2's
`template_fallback` knob and `monerod_source()` accessor sit alongside M1's gate
machinery, `publish_tx_gate_()` intact and still called from both the sync-state
transition and the startup path. Nothing of M2's content is revised in that
commit, so what follows can be read against it.

## 2. THE ACCEPTANCE TEST — the exact test that falsified M2

Isolated regtest on workstation-191: two peered monerod 0.18.5.1 fakechain
daemons (A on `127.0.0.1:46188/46189`, **B the control peer** on
`127.0.0.2:46288/46289`), a `monero-wallet-rpc` against A, and
`c2pool-v37-xmr` dialling A over **levin** from `127.0.0.11`, built with
`XMR_BUILD_RANDOMX=ON`, `--native-template-fallback off`.

The wallet sent five fee-paying transactions, one of them with **three
destinations** so `weight != blob_size`:

```
71480224bcfd225b   fee 13296300000   weight 2829   blob_size 2369   <- multi-out
cce9fb41793a51f5   fee  2599200000   weight 2166   blob_size 2166
d52fef83c7606be2   fee  2599200000   weight 2166   blob_size 2166
f00d38cec83404c0   fee  2599200000   weight 2166   blob_size 2166
43e4301c96365f0f   fee  2599200000   weight 2166   blob_size 2166   <- the nudge
```

**The control peer B held all of them.** So did the native C3 pool
(`txpool_accepted=5`), over levin, with no RPC:

```
coinbase: n_tx=4 outputs=2 (owed=1 fixed=0 sink=1)
          sum=35197380292250 budget=35197380292250
          (base=35176286392250+fees=21093900000) K_FAIR-SHAPE=OK (gate ok=5 refused=0)
native: arm=native resolves=195 (native=195 daemon=0) template-path get_miner_data=0
        | verified_frontier=120 peers=1 txpool_accepted=5 backlog offered=5 selected=4
wire RPC to monerod: TOTAL=231 = template-path 0 + node(parity+submit) 36
                               + tip-poll 194 + pool-other 1 [failures node=0 pool=0]
```

The arithmetic closes exactly. `3 x 2599200000 + 13296300000 = 21093900000` — the
four transactions past the assembler's age gate — and
`35176286392250 + 21093900000 = 35197380292250`, the sum of the amounts as
re-parsed out of the assembled `miner_tx`. Zero `get_miner_data` on the template
path across 195 arm resolutions with fallback OFF, and `verified_frontier=120`
with RandomX compiled in and `--native-allow-unverified-pow` **not** passed, so
the frontier is really verified.

The fifth transaction is *correctly* absent: it was 5 s old at that rebuild and
the assembler applies monerod's own age gate. Which is the point of `offered=5
selected=4` being two printed numbers — see F-3.

**Scope, stated plainly: this is regtest and minutes. It is not the stagenet
milestone.** One pinned peer, `--native-force-synced`, a fakechain emission
curve, no reorg, no epoch edge, no adversarial peer, and no sustained soak. What
it establishes is that the wire is connected and the arithmetic is right, not
that the arm is production-ready.

### F-4 (MEDIUM) — P-TPL could not see an empty template

`tx_backlog_count` is a `Regime::Measurement`, so it is never scored. The six
required EQUALITY rows — `prev_id`, `major_version`, `difficulty`, `seed_hash`,
`median_weight`, `already_generated_coins` — are every one of them a property of
the **parent block**. A native arm serving `n_tx=0` forever, while the monerod
shadow arm held a full pool, therefore read **P-TPL CLEAN**. The detector was
standing next to the regression it existed to catch.

The count stays a Measurement — two pools legitimately differ at any instant and
an equality gate would fire on propagation skew. What becomes a **CONSTRAINT** is
the famine *shape*: `native == 0 AND shadow > 0`, sustained across **K
consecutive decidable samples AND S seconds**. Two bounds, because refresh bursts
would satisfy a count bound alone while observing one propagation window. An
undecidable sample (misaligned tips, an unanswered arm) neither grows nor clears
the streak — clearing on absence would let a flaky shadow arm hold it at zero
forever. The refusal is sticky until a sample shows the condition resolved.
`COMPARATOR_VERSION` 2 -> 3, so no streak earned by the old judge carries.

**Proven both ways**, in `xmr_native_backlog_famine_kat`:

- **Suite B** runs `compare_seam` *twice over the same pair of arm
  observations* — six EQUALITY fields, identical, agreeing — and gets CLEAN
  without the constraint and FAIL with it. Nothing else about the sample changes.
- **Suite C** drives the real `ParityOracle` over the pre-restack behaviour
  (native 0, shadow 5): with the guard off, all 8 samples CLEAN — the blind spot,
  reproduced. With it on, the run ends in refusal.
- **Suite D** drives the same oracle over the post-restack behaviour and requires
  silence, including the two cases a naive gate would false-positive on: holding
  1 of the shadow's 50 (skew) and a quiet chain where both pools are empty.

It also behaved correctly **live**: the evidence run opened with one `starved`
sample — the leftover daemon-pool transactions relayed before c2pool connected —
then `fed=3` and `streak=0/6`. Evidence accumulated, threshold not crossed, no
false refusal.

### F-3 (MEDIUM) — the RPC headline was a tenth of the truth

The line printed `node rpc(parity+submit)=N` from the embedded node's own
transport. The pool's consumer transport is a **second socket to the same
daemon**, and without libzmq its `pump_poll()` issues a `get_miner_data` on a
timer for the life of the process. An auditor with tcpdump would have seen ten
times the claimed traffic and read the gap as a false claim rather than a missing
summand. `LiveMonerodTransport` now counts every POST, split by driver, and the
terms add up to a TOTAL that matches the wire. Same discipline for the backlog:
`offered` and `selected` are printed separately, because one number would leave
`backlog=4` sitting next to a coinbase paying three fees with no way to tell a
filter from a bug.

### F-2 (SHOULD) — a pump count is not an arm count

`daemon_pumps()` answers "did the template path make a round trip?". It does not
answer "which arm built this?": with the daemon's cache warm and fresh the
resolver hands that arm back for free, the pump branch is never entered, and a
daemon-built template lands inside a run reported as `template-path
get_miner_data=0`. `ResolvedMinerDataSource` now also counts the latched arm by
pointer identity. Both numbers are printed, because neither implies the other.
Suite E pins the case: fallback to a warm daemon arm gives
`daemon_pumps()==0` **and** `served_by_daemon()==1`.

### F-2b (SHOULD) — readiness expires

`MonerodMinerDataSource` set `have_` on the first successful poll and never
cleared it, and `poll()` deliberately leaves the cache in place on failure so a
blip does not cost a template. A daemon that died at 09:01 left the arm reporting
all six readiness flags at 17:00 off an eight-hour-stale `miner_data` — and the
resolver gates fallback on `readiness().ok()`, so the pool would have fallen back
onto a dead tip. The cache now carries its fill time; `readiness()` and
`snapshot()` refuse past `max_age_ms` (default one Monero target block interval,
120 s). `epoch()` is untouched: it is the rebuild trigger, not a readiness claim.
Suite F ages a cache across the bound with an injected clock and back again.

## 4. One more defect, found while proving the above

With `backlog_refresh_s` at its shipped default of `0`, the served template is
rebuilt **only when the parent tip moves**. A block consumes the pool, so at the
instant of that rebuild the pool is empty, and everything arriving during the
block interval waits for the next tip move — by which time somebody else has
mined it. A tip-only native arm collects almost no fees *however well its pool
ingests*, which means M1's gate fix alone would not have produced revenue.

`--native-backlog-refresh <s>` admits the pool's `backlog_version` as a second,
rate-limited trigger. **The default does not move** (`0`, byte-identical to what
shipped), because a rebuild restamps the header timestamp under miners mid-grind;
the flag is what the evidence run turns on, and the trade-off is an operator's to
make.

## 5. Tests

New: `xmr_native_backlog_famine_kat` — **119 checks / 0 failures**, six suites.
Registered with `add_test` **and** in **both** `build.yml` `--target` lists (the
c2pool#1539 `***Not Run` lesson).

Unchanged and green on this branch:

| target | result |
|---|---|
| `v37_xmr_m2_native_settlement_kat` | 85 / 0 |
| `xmr_native_template_kat` | 157 / 0 |
| `xmr_native_parity_kat` | 833 / 0 |
| `xmr_native_txpool_parity_kat` | 210 / 0 |
| `xmr_native_node_kat` | 88 / 0 |
| `xmr_native_contracts_kat` | 392 / 0 |

Two **test literals** moved with the deliberate `COMPARATOR_VERSION` bump
(`xmr_parity_kat.cpp`, `xmr_txpool_parity_kat.cpp`). Both pins are exact rather
than floors on purpose — their job is to make anyone who edits a table move the
ledger key too — so they had to move, and they say why.

## 6. Known gaps and pre-existing defects

- **Not stagenet.** Regtest, one pinned peer, forced sync, minutes. The stagenet
  soak, reorgs, epoch edges and a real peer set are all still owed.
- **The age-gate hole this branch does not close.** A transaction filtered for
  being younger than 5 s is revisited only when the pool version moves again. On
  a busy chain that is continuous; on a quiet one the transaction can sit out
  several templates. Recorded for M3, not fixed here.
- **`--native-backlog-refresh` is opt-in and unset by default**, so the
  fee-revenue defect in section 4 is documented and reachable, not fixed. Whether
  the default should move is an operator's call.
- **Teardown `rc=139`** on the M2 rig is a pre-existing shutdown-order segfault,
  observed again here and not touched; it happens after the run and costs no
  evidence.
- **The option-B KAT wrong-literal** is already fixed on `v37-dev` and is not
  re-fixed here.
- **`--randomx` (the stratum submit path) was off** during the evidence run: the
  claim is about the template path, and the native node's own L4 verifier was
  live regardless (`verified_frontier=120`, no `--native-allow-unverified-pow`).

**Scope:** `src/impl/xmr/native/` and the consumer tree. **Zero files under
`src/sharechain/`.**


frstrtr:

PR #1591, opened 2026-09-11
xmr(native): M3 — daemonless FIND (native levin tip, switchable arm-order) [DO NOT MERGE]

frstrtr:
**DO NOT MERGE.** Stacked DRAFT on `v37/xmr-native-m2h` (`c2pool#1588`). Base = `v37/xmr-native-m2h` @ `f44886c8`.

M2h closed F-1 and proved the option-B K_fair settlement template is assembled from a real native backlog with no daemon call on the **template** path. What was still daemon-driven was the **find** path: `LiveMonerodTransport::pump_poll()` re-issuing `get_miner_data` on a timer, the X2 adapter's monerod-mirroring index answering the finalize driver's canonical test, and `submit_block` publishing the block. M3 cuts all three over, and makes which arm drives them a runtime switch.

## R-ARMORDER = SWITCHABLE (DaemonFirst stays the default)

`--arm-order daemon-first` (default) is byte-for-byte the M0..M2h posture. `--arm-order p2p-first` moves the whole find path onto the embedded node:

| | daemon-first (default) | p2p-first |
|---|---|---|
| tip | `get_miner_data` poll / ZMQ -> X2 MainchainIndex | native C2 index, levin + RandomX-verified |
| canonical test at maturity | the same monerod mirror | the same native index |
| publish a found block | `submit_block` | levin 2008 `NOTIFY_NEW_FLUFFY_BLOCK`, `ArmOrder::P2pOnly` |
| `rpc_on_find_path` | non-zero | **0** |

The switch is fail-closed on its own preconditions (`arm_order_refusal()`, a pure function so the daemon and the KAT refuse on the same code): p2p-first REFUSES option A and the monerod template source outright, and pins the template fallback OFF -- a daemonless find whose template came from `get_miner_data`, or that can quietly fall back to it, is not one. `--no-daemon-rpc` goes further and gives the embedded node no daemon endpoint at all.

## The two seams

* `NativeNode::drain_mainchain_events()` -- every event the native index flushes, queued on the verify thread and drained by the consumer's main loop at exactly the point where `pump_poll()` used to issue an RPC. A queue and not a callback: everything downstream of the finalize driver is main-thread-owned, and handing the verify thread a callback into it would put consensus bookkeeping behind a lock it has never taken.
* `NativeNode::block_relay()` -- C5, so the winning share can be published without an RPC. `o2::P2pBlockPublisher` is the `IShareSink` that does it, behind the SAME exact 128-bit RandomX gate as the daemon submitter.

`XmrNode` builds no X2 adapter at all in p2p-first, and REFUSES to start unless the canonical test was installed before `bring_up()`: the predicate is asked at maturity, a false answer orphans a perfectly canonical block, and there is no third value to return.

## The defect this found

C5 handed `IBroadcastPort` a **complete levin frame**, while the transport frames what it is given (`LevinLink::send_notify` -> `make_notify`). Every 2008 went out doubly framed, and every Monero peer answered

```
portable_storage: wrong binary format - signature mismatch
Failed to load_from_binary in notify 2008
```

and dropped it -- with `peers_written` reading perfectly the whole time, because the bytes really did reach the socket. Found only by relaying at a real monerod; no fake port could have shown it, because a fake port frames nothing. C5 now emits the message BODY, `contracts/broadcast.hpp` says so in those words, and both `xmr_native_block_relay_kat` and the new M3 KAT pin the epee storage signature (`01 11 01 01 | 01 01 02 01 | 01`) at the first byte of what the port is handed.

## Evidence (isolated regtest, own datadirs and ports, `XMR_BUILD_RANDOMX=ON`, RandomX mode=jit)

**p2p-first.** `rpc_on_find_path=0` in the program's own counters, and an independent TCP tap in front of monerod's RPC recorded `CONNECTIONS: 0` for the whole run -- not zero requests, zero sockets. The tip advanced 333 -> 366 entirely from levin events (`native tip feed: events=366 extend=366`). `p2p publish: relayed=5407 peers_written=5407 refused=0`. FOUND -> SETTLED ran to completion.

Block **335** on monerod's own chain -- and on an independent control daemon peered to it, so real Monero P2P carried it a second hop:

```
height 335  hash fcfa825b...  num_txes 3  reward 35169728911898  block_weight 6662
miner_tx outputs: 2
   amount 1000000000          <- K_fair OWED payee
   amount 35168728911898      <- exact-sum residual sink
```

which matches the pool's own gate line for that template:

```
coinbase: n_tx=3 outputs=2 (owed=1 fixed=0 sink=1)
          sum=35169728911898 budget=35169728911898
          (base=35161931311898+fees=7797600000) K_FAIR-SHAPE=OK
```

The three transactions are the ones a wallet relayed to monerod, which reached the native C3 pool over levin (`txpool_accepted=3`) and were selected by the native arm.

**daemon-first, same binary, same rig.** `rpc_on_find_path` climbing (2314 and rising), `submit calls=1036 ok=1036`, `id-source=monerod` on every FOUND, and the tap recording `CONNECTIONS: 2713` with `submit_block: 1195`, `get_miner_data: 275`. The default path is unchanged and demonstrably still daemon-driven.

## Tests

`v37_xmr_m3_arm_order_kat` -- **40 checks, 0 failures**. Both arm orders against one counting monerod transport: daemon-first must TOUCH the daemon (a zero there is the default path having broken, not M3 having worked); p2p-first must make EXACTLY ZERO touches across bring-up and a complete FOUND -> SETTLED driven by a pumped native chain. Plus the fail-closed rules, and the publisher refusing to push a FOUND event for a block that reached no peer. No sockets, no RandomX, no live daemon -- it builds and runs on BOTH build.yml legs, and is registered with `add_test` AND in **both** `--target` lists (the #1539 "***Not Run" lesson).

Regressions on the same build: `xmr_native_block_relay_kat` 98/0, `xmr_native_node_kat` 88/0, `v37_xmr_m2_native_settlement_kat`, `v37_xmr_node_smoke`, the C1/C2/C3 suites -- all green.

## Scope, honestly

* **Regtest and minutes is NOT the stagenet milestone.** M4 owns the parity soak; nothing here has run against a real network or for longer than a few minutes.
* Levin delivery is a DELIVERY, not an adjudication: no peer answers an unsolicited block notification. What backs the block instead is the chain -- a rejected block never reaches D_conf burial, the canonical test fails at maturity, and the settlement is disposed as an orphan. The daemon's synchronous yes is traded for the chain's asynchronous one, which is the trade a daemonless node is for.
* At regtest difficulty 1 the miner wins on nearly every nonce, so the run produced thousands of same-height blocks and a large orphan count; later blocks reported `reached_nobody` once the peer stopped taking them. An artifact of the rig's pace, not of the path.
* `header=MISMATCH` on the daemon-first FOUND records is the confirmation check racing a chain advancing hundreds of blocks a second at difficulty 1 -- the same artifact M2h's runs showed, not an M3 regression.
* Disclosed pre-existing, NOT fixed here: the teardown `rc=139` in `serve_and_run`'s final `status()` after `node.stop()` (it did not reproduce in p2p-first, which builds no adapter to tear down); the option-B KAT wrong-literal already fixed on `v37-dev` by `c2pool#1544`.

Canon untouched: **zero files under `src/sharechain/`**. Everything is under `src/impl/xmr/native/**`, `src/c2pool/v37/xmr/**`, `main_v37_xmr.cpp`, CMake and `build.yml`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -248,6 +248,7 @@ jobs:
                     xmr_native_backlog_famine_kat \
                     xmr_native_node_kat xmr_native_node \
                     v37_xmr_m2_native_settlement_kat \
+                    v37_xmr_m3_arm_order_kat \
             -j"${BUILD_J:-8}" 2>&1 | tee "${GITHUB_WORKSPACE}/ci-logs/build-tests.log"
 
       - name: Run tests
@@ -689,6 +690,7 @@ jobs:
                     xmr_native_backlog_famine_kat \
                     xmr_native_node_kat xmr_native_node \
                     v37_xmr_m2_native_settlement_kat \
+                    v37_xmr_m3_arm_order_kat \
             -j4  # ASan+UBSan: cc1plus RSS ~2-3GB each; -j8 wedged runner cgroup memory.high (throttle-hang). -j4 caps peak <12G.
                  # xmr_provenance_gate: previously omitted here because
                  # -fsanitize=undefined rejected its constexpr pointer comparison
--- a/src/c2pool/v37/main_v37_xmr.cpp
+++ b/src/c2pool/v37/main_v37_xmr.cpp
@@ -86,6 +86,8 @@
 #include "xmr/xmr_o2_settlement_provider.hpp" // O-2 option B: v37 K_fair settlement template provider + source
 #include "xmr/xmr_settlement_coinbase_shape.hpp"  // M2: the K_fair shape gate, read off the assembled block bytes
 #include "xmr/xmr_native_template_backend.hpp"    // M2: the native-minimal Monero node as the miner-data source
+#include "xmr/xmr_native_chain_source.hpp"        // M3: the native levin chain as the tip + canonical test
+#include "xmr/xmr_p2p_block_publisher.hpp"        // M3: found block -> levin 2008 (no submit_block)
 
 using namespace c2pool::v37n::xmr;
 namespace strat = ::v37::xmr::stratum;
@@ -172,8 +174,13 @@ static int run_mock_smoke() {
 template <class Provider, class Snapshot>
 class GatedShareSinkT final : public strat::IShareSink {
 public:
+    // `inner` is the PUBLISHER: sub::LiveSubmitShareSink (monerod submit_block)
+    // under --arm-order daemon-first, o2::P2pBlockPublisher (a levin 2008)
+    // under p2p-first. Taken as the INTERFACE rather than the concrete type so
+    // that the exact 128-bit gate in front of it is the same code on both arm
+    // orders -- the one property that must not fork when the publish path does.
     GatedShareSinkT(o2::O2RandomXVerifier& rx, const Provider& provider,
-                    strat::ITemplateSource& source, sub::LiveSubmitShareSink& inner)
+                    strat::ITemplateSource& source, strat::IShareSink& inner)
         : m_rx(rx), m_provider(provider), m_source(source), m_inner(inner) {}
 
     void on_accepted_share(const strat::AcceptedShare& s) override { m_inner.on_accepted_share(s); }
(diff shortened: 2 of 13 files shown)


frstrtr:

PR #1592, opened 2026-09-11
xmr(native): multi-node prefer-own same-height tiebreak [DO NOT MERGE]

frstrtr:
**DO NOT MERGE** — DRAFT, stacked on `v37/xmr-native-m3` (c2pool#1591, the daemonless node that finds and finalizes). Base is that branch, not `v37-dev`.

Closes the owed **c2pool#1551** item: *same-height double-block refusal -> prefer-own*.

## The situation

Two Monero blocks at one parent height. One carries our K_fair settlement coinbase and settles our ledger; the other is a stranger's — a competing template, a P2Pool aux block, a control peer's `generateblocks` — and settles us nothing. Until this branch the accounting had no answer for that at all. The F1 finalize driver credited whatever was buried and canonical at maturity, which is correct, and *nothing anywhere* recorded that the height had been contested, said which block we expected to be paid for, or checked the credit against a second opinion.

## The rule, and where it stops

The race is decided by **our benefit**, so every lever we control is biased toward our own block. There are exactly three and they are all local:

1. **What we build on** — the native fork choice adopts our own block at *equal* cumulative difficulty (D-14, `xmr_fork_choice.hpp`). Monero's rule there is first-seen; equal work means the network has not decided, so preferring our own tip is a choice inside the space consensus left open.
2. **What we re-announce** — a contested height is a relay race, so a small, bounded re-push of our own block.
3. **What we nominate** — the accounting says out loud, per height, which block it expects to be credited, so an operator sees a contest happening instead of finding out from a missing payout.

One flag (`--same-height-tiebreak`) drives all three, because two flags could disagree.

And there the bias stops:

- **(a) We cannot override Monero consensus.** If the network buries the rival, that is the chain.
- **(b) Credit requires burial.** A block is credited only once buried `D_conf` deep on the best chain. **No orphan-credit.**
- **(c) No double-credit (R-7).** A height credits at most one block, once; a later reorg neither mints a second credit nor moves the first.

## The shape that makes (a)/(b)/(c) checkable

`nominate()` is policy-**dependent**. `decide()` is policy-**independent** — a pure function of *(own candidates, other candidates, burial, canonicality)* with no policy argument to pass, so there is no setting a caller could reach for. The KAT runs one fixture under **both** tie-break policies and asserts the verdicts are byte-identical while the nominations differ. "Prefer-own never buys credit" is therefore a property that is run, not a sentence in a header.

The gate keeps its **own** verdict rather than doing the crediting, and `FinalizeConnect` cross-checks every settlement the F1 driver takes against it (`r7_violations`, which must stay at zero). A gate that also credited would agree with itself by construction and would prove nothing.

Alt-pool candidates are swept in alongside mainchain events, because the branch where our own block *stays best* produces no event for the rival at all — an accounting layer fed only by the event stream would report that height uncontested and credit as if it had run unopposed.

### The decision table (pinned by the KAT)

| own | other | buried | ours canonical | verdict | credit |
|----|----|----|----|----|----|
| 0 | 0 | – | – | `no-candidate` | none |
| 0 | >=1 | – | – | `other-only` | none |
| >=1 | * | no | – | `defer-unburied` | none (**the refusal**) |
| >=1 | * | yes | yes | `credit-own` | ours |
| >=1 | * | yes | no | `refuse-orphaned` | none (**no orphan-credit**) |
| * | * | * | * | `already-credited` | frozen (**R-7**) |

## What changed

| file | what |
|---|---|
| **new** `src/c2pool/v37/xmr/xmr_same_height_race.hpp` | `SameHeightRaceLedger`: the per-height race book, `credit_rule()` as a pure function, the nomination, the R-7 credit record, the bounded re-announce budget, the journal line |
| `src/c2pool/v37/xmr/xmr_o2_finalize_connect.hpp` | observes own FOUNDs and chain/alt blocks, runs the gate each tick, journals verdict **transitions**, cross-checks every credit |
| `src/c2pool/v37/xmr/xmr_node.hpp` | one chain-observation seam for **both** arm orders; `chain_carries()` so the gate asks the same oracle the finalize driver does |
| `src/c2pool/v37/xmr/xmr_node_config.hpp` | `--same-height-tiebreak` / `--same-height-renotify` / `--same-height-journal`; `d_conf` stays the single source of truth for the burial bar |
| `src/c2pool/v37/xmr/xmr_native_chain_source.hpp`, `xmr_native_template_backend.hpp` | alt-candidate feed; the D-14 tie-break plumbed from the same flag |
| `src/impl/xmr/native/chain/xmr_chain_index.hpp`, `xmr_fork_choice.hpp` | `alt_tips()` / `AltPool::for_each` — the candidates we hold but did not adopt; `ChainIndexOptions::tie` becomes a setting |
| `src/impl/xmr/native/relay/xmr_block_relay.hpp` | `renotify()` — re-announce a retained block once more (lever 2) |
| `src/c2pool/v37/test/v37_xmr_prefer_own_race_kat.cpp` | the KAT |
| `.github/workflows/build.yml` | the new target in **both** `--target` lists |

**Canon wall:** zero files under `src/sharechain/`. No consensus digest is defined or altered; this module sequences and audits calls the merged `OwedLedger` and F1 driver already own.

## Tests

`v37_xmr_prefer_own_race_kat` — **95 checks, 0 failed**, six suites:

- **A** `credit_rule()` exhaustively, including all 24 combinations where `already-credited` dominates
- **B** the ledger: observation, nomination, and credit identical under both policies at every high-water while the nomination differs
- **C** R-7: one credit per height, blocked replay, blocked move, frozen against a post-credit reorg
- **D** the re-announce lever and its five refusals (already ours / bias off / uncontested / buried / budget 0)
- **E** end to end through the real `XmrNode` + `FinalizeConnect` on a pumped native chain: ours-wins credits, ours-orphaned does not, a fork we were not in does not
- **F** two nodes, each holding its own candidate at one contested height: the winner credits once, the loser fabricates nothing, the fleet-wide credited set has exactly one entry

Registered with `add_test` **and** in both `build.yml --target` lists (the #1539 `***Not Run` lesson). Siblings re-run green: `v37_xmr_o2_finalize_connect_selfcheck`, `v37_xmr_m3_arm_order_kat`, `v37_xmr_node_smoke`, `xmr_native_chain_index_kat`, `xmr_native_block_relay_kat`, `xmr_native_node_kat`. Built with `XMR_BUILD_RANDOMX=ON`, g++-15, `cppstd=20`.

## Live proof (isolated regtest, minutes — NOT a stagenet milestone)

Rig: two **partitioned** regtest monerods (neither dials out; they do not know each other), a **control peer** that dials both and whose presence is the merge and whose absence is the partition, and **two daemonless c2pool nodes** (`--arm-order p2p-first`, `--coinbase v37`, `--xmr-template-source native`, `D_conf=3`, distinct settlement payees). `rpc_on_find_path=0` on both nodes throughout.

**Race 1 — two c2pool nodes, each with its own settlement block at h=31.** Partitioned, each node mined and relayed its own block; monerod A kept `f3ddad…`, monerod B kept `1d787e…`, same height, different tips. Both nodes refused to credit anybody while unburied. The control peer merged the partitions and the A branch was extended, so the network buried `f3ddad…`:

```
node1  h=31 verdict=defer-unburied   hw=30 need=34 own=5 other=0  credit=-
node1  h=31 verdict=already-credited hw=37 need=34 own=9 other=0  credit=f3ddad7bc211…
node2  h=31 verdict=defer-unburied   hw=30 need=34 own=1 other=0  credit=-
node2  h=31 verdict=refuse-orphaned  hw=37 need=34 own=6 other=1  credit=-        <- own-vs-other
```

Node 2's own view of the contest, with prefer-own still nominating its own block while refusing to credit it:

```
race h=31 refuse-orphaned [own=6 other=1 OWN-vs-OTHER] hw=37 need=34
          nominated=1d787ed1b6d1… (ours) :: prefer-own: our settlement block over the rival
contested h=31: 1d787ed1b6d1(ours) 336e53cdc989(ours) 78dd0d9ec94e(ours)
                8fbe2ab6afa8(ours) a1b4ea41ad7e(ours) fbea75d4ef7b(ours) f3ddad7bc211(theirs)
```

**Race 2 — our block against a block with no c2pool coinbase, and theirs wins.** Partitioned again; node 1 mined at h=44 on one side while the other side generated a plain monerod block. The aux branch was made heavier and the merge buried **theirs**:

```
node1  h=44 verdict=defer-unburied  hw=44 need=47 own=3 other=1  credit=-
node1  h=44 verdict=refuse-orphaned hw=48 need=47 own=3 other=1  credit=-
contested h=44: c88ae59342e4(ours) ce298a04b93b(ours) f827ac1f50a1(ours) 09575036856b(theirs)
```

All three monerods agree `h=31 -> f3ddad…` and `h=44 -> 09575036…`. Across both nodes' journals the fleet credited **one height, one block** — the one Monero buried — and `R-7 violations = 0` on both nodes in both races.

## Honest scope and disclosures

- **Regtest and minutes is not a stagenet milestone.** `D_conf=3` and `--fixed-difficulty 1` make a same-height race happen on demand in seconds; on stagenet `D_conf=60` and real inter-block times, a same-height race is rare and a reorg deeper than `D_conf` is rarer still. This branch proves the decision table and the two bounds on a real chain with real RandomX-verified blocks between real levin peers — it does not prove anything about stagenet cadence, and nothing here has been run against mainnet value.
- **Found in the base, disclosed, not fixed here:** a **running** native node did not follow the winning branch after losing a same-height race. It parked the rival branch and stayed on its own tip for several minutes; both races needed a node restart (a fresh sync from genesis) before the node adopted the buried chain. The finalize/refusal path is correct either way — the restarted node re-drove its pending FOUNDs from the sidecar and refused them at maturity, which is what the journal above records — but the reorg-follow path in `M3`'s sync driver needs its own look. It is out of this change's blast radius (the tie-break is accounting; the gap is chain sync) and is tracked for the next batch rather than patched blind here.
- **The race book's credited map is in-memory.** Across a restart the R-7 guard is the merged `OwedLedger`'s durable `is_settled()`, not the race book; the book re-observes and re-decides. That is correct today because a settled bid can never re-enter the pending set, but a durable credited map belongs with the W6 store and is a follow-on.
- **Under `--arm-order daemon-first` there is no re-announce lever** (the block went out through monerod's `submit_block`, and re-submitting a block monerod already holds is a no-op) and the X2 adapter keeps no alternatives, so `alt_candidates` is unset. Both are stated in the code rather than papered over; the gate still runs off the mainchain event stream.
- **CI on this PR is the two gates only.** `build.yml` triggers on PRs into `master` / `btc-embedded` / `dash-spv-embedded` / `dgb/**`; a PR based on `v37/xmr-native-m3` gets no build matrix, so `attribution-gate` and `reap-guard-honesty` are the only checks that run here. The build/test evidence above is the local run on the build host (g++-15, `XMR_BUILD_RANDOMX=ON`, `cppstd=20`), and the new target is registered in both `--target` lists so it is covered the moment this stack reaches a branch CI does watch.
- **Pre-existing, untouched:** teardown `rc=139` seen on other XMR rigs; the option-B KAT wrong-literal already fixed on `v37-dev` (c2pool#1544).
- The rig stood up its own monerods, datadirs and ports and was fully torn down; the `.40` observer and the `.44` stagenet daemon were never touched.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -248,7 +248,7 @@ jobs:
                     xmr_native_backlog_famine_kat \
                     xmr_native_node_kat xmr_native_node \
                     v37_xmr_m2_native_settlement_kat \
-                    v37_xmr_m3_arm_order_kat \
+                    v37_xmr_m3_arm_order_kat v37_xmr_prefer_own_race_kat \
             -j"${BUILD_J:-8}" 2>&1 | tee "${GITHUB_WORKSPACE}/ci-logs/build-tests.log"
 
       - name: Run tests
@@ -690,7 +690,7 @@ jobs:
                     xmr_native_backlog_famine_kat \
                     xmr_native_node_kat xmr_native_node \
                     v37_xmr_m2_native_settlement_kat \
-                    v37_xmr_m3_arm_order_kat \
+                    v37_xmr_m3_arm_order_kat v37_xmr_prefer_own_race_kat \
             -j4  # ASan+UBSan: cc1plus RSS ~2-3GB each; -j8 wedged runner cgroup memory.high (throttle-hang). -j4 caps peak <12G.
                  # xmr_provenance_gate: previously omitted here because
                  # -fsanitize=undefined rejected its constexpr pointer comparison
--- a/src/c2pool/v37/main_v37_xmr.cpp
+++ b/src/c2pool/v37/main_v37_xmr.cpp
@@ -451,6 +451,40 @@ static int serve_and_run(const XmrNodeConfig& cfg, LiveMonerodTransport& transpo
                     static_cast<unsigned long long>(fs.orphaned),
                     static_cast<unsigned long long>(fs.refused),
                     fc.pending().size());
+        // c2pool#1551. r7=0 is the claim that matters: it counts settlements the
+        // same-height gate did not authorise, which is the only shape an
+        // orphan-credit or a double-credit can take.
+        {
+            const auto& rs = fc.race().stats();
+            const auto contested = fc.race().contested_heights();
+            std::printf("  same-height: tiebreak=%s D_conf=%llu | own=%llu other=%llu "
+                        "contests=%llu (own-vs-other=%llu, open now=%zu) | credited=%llu "
+                        "refused[orphaned=%llu other-only=%llu] deferred=%llu renotified=%llu "
+                        "| R-7 violations=%llu double-credit blocked=%llu\n",
+                        to_string(cfg.same_height_tiebreak),
+                        static_cast<unsigned long long>(cfg.d_conf),
+                        static_cast<unsigned long long>(rs.own_observed),
+                        static_cast<unsigned long long>(rs.other_observed),
+                        static_cast<unsigned long long>(rs.contests_opened),
+                        static_cast<unsigned long long>(rs.own_vs_other_opened),
+                        contested.size(),
+                        static_cast<unsigned long long>(fs.race_credited),
+                        static_cast<unsigned long long>(fs.race_refused_orphaned),
+                        static_cast<unsigned long long>(fs.race_refused_other_only),
+                        static_cast<unsigned long long>(fs.race_deferred),
+                        static_cast<unsigned long long>(fs.race_renotified),
+                        static_cast<unsigned long long>(fs.r7_violations),
+                        static_cast<unsigned long long>(rs.double_credit_blocked));
+            for (const std::uint64_t h : contested) {
+                const auto* cs = fc.race().candidates_at(h);
+                if (!cs) continue;
+                std::string line;
+                for (const auto& c : *cs)
+                    line += " " + c.bid.substr(0, 12) + (c.own ? "(ours)" : "(theirs)");
+                std::printf("    contested h=%llu:%s\n",
+                            static_cast<unsigned long long>(h), line.c_str());
+            }
+        }
         std::printf("  %s\n", rx.describe().c_str());
         const std::string g = sink.last();
         if (!g.empty()) std::printf("  gate: last=%s\n", g.c_str());
@@ -587,6 +621,11 @@ static std::unique_ptr<o2::NativeTemplateBackend> start_native_backend(const Xmr
     ncfg.c2pool_commit        = C2POOL_VERSION;
     ncfg.ready_timeout_s      = cfg.native_ready_timeout_s;
     ncfg.backlog_refresh_s    = cfg.native_backlog_refresh_s;
+    // D-14 lever (1): what the fork choice adopts at EQUAL work at the same
+    // height. Same flag as the accounting tiebreak -- see xmr_same_height_race.hpp.
+    ncfg.fork_tie             = (cfg.same_height_tiebreak == SameHeightTieBreak::PreferOwn)
+                                    ? ::c2pool::xmr::native::TieBreak::PreferOwn
+                                    : ::c2pool::xmr::native::TieBreak::FirstSeen;
 
     std::printf("template source: NATIVE — the embedded Monero node (levin, %zu pinned peer(s)) "
                 "feeds the option-B assembler%s\n",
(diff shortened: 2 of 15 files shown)


frstrtr:
Landed via #1612 (daemonless tower M0..M4, merged to master): every file this PR touches — xmr_fork_choice.hpp / xmr_peer_pool.hpp / xmr_same_height_race.hpp and their KATs — is present on master, which only carries further additions. Closing as superseded. Branch kept (no deletion).

PR #1593, opened 2026-09-11
xmr(p2pool): read-only sidechain observer [DO NOT MERGE]

frstrtr:
## What this is

A **strictly read-only** client for the live P2Pool (SChernykh/p2pool) Monero
sidechain, stacked on `v37/xmr-native-m3` so the Monero block template embedded
in every P2Pool sidechain block is read by the **same** parse and identity code
the native node runs against monerod.

P2Pool is the only large, live, adversarial instance of the problem the XMR lane
is solving. This tree watches it: how often two miners land on the same height,
how wide that window is, how the share set moves, how a template tracks the
Monero tip. It is telemetry — a calibration input for the prefer-own tiebreak
steer — and it participates in nothing.

`src/impl/xmr/p2pool/**` is a new directory. **Zero files under
`src/sharechain/`**; no consensus is activated and no v37 digest is touched.

## Read-only is a shape, not a flag

Seven of the protocol's twelve message ids publish or announce
(`BLOCK_RESPONSE`, `BLOCK_BROADCAST`, `BLOCK_BROADCAST_COMPACT`,
`MONERO_BLOCK_BROADCAST`, `AUX_JOB_DONATION`, `BLOCK_NOTIFY`, `LISTEN_PORT`).
**This tree can encode none of them.** `p2pool_wire.hpp` declares

```cpp
enum class ControlMessage : std::uint8_t {
    HandshakeChallenge, HandshakeSolution, BlockRequest, PeerListRequest
};
```

and one `encode()` total over it. There is no `PoolBlock` serialiser anywhere
under `p2pool/` and no `bind()`, so there is no publish path and no inbound
path. The KAT pins the emitted message-id set at `{0, 1, 3, 6}`; the live tool
prints the same thing as a per-message byte ledger from the run itself.

`LISTEN_PORT` is the deliberate omission — sending it is what puts an address
into every peer's list and onward into the network's own directory. The
documented cost is that peers then never send us broadcasts
(`P2PClient::is_good()` gates `on_broadcast()`), so the observer **pulls**:
`BLOCK_REQUEST` for the tip, then a parent walk. That has one hard timing
constraint taken from upstream's own idle sweep — a peer bans a connection that
has not broadcast in 900 s *unless* it requested a block within the last 10 — so
the tip poll is **5 s** and is not a tunable knob.

## Live proof — real P2Pool MINI sidechain, from .191, outbound only

Not regtest. Not a fixture. Live network observation, 420 s window:

```
p2pool-observer[mini] READ-ONLY
  peers        : connected=6 known=253 gossiped=412
  sidechain    : tip_height=14757264 tip_id=a7c93f6ec766af6d27c6c722a510d8499b753f836f29144cb3355b502f11fe84
  difficulty   : tip=248151826 cumulative=2443546641127031 implied_hashrate=24815182 H/s
  cadence      : 9.8504 s/height over 41 LIVE heights above frontier 14757213 (target 10 s)
  blocks       : distinct=106 height_span=113 duplicates=439 parse_failures=0 broadcasts=0
  same-height  : contested=2 uncles_named=3 uncles_resolved=2
  monero       : template_heights=6 highest=3760085
  outbound     : challenge=7 solution=6 block_req=550 peerlist_req=29 bytes=18544
  contested h=14757198 blocks=2 spread_ms=1
      f5a8f38318c82476f036e854f937ded3084db32a7930e2b8c378114b62ca9fe8
      5c1379aaa6d3e675959c2553ff7ff3508c52cc6f420ffcb489d693159e267e24
  contested h=14757200 blocks=2 spread_ms=0
      90f7c704f429d3d5be163901bdde4eaf10ed66fb723c2c016f7946dd1a87a244
      323a2ce3748013ff266096cd3b883863188876016b7220e0daab7fb72b235b4c
  monero xcheck: height=3760084 hits=300 misses=0
```

Six real peers, each authenticated by the mini consensus id in the handshake:
`184.16.119.208`, `141.126.51.200`, `91.122.226.236`, `46.128.213.183`,
`135.181.76.219` and `seeds-mini.p2pool.io`, all on port 37888.

Specific parsed sidechain blocks, ids and heights over the wire:

| sidechain height | sidechain id | Monero template height | PPLNS payout lines | txs |
|---|---|---|---|---|
| 14757167 | `4443f87ad2e604d6d7c92eae496ea9cbae189413d298f0e2bc070c7221665f6d` | 3760080 | 688 | 91 |
| 14757139 | `f11e43b57722af45ea9f415360753471f852bc07ec661cdd9d8f27e81cea31fb` | 3760078 | 690 | 152 |
| 14757216 | `86af7b2bd92dae7b…` (carried 1 uncle) | 3760084 | 694 | 21 |
| 14757264 | `a7c93f6ec766af6d27c6c722a510d8499b753f836f29144cb3355b502f11fe84` | 3760085 | — | — |

Every sidechain id above was **recomputed** from the blob
(`keccak256(bytes with nonce/extra_nonce/mm-root zeroed ‖ consensus_id)`), not
read off the wire — and the recomputation is confirmed by the network's own
linkage: `4443f87a…`'s `parent` field, straight off the wire, is
`5249d3f80d397f1e…`, the id this same derivation produced for the previous
block of the same session.

**Cross-check against a public source.** Each block carries the id of the Monero
block below its template. The run asserted every block templated on Monero
height 3760084 carries `e468f18d7628ec96dc968dffebc4d50c325a46e1216406a4fc0c59d1f3c62fe6`,
the id a public Monero source reported for height 3760083 — **300 hits,
0 misses**. Earlier runs anchored the same way on 3760079 (`4a051a76…19dd0`) and
3760077 (`79f93953…53a56`); both are pinned by name in the KAT.

An earlier MAIN-sidechain run (port 37889) validated the same code path:
heights 15199524–15199536 at difficulty ≈4.44e9.

### A correction worth recording

"mini is the fast chain" is not true. `default` and `mini` both set
`target_block_time = 10` and `min_difficulty = 100000`; they differ only in the
pool name that generates the consensus id and in the hashrate that shows up,
which moves the **difficulty**, not the cadence. `nano` is the one with a
different cadence (30 s). The KAT pins all three.

## Two real defects found by running

**1. P2Pool's hash order is not byte order.** Upstream's `struct hash`
(`src/common.h`) compares 32-byte ids as four little-endian words, **most
significant first**, and the merge-mining extra list is emitted from a
`std::map<hash, …>`, so it arrives in that order. The first cut used
`std::array`'s lexicographic `operator<` for the "ids must increase" duplicate
check and rejected a steady trickle of perfectly legal live blocks
(`BadSidechain`, ~1 in 6). The counter-example, kept by the observer's
`--dump-failed` path, is mini block `f11e43b5…ea31fb` at height 14757139:

```
40b20e14c59edc3257b171b0f37627018e9245edd52a690bf6d9e621a098b96a
01f0cf665bd4cd31cbb2b2470236389c483522b350335e10a4a5dca34cb85990
```

descending lexicographically, ascending in P2Pool's order. It is **golden B** in
the KAT; reverting the comparator makes the KAT fail with exactly the live
error (verified by mutation: `checks=208 failures=4`).

**2. Backfill was poisoning the cadence.** A pulling observer fetches the tip
then walks parents, so most heights were *received* seconds apart though *mined*
ten seconds apart — the first run reported **4.4 s/height** on a chain targeting
10, a number that looks like a finding and is an artefact of our own fetch rate.
Cadence is now measured only over the **live window**: heights strictly above
the frontier (the tip that already existed when the run started). The rerun
reports **9.85 s over 41 live heights**.

## What is reused from M3

A sidechain block is a complete Monero block blob with P2Pool's record welded
onto the end, so the observer finds the boundary and hands bytes
`[0, sidechain_offset)` to `native/consensus/xmr_block_parse.hpp` and
`xmr_block_id.hpp` — the same coinbase parse, tree hash and length-prefixed
block id the native node runs against monerod. Every read goes through Wave 0's
bounds-checked `BlobReader`. The two readings of the shared prefix (timestamp,
prev_id, nonce) are cross-checked at runtime and a disagreement is loud.

RandomX is **not** run: a sidechain block is not a Monero block unless its PoW
also cleared the Monero network target, and an observer holding no Monero chain
has no target to judge against. The status line says so rather than claiming a
"blocks found" number it cannot support.

## KAT

`xmr_p2pool_parse_kat` — **253 checks, 0 failures**, offline (both goldens
embedded as hex; no network, no daemon, no fixture file). The goldens are
complete `BLOCK_RESPONSE` frames captured from the live session with nothing
stripped, and their expected values are generated by
`test/gen_p2pool_golden.py`, an **independent Python reading** of the same bytes
with its own keccak, varint decoder and CryptoNote tree hash — so the KAT is a
cross-implementation agreement, not a photograph of the C++ output.

Covers: frame splitting (every id's exact length, bad id fatal, oversize fatal),
the full parse against both goldens, the embedded Monero block through the
native lane, the wrong-consensus-id rejection, P2Pool hash order, the four-value
write surface and its exact byte lengths, the handshake and its PoW bound
against the 128-bit product, every truncated prefix of a real frame, a tampered
byte changing the id, and the read model's contest/cadence/uncle logic.

Registered with `add_test` **and** added to **both** `--target` lists in
`.github/workflows/build.yml` together with `xmr_p2pool_observer` (#1539
lesson). Verified through a real CMake configure + build on gcc-15:
`ctest -R '^xmr_p2pool_parse_kat$'` → `100% tests passed`.

`xmr_p2pool_observer` is a harness in the `xmr_native_node` mould: CI **builds**
it so it cannot bit-rot and **never runs** it, because running it dials a public
network. It registers no ctest test.

## Scope and honesty

- No fork choice, no PPLNS ledger: both would need re-deriving P2Pool's
  consensus or holding its full window, and a partial answer that looks
  authoritative is worse than none.
- "Mainnet blocks P2Pool found in the window" is **not** reported as a number —
  see the RandomX note above. What is reported is the count of distinct Monero
  template heights crossed.
- IPv6 peers are skipped by the dial queue: the run host has no IPv6 route.
- Compact/pruned broadcast blobs are parsed, but their ids are taken from the
  wire and flagged `unverified`, because rebuilding elided outputs needs the
  PPLNS window this observer deliberately does not keep. A pulling observer
  receives neither shape in practice (`broadcasts=0` in the run above).

**DO NOT MERGE** — draft for review, stacked on `v37/xmr-native-m3`.


frstrtr:

PR #1594, opened 2026-09-11
xmr(native): M4 parity-soak harness + GraduationLedger (regtest-proven, stagenet-ready) [DO NOT MERGE]

frstrtr:
**DO NOT MERGE.** Stacked on `v37/xmr-native-m3` (c2pool#1591). Held DRAFT for review.

## What this is

M4's exit criterion is not a property of a function. It is a property of a
*process*: `serve=monerod/shadow=native` for a sustained window with every
sample CLEAN, then `serve=native/shadow=monerod` for another, and only then is
the daemon demotable. C6 (c2pool#1591's ancestry) built the judge and a
per-seam ledger. This builds the loop around them, and the artefact the gate is
read off.

Two things turned out to be **missing rather than merely unwritten**, and both
would have made a soak on the pre-M4 code report a claim it had not tested:

* **Nothing ever called `on_serve`.** The tip probe fires by itself, because the
  chain index publishes an event at every height. The template probe only fires
  when somebody *serves* a template — and in this harness nobody did. A soak run
  before this change would have accumulated a long P-TIP streak having **never
  once compared a template**, while "P-TPL CLEAN on every sample" is half of
  what M4 asks for. `SoakDriver` is that caller; `SoakThresholds::
  min_template_clean` is what stops the cheap seam carrying the expensive one.
* **The posture had no way to be exact.** `ArmResolver::serving()` falls back to
  the other arm when the configured one is not ready — correct in production,
  fatal to a parity claim, because a leg that quietly served monerod for an hour
  is not evidence about the native arm. `NativeNode::m4_serve_probe()` resolves
  `arm(serve_arm)` strictly; an unready arm yields no sample and says why.

## The ledger

`M4GraduationLedger` records two legs. A leg is a streak, fragile by
construction:

| verdict | effect on the streak |
|---|---|
| CLEAN | accrues: one sample, possibly one block, more wall clock |
| FAIL | **resets to zero** and takes the leg's qualification with it |
| SERVED-MISMATCH | resets **and revokes the key** |
| VOID | neither accrues nor resets… |
| VOID × `max_void_run` | …but a *run* of them resets anyway |

There is no averaging, no "mostly clean", no budget of allowed failures: the
criterion is every sample. The void-run rule is the DASH lesson at its sharpest
— a leg that has stopped producing judgeable samples is a probe that stopped
firing, not a streak quietly holding, and letting it coast would make silence
indistinguishable from agreement.

`blocks` counts **forward height progress only**, so a reorg contributes none
(the conservative direction: it can only make a 720-block claim harder to earn).
The key is C6's four-part `{commit, comparator_version, monerod_version, net}`
and is stamped with the **running** `COMPARATOR_VERSION` — so a streak earned
under an older comparator cannot be inherited across a bump. That is the M1
lesson made mechanical rather than remembered, and the KAT pins it by rewriting
a graduated ledger's version by one and watching a v3 binary refuse it. The
state on disk is treated as a *cache of a decision* and re-derived from the
counters on load, so a file claiming `Graduated` over legs that did not qualify
is corrected rather than believed.

The two postures are **two runs of the binary against the same `--m4-ledger`
file** — the shape a 72 h + 72 h stagenet soak has. Persistence is not an
optimisation here; it is the mechanism by which a leg survives its own posture
flip.

## The negative control

A harness that has never been seen to refuse is not a gate.
`TipFaultInjection` / `PerturbingTipObserver` decorate the native arm's tip
observer, are disarmed unless a field is named, and perturb one required
EQUALITY field by **one unit** (or withhold it) — the same mutation
`xmr_native_parity_kat` pins offline, applied **live**, through the real
observer, the real comparator and the real ledger. `--m4-require refusal`
asserts the opposite claim from a normal run: the ledger must **not** graduate,
and it must have reset a streak on a real FAIL rather than merely failed to
accumulate one.

## Regtest mini-soak — what was actually run

Isolated rig on workstation-191: own datadirs, own ports (48488/48489 and
48588/48589), two `monerod v0.18.5.1-release` in `--regtest --fixed-difficulty
1`, never peered to each other. Nothing shared with the `.40` observer, the
`.44` stagenet daemon, or the CI heavy runners. Every daemon stopped afterwards.

Thresholds for the rehearsal: `clean>=24 blocks>=6 seconds>=40 tip>=6 tpl>=12
void_run<=16`. Build: `g++-15`, `-DXMR_BUILD_RANDOMX=ON`, so the L4 RandomX
verifier is live (`rx: mode=1`, `foreign=0` throughout).

Seven legs, each a separate run of `xmr_native_node` (`--run-seconds 150`, one
block every 6 s, P-TPL cadence 1.5 s). All seven exited `M4-RUN: PASS` — for the
clean legs that means "graduated as required", for the control legs it means
"refused as required".

| leg | posture | streak | blocks | span | P-TIP / P-TPL | fail | resets | ledger |
|---|---|---|---|---|---|---|---|---|
| legA | serve=monerod | 105 | 17 | 147 s | 17 / 88 | 0 | 0 | Observing, `legs_qualified=1/2` |
| legB | serve=native | 114 | 17 | 147 s | 17 / 97 | 0 | 0 | **GRADUATED**, `legs_qualified=2/2` |
| nc1a | serve=monerod | 110 | 17 | 147 s | 17 / 93 | 0 | 0 | Observing, leg qualified |
| **nc1** | serve=native, `reward` +1 unit after 10 observations | 11 | 0 | 14 s | 0 / 11 | **7** | **7** | Observing, **NOT GRADUATED** |
| nc1sa | serve=monerod | 117 | 17 | 146 s | 17 / 100 | 0 | 0 | Observing, leg qualified |
| **nc1s** | serve=native, `cumulative_difficulty` +1 unit (a **sentinel**) | 11 | 0 | 13 s | 0 / 11 | **7** | **7** | **Revoked** |
| **nc2** | serve=native, judge = the **divergent** daemon | 0 | 0 | 0 s | 0 / 0 | 0 | 0 | Observing, `void=115`, **NOT GRADUATED** |

The two postures of the clean run are two processes against one ledger file, and
the second one is what flipped it:

```
M4-VERDICT: GRADUATED posture=serve-native state=Graduated streak=114 blocks=17
            span=147s tip=17 tpl=97 clean=114 fail=0 mismatch=0 void=3 resets=0
            qualified=1 legs_qualified=2/2
```

```json
"key": {"c2pool_commit": "...-m4harness", "comparator_version": 3,
        "monerod_version": "0.18.5.1-release", "net": "regtest"},
"state": "Graduated", "graduated_unix": 1789127901
```

**nc1 is the required negative control and it is not vacuous.** The
`serve-native` leg soaked 110 CLEAN samples and *qualified*; then one required
EQUALITY field of the native tip observation was perturbed by one unit, and the
real comparator failed the sample:

```
last_reset_why: "P-TIP FAIL at height 184: 1 field(s) differed; [P-TIP first=native]"
revocation    : "posture serve-native lost its qualification:
                 P-TIP FAIL at height 178: 1 field(s) differed"
qualified     : false      resets: 7      graduated: NO
```

nc1s shows the sentinel path: `REVOKED: sentinel field 'cumulative_difficulty'
differed: 0:218 vs 0:217` — the one-unit perturbation, visible in the reason.
nc2 is the control with **no injection at all**: the judge is a second regtest
daemon at the same height on a different chain, every sample is VOID on the
prev_id alignment key, the streak never starts, and the ledger refuses.

`randomx: mode=1 foreign=0` throughout — the L4 verifier was live on the verify
thread and never on an io thread.

## What this does NOT prove

**The real M4 thresholds are not met and this is not an M4 milestone.**
72 hours and 720 blocks *per posture* on **stagenet** against a synced daemon is
a host-and-days run; minutes on regtest is a rehearsal of the gate, never the
gate. `SoakThresholds::stagenet_m4()` carries the plan's numbers unscaled and is
the default for any non-regtest network, so a stagenet run cannot silently
inherit the mini-soak's numbers. M1's own floors (≥10 000 relayed txs, the
≥720-block `already_generated_coins`/`median_weight` streak) are a separate
claim this branch does not touch.

Also out of scope here: P-SUB contributes no samples in this rig (no block was
found), and the P-TPL backlog-famine CONSTRAINT is oriented to the **served**
arm, so in `serve=monerod` it reads as a served-arm famine claim rather than a
native-arm one. Both are noted in the code rather than left to be discovered.

## Tests

`xmr_native_m4_soak_kat` — 165 checks, 0 failures. The streak machine (CLEAN
accrues, FAIL/SERVED-MISMATCH reset, VOID neither, a void *run* does, a reorg
contributes no block), both graduation halves (one leg is not two; a tip-only
stream never satisfies the template floor; an incomplete key cannot graduate; a
later FAIL takes a qualification away), persistence (round trip, the
version-bump refusal, the other three key mismatches, an unbound daemon version
resuming, a forged `Graduated` corrected, a missing file not an error), the
injector end to end through `compare_seam` (eight fields, each perturbed by one
unit, each must FAIL; a withheld field must FAIL too), and the driver (P-POOL
skipped, silence counted, failures held apart from the bounded tail).

STL-only and clock-free — 72 hours pass in microseconds. Registered with
`add_test` **and** in BOTH `--target` lists in `.github/workflows/build.yml`
(the c2pool#1539 `***Not Run` lesson).

## Incidental

* `ITipObserver` moves to `parity/xmr_tip_observer.hpp` so a decorator of the
  seam does not drag C2c's chain view — and the whole vendored consensus link —
  behind one pure virtual function. `xmr_parity_sources.hpp` includes it and
  exports every name it did before.
* `xmr_native_node` now links `xmr_node`: driving P-TPL means polling the daemon
  arm, which instantiates the production `get_miner_data` parser. A second
  decode path would make "we match the daemon" a claim about a re-typed struct.

## Scope

`src/impl/xmr/native/**` only, plus the two `--target` lines in
`.github/workflows/build.yml`. **Zero files under `src/sharechain/`.** No
consensus digest, no w4 owed-digest fold.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -246,6 +246,7 @@ jobs:
                     xmr_native_block_relay_kat xmr_c5_regtest_push \
                     xmr_native_parity_kat xmr_native_txpool_parity_kat \
                     xmr_native_backlog_famine_kat \
+                    xmr_native_m4_soak_kat \
                     xmr_native_node_kat xmr_native_node \
                     v37_xmr_m2_native_settlement_kat \
                     v37_xmr_m3_arm_order_kat \
@@ -688,6 +689,7 @@ jobs:
                     xmr_native_block_relay_kat xmr_c5_regtest_push \
                     xmr_native_parity_kat xmr_native_txpool_parity_kat \
                     xmr_native_backlog_famine_kat \
+                    xmr_native_m4_soak_kat \
                     xmr_native_node_kat xmr_native_node \
                     v37_xmr_m2_native_settlement_kat \
                     v37_xmr_m3_arm_order_kat \
--- a/src/impl/xmr/native/CMakeLists.txt
+++ b/src/impl/xmr/native/CMakeLists.txt
@@ -140,7 +140,15 @@ endif()
 if (BUILD_TESTING AND UNIX)
     add_executable(xmr_native_node node/tools/xmr_native_node_main.cpp)
     target_include_directories(xmr_native_node PRIVATE ${CMAKE_SOURCE_DIR}/src)
-    target_link_libraries(xmr_native_node PRIVATE xmr_native_txpool)
+    # xmr_node carries the PRODUCTION get_miner_data parser
+    # (MoneroDaemonRpc::body_get_miner_data / parse_miner_data). It became a
+    # link dependency at M4 rather than earlier because nothing in this binary
+    # had ever POLLED the daemon template arm: the M4 soak drives P-TPL, which
+    # means calling MonerodMinerDataSource::poll(), which instantiates that
+    # parse. The alternative -- a second decode path for the daemon arm -- would
+    # make "we match the daemon" a claim about a re-typed struct, which is
+    # exactly what the C4/C6 targets link xmr_node to avoid.
+    target_link_libraries(xmr_native_node PRIVATE xmr_native_txpool xmr_node)
     target_compile_features(xmr_native_node PRIVATE cxx_std_20)
     if (TARGET Boost::headers)
         target_link_libraries(xmr_native_node PRIVATE Boost::headers)
--- a/src/impl/xmr/native/node/tools/xmr_native_node_main.cpp
+++ b/src/impl/xmr/native/node/tools/xmr_native_node_main.cpp
@@ -53,6 +53,7 @@
 #include <cstdio>
 #include <cstring>
 #include <chrono>
+#include <ctime>
 #include <iostream>
 #include <set>
 #include <string>
@@ -60,7 +61,9 @@
 #include <vector>
 
 #include "impl/xmr/native/node/xmr_native_node.hpp"
+#include "impl/xmr/native/parity/xmr_graduation_ledger.hpp"
 #include "impl/xmr/native/parity/xmr_parity_report.hpp"
+#include "impl/xmr/native/parity/xmr_soak_driver.hpp"
 
 namespace rt     = ::c2pool::xmr::native::rt;
 namespace native = ::c2pool::xmr::native;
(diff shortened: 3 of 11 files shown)


frstrtr:

PR #1595, opened 2026-09-11
xmr(p2pool): all-in-one 3-sidechain monitor + fullscreen TUI [DO NOT MERGE]

frstrtr:
## What this is

One **read-only** process watching **all three P2Pool sidechains at once** — main
(37889), mini (37888), nano (37890) — three consensus contexts multiplexed through
**one `poll()`**, behind a fullscreen ANSI/termios TUI. Stacked on c2pool#1593
(`v37/xmr-p2pool-observer`): same wire, same parser, same read model, **no new
encoder**.

```
xmr_p2pool_monitor                          fullscreen, all three chains
xmr_p2pool_monitor --chains mini,nano       two of them
xmr_p2pool_monitor --snapshot --seconds 90  one plain-text frame to stdout
```

## Proven live

A four-minute `--snapshot` run against the real networks on workstation-191, one
process, all three chains populated:

```
 p2pool monitor  READ-ONLY  chains 3/3  up 00:04:00  peers 9/279  blocks 237  races 3  19.7 KiB out
----------------------------------------------------------------------------------------------------
 MAIN  port 37889  LIVE  target 10s ----------------------------------------------------------------
   tip      15200706 0688ec8a  diff 4.35 G  hash 434.79 MH/s  cum 30.91 P
   cadence  13.25 s / 10 s  [##########------]  over 16 live heights above 15200688
   pulse    ._*_+_##-.__+..  gaps, newest right (15)
   races    contested 2  uncles named 3 resolved 2  monero tpl 3760156 over 6 heights
   net      4 up / 4 sock / 124 known / 173 gossip / 164 queued  blocks 81 (dup 165, bad 0, bc 0)
   out      chal 5  sol 4  block_req 246  peer_req 12  = 8.2 KiB sent
   feed     15200706 0688ec8a  30s 39sh u1   15200705 42be9aa4  35s 39sh

 MINI  port 37888  LIVE  target 10s ----------------------------------------------------------------
   tip      14758158 68937209  diff 235.72 M  hash 23.57 MH/s  cum 2.44 P
   cadence  13.55 s / 10 s  [##########------]  over 18 live heights above 14758136
   pulse    ==+#.===##+==..#+  gaps, newest right (17)
   races    contested 1  uncles named 2 resolved 1  monero tpl 3760156 over 7 heights
   net      1 up / 3 sock / 41 known / 47 gossip / 0 queued  blocks 83 (dup 28, bad 0, bc 0)
   out      chal 1  sol 1  block_req 111  peer_req 3  = 3.6 KiB sent
   feed     14758158 68937209  3s 702sh   14758157 f5fd39c8  18s 703sh

 NANO  port 37890  LIVE  target 30s ----------------------------------------------------------------
   tip      1392499 4f099523  diff 128.90 M  hash 4.30 MH/s  cum 136.54 T
   cadence  31.61 s / 30 s  [########:-------]  over 8 live heights above 1392490
   pulse    ._.*+_#  gaps, newest right (7)
   races    contested 0  uncles named 0 resolved 0  monero tpl 3760156 over 14 heights
   net      4 up / 4 sock / 114 known / 164 gossip / 144 queued  blocks 73 (dup 164, bad 0, bc 0)
   out      chal 5  sol 4  block_req 237  peer_req 12  = 7.9 KiB sent
   feed     1392499 4f099523  12s 404sh   1392498 b9e4ea7f  1m26s 404sh

----------------------------------------------------------------------------------------------------
 emitted ids {0,1,3,6}  CHALLENGE  SOLUTION  BLOCK_REQUEST  PEER_LIST_REQUEST
 never emitted: LISTEN_PORT, BLOCK_RESPONSE, BLOCK_BROADCAST, BLOCK_NOTIFY -- no encoder exists
```

`main` 13.25 s over 16 live heights, `mini` 13.55 s over 18, `nano` 31.61 s over 8;
9 handshaken peers; 237 distinct blocks; 3 same-height races with 3 uncles resolved;
19.7 KiB sent, all of it handshakes, block requests and peer-list requests. All three
panels carry the **same Monero template height (3760156)** — the cross-chain sanity
check that these are three views of one network rather than three copies of one chain.

## One poll(), three chains, no threads

An `Observer` **is** one sidechain (one consensus id, one peer set, one read model,
one byte ledger) and that shape is kept — the monitor holds three of them. What
cannot be had three times is the event loop: three `::poll()` calls in sequence means
each blocks the other two for its timeout, and the five-second tip-poll rule that
keeps a pulling observer from being banned starts to slip. Threads would fix the
stall and bring a mutex around every read model for three sockets' worth of traffic.

So the observer's loop is split into the three phases it already contained:

```
for each chain: begin_tick()                       dial, reap
for each chain: begins[i] = pfds.size(); collect() append its sockets
append the tty                                     a keypress wakes the loop
::poll(pfds, timeout)                              the only blocking call
for each chain: dispatch(pfds + begins[i], n_i)    its own slice, only
```

**The range is the tag.** A chain's sockets occupy a contiguous slice of the shared
array, so there is no per-fd chain map to build, look up or leave stale, and a chain
cannot be handed another chain's `revents` — it is handed a pointer and a length.
`run()` is rewritten in terms of the same three calls, so the single-chain harness
and the monitor cannot drift apart. Per-connection timers needed nothing:
`next_tip_poll_ms_` already lives inside `PeerSession`.

## Still read-only, and the frame says so from the encoder

No new encoder, no `PoolBlock` serialiser, no `LISTEN_PORT`, no `bind`/`listen`/
`accept`. The emitted id set is still exactly **{0, 1, 3, 6}**, the new KAT
re-asserts it for this second binary, and the footer of every frame prints that set
**as computed from the encoder** — encode each `ControlMessage`, read the id byte
back — rather than as a typed-in list. A fifth encoder would start announcing itself
on screen in the same breath as it failed the test.

## The TUI has no dependencies

ANSI escapes and POSIX `termios`. Raw mode is four flags, the alternate screen and
the cursor are two escapes each, the size is `TIOCGWINSZ`, `SIGWINCH` sets a flag.
**No ncurses, no boost, no asio, no threads.** The restore path is installed three
ways over — a quit flag on `SIGINT`/`SIGTERM`, the `TerminalUi` destructor, an
`atexit` hook — because a monitor that dies in raw mode leaves someone with a shell
that does not echo. Everything drawn is printable ASCII, so one byte is one column
and the alignments are not guesses.

## A pure renderer, pinned against a golden frame

`p2pool_tui.hpp` is deliberately POSIX-free: no socket, no terminal, no clock — ages
and uptime resolve at one instant handed in by the caller. That is what lets
`xmr_p2pool_monitor_kat` fill three read models with fixed synthetic observations at
fixed timestamps and compare the **whole frame**, byte for byte, against
`test/p2pool_monitor_golden.inc` (kept in its own file so a reviewer can read it as a
picture). A dashboard is the easiest place in a codebase to hide a wrong number,
because the only thing that would catch it is somebody looking at it.

Colour is additive and nothing else: `strip_ansi(render(colour))` equals
`render(plain)` line for line, which the KAT asserts — so `--snapshot` is the same
renderer with the escapes off, not a second one that could disagree with the screen.

The panels inherit the read model's refusals rather than softening them for looks:
below two live samples the cadence line says `warming` instead of printing a number,
the pulse shows the individual live gaps because a mean hides the difference between
a steady beat and a stall between bursts, and `monero tpl` is labelled a template
height — deciding that a sidechain block cleared Monero's target needs a PoW target
this observer does not hold.

## What the live runs found

**A whole sidechain went dark, and the panel is how it was caught.** On the second
live run main reported `DARK — 0 up / 0 sock / 0 known / 0 queued / 0 B sent` for
four minutes while mini and nano were fine. Three real bugs behind it, all fixed:

* both of main's DNS seeds failed inside `start()`, and `fail()` only logged when the
  session had already left `State::Closed` — which a session that never connected
  never does. The failure was **silent**. It logs now, and the very next run printed
  `peer main.p2poolpeers.net:37889 closed: dns`, naming a seed host that does not
  resolve;
* a failed dial then took the same five-minute cool-off as a peer that had talked and
  hung up. A dial that never reached a socket is a different fact and now earns a
  15-second retry;
* with no peer connected there is no peer-list gossip, so the seeds were the only way
  back. `reseed()` now clears the cool-off on the seed endpoints of a chain that has
  nothing, and the monitor reseeds a dark chain every 20 seconds.

Two display defects as well, invisible until three real chains were on screen at
once: the final snapshot was taken *after* the sockets were closed and so reported
`4 up / 0 sockets`, and the `net` line overflowed 100 columns once peer counts reached
three digits, truncating `bcast` mid-word.

`DARK` next to `0 queued` said on sight that the chain had nothing left to dial —
a different failure from "peers are refusing us", and the two look identical in a log.

## Tests

| | |
|---|---|
| `xmr_p2pool_monitor_kat` | **181 checks, 0 failures** — emit-set `{0,1,3,6}`, golden frame, geometry at 12 sizes, colour-is-additive, 3-chain aggregation, the honesty rules |
| `xmr_p2pool_parse_kat` | **253 checks, 0 failures** — unchanged, still green |

Both are offline: no network, no daemon, no tty. Both registered with `add_test` and
in **both** `--target` lists in `build.yml`, along with both live binaries, because
this directory sits outside the coin-test allowlist glob (c2pool#1539: a target
registered with `add_test` but never built reports `***Not Run` and exits non-zero).

`xmr_p2pool_monitor` dials three public networks and its fullscreen mode wants a tty,
so — exactly like `xmr_p2pool_observer` — CI **builds** it and **never runs** it, and
it registers no ctest test.

## One behaviour change to the observer

Re-dialling. Never re-dialling is right for a 300-second harness and wrong for a
process left running for hours, where every peer that ever hung up would be excluded
permanently and the dial queue would drain. `redial_after_ms` defaults to **0
(never)**, so `run()` behaves exactly as before; the monitor sets five minutes, and
the queue is capped so gossip cannot grow it without bound overnight.

## Scope

Everything under `src/impl/xmr/p2pool/` plus its two `CMakeLists.txt` and the two
`build.yml` target lists. **Zero files under `src/sharechain/`, no consensus edit.**

DRAFT, stacked on c2pool#1593 — do not merge ahead of it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -250,6 +250,7 @@ jobs:
                     v37_xmr_m2_native_settlement_kat \
                     v37_xmr_m3_arm_order_kat \
                     xmr_p2pool_parse_kat xmr_p2pool_observer \
+                    xmr_p2pool_monitor_kat xmr_p2pool_monitor \
             -j"${BUILD_J:-8}" 2>&1 | tee "${GITHUB_WORKSPACE}/ci-logs/build-tests.log"
 
       - name: Run tests
@@ -693,6 +694,7 @@ jobs:
                     v37_xmr_m2_native_settlement_kat \
                     v37_xmr_m3_arm_order_kat \
                     xmr_p2pool_parse_kat xmr_p2pool_observer \
+                    xmr_p2pool_monitor_kat xmr_p2pool_monitor \
             -j4  # ASan+UBSan: cc1plus RSS ~2-3GB each; -j8 wedged runner cgroup memory.high (throttle-hang). -j4 caps peak <12G.
                  # xmr_provenance_gate: previously omitted here because
                  # -fsanitize=undefined rejected its constexpr pointer comparison
--- a/src/impl/xmr/p2pool/CMakeLists.txt
+++ b/src/impl/xmr/p2pool/CMakeLists.txt
@@ -21,19 +21,35 @@
 # a sidechain blob is handed to THOSE parsers rather than to a second copy
 # written here -- which is the whole reason this component is stacked on M3.
 #
-# Two targets, both under BUILD_TESTING:
+# Four targets, all under BUILD_TESTING:
 #
-#   xmr_p2pool_parse_kat   the deterministic parser KAT over two real captured
-#                          frames. REGISTERED with add_test and run in both CI
-#                          legs. Offline: the goldens are in the header.
-#   xmr_p2pool_observer    the live harness. BUILT by CI so it cannot bit-rot,
-#                          RUN by nobody in CI, because running it dials a
-#                          public network. Registers no ctest test, so the
-#                          #1539 Not-Run rule does not apply to it.
+#   xmr_p2pool_parse_kat    the deterministic parser KAT over two real captured
+#                           frames. REGISTERED with add_test and run in both CI
+#                           legs. Offline: the goldens are in the header.
+#   xmr_p2pool_monitor_kat  the deterministic RENDER KAT: three fixed read
+#                           models drawn to a golden frame, plus the emitted
+#                           message-id pin re-asserted for the monitor. Also
+#                           REGISTERED, also offline -- it includes no socket
+#                           header and needs no terminal.
+#   xmr_p2pool_observer     the live single-chain harness. BUILT by CI so it
+#                           cannot bit-rot, RUN by nobody in CI, because running
+#                           it dials a public network. Registers no ctest test,
+#                           so the #1539 Not-Run rule does not apply to it.
+#   xmr_p2pool_monitor      the live all-in-one monitor: all three sidechains in
+#                           ONE poll() loop behind a fullscreen ANSI/termios
+#                           TUI. Same rule as the observer -- BUILT by CI, never
+#                           RUN by it (it dials three public networks and the
+#                           fullscreen mode wants a tty), and it registers no
+#                           ctest test either. What CI runs is the render KAT.
+#
+# THE TUI HAS NO DEPENDENCIES. It is ANSI escapes plus POSIX termios: no
+# ncurses, no boost, no asio, no threads. The renderer itself
+# (p2pool_tui.hpp) is pure STL and deliberately POSIX-free, which is what lets
+# the KAT pin a whole frame without linking a socket layer.
 #
 # DRIFT-GUARD: tools/ci/check_test_target_allowlist.py audits
 # src/impl/<coin>/test/CMakeLists.txt; this tree sits one level deeper
-# (src/impl/xmr/p2pool/test/) and is outside that glob, so BOTH targets are
+# (src/impl/xmr/p2pool/test/) and is outside that glob, so ALL FOUR targets are
 # added to BOTH `--target` lists in .github/workflows/build.yml by hand. A
 # target registered with add_test but never built reports "***Not Run" and
 # exits non-zero -- the #1539 lesson.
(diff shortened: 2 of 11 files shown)


frstrtr:

PR #1597, opened 2026-09-11
xmr(native): autonomous reorg-follow after same-height race loss [DO NOT MERGE]

frstrtr:
**DRAFT — DO NOT MERGE.** Stacked on `v37/xmr-native-multinode` (c2pool#1592, the multi-node prefer-own same-height tiebreak). Node behaviour only; no consensus rule changes, nothing under `src/sharechain/`.

## The bug

A native node that lost a same-height race **sat on its own tip forever**. Its daemon reorganised onto the winning branch and relayed the blocks; the node dropped every single one of them. Only a process restart appeared to fix it — because a restart re-syncs through the TIP path, where the branch path never runs.

## Cause: a third gate, in the chain index, specific to a young chain

`RowStore::difficulty_window_ending_at` refused any 735-row window whose span reached below height 0. A genesis-booted index (`ChainBoot::try_seed_`) seeds **one** difficulty row, at height 0 — so `branch_difficulty_locked_` could not price a single block that attached anywhere but the tip until the chain was 735 blocks long. That is **every regtest rig, always**, and the first day of any from-genesis sync.

The tip fast path was never affected: it reads the consensus state's own rolling window and gets whatever rows exist, which is exactly what monerod's `get_difficulty_for_next_block` does with its `min(height, 735)`. So the branch path was refusing a window the tip path was happily computing over, for the same rows.

Anchored boots (stagenet, mainnet) seed a full 735-row pre-window — and so does the C2c KAT fixture. That is why `xmr_native_chain_index_kat` passes 1717/0 **with a reorg suite in it** while a from-genesis node could not reorganise at all.

**A second defect made it silent.** On that refusal `offer_locked_` answered `ParkedOrphan` and returned *without inserting the block*, without naming anything for the fetcher and without a journal line — the block was destroyed. Every re-delivery hit the same return, so `AltPool::heaviest()` never had a candidate, `maybe_switch_locked_` never opened a record, and nothing anywhere reported that a block had been thrown away.

## The fix

| file | change |
|---|---|
| `chain/xmr_row_store.hpp` | `difficulty_window_ending_at(..., allow_young_chain)`, granted **only** when `base_height() == 0` — we hold the chain from its first block. A store that merely *starts* high still refuses: that is a real refusal and stays one. |
| `chain/xmr_chain_index.hpp` | `branch_difficulty_locked_` is the only caller that opts in. `offer_locked_` **parks** the unjudgeable block for real (unresolved, un-adoptable — `heaviest()` skips it, the fail-closed bound still refuses any branch containing it) so `resolve_descendants_locked_` can pick it up when the parent resolves. `refetch_wanted()` stops naming ids whose bodies we already hold. The tip fast path runs the fork choice after draining parked children. |
| `p2p/xmr_peer_pool.hpp` | Folds `current_blockchain_height` from block pushes and objects responses into the peer's sync snapshot. Only TIMED_SYNC did, on a 60 s beat, so for up to a minute after a peer moved we asked it nothing. The **height** only — `cumulative_difficulty` stays what the peer actually signed for. |

## KAT — `xmr_native_reorg_follow_kat`, 135 checks

Boots the way a regtest node boots (one row at height 0) — a state the anchored C2c fixture cannot reach, which is why it is a separate target rather than more cases in `xmr_chain_index_kat.cpp`.

* **A** the young-chain window at `RowStore` level: granted at `base_height()==0`, refused for a store that starts high, and the "do not invent a window past the pre-window" refusal survives the opt-in
* **B** genesis-booted index at height 8 is handed the rival of a same-height race and then its extension: it holds, judges, and **adopts** — one live object, no re-seed, no snapshot reload
* **C** delivery bottom-up with a hole: a block above a parked UNRESOLVED parent is held (not dropped), the missing parent is named in `refetch_wanted()`, filling the hole resolves the branch and switches; ids we hold are no longer re-asked
* **D** bounds unmoved: deeper than `max_reorg_depth` is refused **and journaled**; equal work against our own block is KEPT (D-14) with the rival held so the accounting can see the race; a lighter branch never switches
* **E** the relaxed `RowStore` window and the consensus state's rolling window agree on the difficulty, or the switch would roll itself back

Built against `613d7cc1` the same file **fails 42 checks** with `the retained window does not reach the fork point` and an alt pool holding nothing. Registered with `add_test` **and in both `build.yml` `--target` lists** (the c2pool#1539 Not-Run lesson).

Green on workstation-191, g++-15, `XMR_BUILD_RANDOMX=ON`:
`xmr_native_reorg_follow_kat` 135/0 · `xmr_native_chain_index_kat` 1717/0 · `xmr_native_peer_pool_kat` 64/0 · `xmr_native_node_kat` 88/0 · `xmr_native_block_relay_kat` 98/0 · `xmr_native_parity_kat` 833/0 · `v37_xmr_prefer_own_race_kat` 95/0 · `v37_xmr_m3_arm_order_kat` 40/0 · `v37_xmr_m2_native_settlement_kat` 85/0

## Live proof — isolated regtest rig, before and after

Two partitioned regtest monerods (A, B) plus a control hub C, two daemonless native nodes (`--arm-order p2p-first`, `--no-daemon-rpc`, `D_conf=3`, distinct payees, `--randomx`), own ports and datadirs, all torn down afterwards. Shared 20-block prefix, partition, each node mines its own block at **h=21**, A's branch is extended by two, the partition heals, and Monero itself buries B's branch (`REORGANIZE SUCCESS! on height: 21`).

**BEFORE** (binary at `613d7cc1`) — loser frozen, 2+ minutes after the heal, daemon at h=24:

```
node2  status: hw=21 cursor=18 tip=21 | found registered=20 settled=0 orphaned=0 pending=20
       native tip feed: events=21 (extend=21 reorg=0 orphan=0) best=21
       same-height: credited=0 refused[orphaned=0 other-only=0] deferred=1 | R-7 violations=0
```

**AFTER** (this branch) — loser follows on its own:

```
node2  status: hw=25 cursor=22 tip=25 | found registered=10 settled=0 orphaned=10 pending=0
       native tip feed: events=26 (extend=24 reorg=1 orphan=1) best=25
       same-height: credited=0 refused[orphaned=1 other-only=0] | R-7 violations=0
       wire RPC to monerod: TOTAL=0    arm-order=p2p-first rpc_on_find_path=0

node1  native tip feed: events=25 (extend=25 reorg=0 orphan=0) best=25
       same-height: credited=1 refused[orphaned=0 other-only=0] | R-7 violations=0
```

Tip trace (status every ~10 s): `tip=21 refresh=222` → `tip=23 refresh=242` — the switch landed inside one status window of monerod B's own REORGANIZE, roughly 10-20 s after the heal. **No restart**: same process, same pid, from the race through the reorg to the credit.

Settlement journals, the whole ledger at the contested height:

```
node1  h=21 verdict=already-credited  credit=12e7118d34f1...   (its own, canonical)
node2  h=21 verdict=refuse-orphaned   credit=-                 (its own, orphaned)
```

One credited entry fleet-wide at h=21, the loser fabricates nothing, `R-7 violations=0` on both.

## Scope, stated plainly

**Regtest and minutes is not a stagenet milestone.** What is proven here is a depth-1 reorg on a private chain of 26 blocks at difficulty 1, in one live run, on one host. Untouched by this work: reorgs deeper than the fork points a young chain produces, a real stagenet/mainnet anchored boot (which never hit the young-chain gate and is unchanged by the opt-in), RandomX seed rollover across an epoch edge during a switch, and the snapshot base path — which still declines to go below the tip on a young chain, deliberately left alone rather than widened on the same change.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -238,6 +238,7 @@ jobs:
                     xmr_levin_codec_kat xmr_epee_storage_kat xmr_levin_messages_kat xmr_levin_fuzz_kat \
                     xmr_native_block_id_kat xmr_native_consensus_state_kat \
                     xmr_native_anchor_self_check_kat xmr_native_chain_index_kat \
+                    xmr_native_reorg_follow_kat \
                     xmr_levin_socket_kat xmr_levin_link_kat \
                     xmr_native_rct_verify_kat xmr_native_txpool_kat \
                     xmr_native_chain_locator_kat xmr_native_peer_store_kat \
@@ -680,6 +681,7 @@ jobs:
                     xmr_levin_codec_kat xmr_epee_storage_kat xmr_levin_messages_kat xmr_levin_fuzz_kat \
                     xmr_native_block_id_kat xmr_native_consensus_state_kat \
                     xmr_native_anchor_self_check_kat xmr_native_chain_index_kat \
+                    xmr_native_reorg_follow_kat \
                     xmr_levin_socket_kat xmr_levin_link_kat \
                     xmr_native_rct_verify_kat xmr_native_txpool_kat \
                     xmr_native_chain_locator_kat xmr_native_peer_store_kat \
--- a/src/impl/xmr/native/chain/xmr_chain_index.hpp
+++ b/src/impl/xmr/native/chain/xmr_chain_index.hpp
@@ -655,10 +655,25 @@ public:
     // plus anything a refused reorg needs before it can be retried. The sync
     // driver reads this; the index never schedules its own network traffic
     // beyond the single hint request above.
+    //
+    // Ids whose BYTES we already hold are filtered out here rather than pruned
+    // from the lists: `wanted_` is replaced wholesale by the next chain entry
+    // and `refetch_` is deliberately a standing want list, so neither is a
+    // record of what is still missing. Without the filter the driver re-asked
+    // for a block it was already holding every refetch_reask_ms, forever --
+    // which is what a parked branch looks like from the outside, and what made
+    // "we are not making progress" indistinguishable from "the peer is not
+    // answering". Membership is not enough: a bodiless fluffy announcement and
+    // a row whose body has been evicted are both "known" and both still need
+    // fetching, so the test is whether we have bytes we could re-apply.
     std::vector<Hash> refetch_wanted() const {
         std::lock_guard<std::mutex> lk(mu_);
-        std::vector<Hash> out = wanted_;
-        out.insert(out.end(), refetch_.begin(), refetch_.end());
+        std::vector<Hash> out;
+        out.reserve(wanted_.size() + refetch_.size());
+        for (const Hash& id : wanted_)
+            if (!have_body_locked_(id)) out.push_back(id);
+        for (const Hash& id : refetch_)
+            if (!have_body_locked_(id) && !contains_hash_(out, id)) out.push_back(id);
         return out;
     }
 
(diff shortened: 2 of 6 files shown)


frstrtr:
Landed via #1612 (daemonless tower M0..M4, merged to master): the reorg-follow files — xmr_reorg_follow_kat.cpp and the fork-choice follow path — are present on master. Closing as superseded. Branch kept (no deletion).

PR #1598, opened 2026-09-11
xmr(p2pool): monitor durability + absence-vs-zero hardening [DO NOT MERGE]

frstrtr:
**DRAFT — DO NOT MERGE.** Stacked on `v37/xmr-p2pool-monitor` (c2pool#1595), which is itself a DRAFT. Base is that branch, not `v37-dev`.

Two hardening requirements on the three-sidechain P2Pool monitor. Strictly read-only throughout: no encoder is added, the emitted id set is still `{0,1,3,6}`, and persistence writes a **local host file** and nothing else — never a byte to a P2Pool peer.

---

## 1. Durability — the numbers survive the process

Everything the observer learned lived in a `ReadModel` in process memory, and `--snapshot` — the mode actually left running overnight — prints its one frame at the *end*. A six-hour watch killed at hour five produced literally no output, and the rare observation this component exists to make (a race, a stall, a chain going dark at 4am) is exactly the one a session-lifetime memory loses.

The monitor now writes what it knows to `~/p2pmon-state/` on a timer in **both** live modes and reads it back at start-up.

```
p2pmon-state.json      the latest complete snapshot — what is read back on start
p2pmon-history.jsonl   an append-only digest, one line per save, for tailing
p2pmon-history.jsonl.1 one rotated generation
.lock                  flock; a second monitor runs with persistence OFF
```

* **Never half-written.** Temp file in the same directory → `fdatasync` → `rename()` over the real name → `fsync` the directory. Every reader, including the next run, sees either the whole previous state or the whole new one, at every instant. A torn file that parses as *plausible* garbage would be worse than no file in a component whose entire point is not lying.
* **Every time in the file is absolute, and on the wall clock.** "The tip last moved 40 s ago" becomes false while the file sits on disk; `1757620000123` stays true and is turned back into an age against the *current* clock on load. That one rule is why a day-old file renders as `STALE` with the age spelled out rather than as a healthy chain — and it only holds if the instant is on a clock the *next* process agrees with. See §3.
* **Exact numbers.** Difficulty and cumulative difficulty are written as decimal **strings** and read back through a 128-bit accumulator — a JSON double loses the low bits of a cumulative difficulty and comes back plausible. The derived hashrate is written for external readers and recomputed on load, so restored and live panels cannot round differently.
* **The file cannot disagree with the screen.** `MonitorState` is built from the same `tui::MonitorFrame` the renderer draws, at the same instant — not from a second pass over the read models.
* **It degrades, never breaks.** No `HOME`, an unwritable directory, a full disk, another monitor holding the lock: each turns persistence off (or counts an error) and says so on the header row, and the poll loop carries on. There is deliberately **no `/tmp` fallback** — a state file a reboot deletes is the precise outcome this feature exists to prevent.

`--read` renders a saved state through the **same** renderer with no socket opened, under a `RESTORED FRAME` banner and a `[FROM FILE 10s OLD]` marker on every panel.

The state file is JSON written by ~40 lines of string building and read back through the repository's existing header-only `src/impl/xmr/node/minijson.hpp` — included as-is, no edit, no new dependency. The p2pool targets stay a LIGHT link (`xmr_coin` only).

## 2. Absence is not zero

A chain that went away rendered as a complete set of confident zeros in the same shape a live chain uses. `peers 0, blocks 0` on a chain that is fine but that *we* are no longer connected to is indistinguishable, on sight, from "the chain is quiet".

Zero is now reserved for a measurement that came out zero. A figure with no basis renders `--`.

The status word grew from four values (which only described a connection *warming up*) to seven that can also describe one that **stopped**:

| word | meaning | clock |
|---|---|---|
| `DARK` | never connected, nothing in flight | since the session started |
| `DIAL` | sockets in flight, no handshake yet | since the session started |
| `DOWN` | we **had** peers and now have none | since the last peer went |
| `WARM` | handshaken, fewer than two live heights | since the tip moved |
| `QUIET` | tip has not moved for longer than variance explains | since the tip moved |
| `STALE` | tip has not moved for so long the panel is not current | since the tip moved |
| `LIVE` | handshaken, timing the chain, tip moving | — |

`QUIET`/`STALE` are two thresholds on **one** clock — the last time the tip *height increased*; a duplicate, a backfilled parent or a peer-list does not touch it. The thresholds are multiples of the chain's own target block time with a floor, because a sidechain arrival is near-Poisson and a fixed 30-second threshold would cry wolf several times an hour (10 s chain: `QUIET` at 60 s, `STALE` at 180 s; 30 s chain: 120 s and 360 s).

A degraded panel gets a `!` row above its figures saying what stopped and how old the rest of the panel is, and the header carries a **coverage** row beside every three-chain total — an aggregate over a dead source is arithmetically right and editorially false, and dropping the dead chain would hide it. Peer counts are deliberately **not** dashed: `0 up` is a measurement of our own socket table, paired with the word that says what it means.

## Two defects found by running it

Neither was visible in review; both were found on `workstation-191` against the three live networks.

* **`poll_once()` returned early when no socket existed at all — before the freshness sample.** A monitor whose chains were *all* dark therefore never started a single clock and drew `DARK never connected` with no age, on the one run where how long it had been dark was the whole question. The three-chain case hid it: one live chain's sockets make the array non-empty and sample all three.
* **A KAT helper cut panels out of the frame by chain name**, and the new coverage row also names all three chains — so it was asserting against the header instead of the panels.

## KATs

`xmr_p2pool_monitor_kat` (still POSIX-free, still golden-pinned; golden regenerated deliberately) gains:

* the **absence-vs-zero** fixture — one `LIVE`, one `STALE`, one `DOWN` chain that never received a block. The dead panel is walked row by row and **any** digit that is not the chain's identity or an age fails the test; the two live chains are asserted unaffected in the same frame;
* the **seven-state liveness ladder**, including precedence (no peers outranks a fresh tip; `STALE` outranks a computable cadence) and the age-clamp for a clock that runs backwards;
* the **state-file codec round trip**, field by field, including a `2^128-1` cumulative difficulty and the refusal of empty, foreign-schema and truncated files.

`xmr_p2pool_persist_kat` is **new** and makes the durability claim operationally rather than syntactically: it **forks a child, has it save a state, kills it with `SIGKILL`** — no unwinding, no `atexit`, no flush — and reads every number back in the parent. Plus the directory lock (a second monitor must not clobber the live writer's file, and a reader takes no lock at all), the no-`HOME` refusal, journal rotation, and a corrupt file being refused rather than half-believed.

Registered with `add_test` **and** appended to **both** `--target` lists in `.github/workflows/build.yml` — a target registered but not built reports `***Not Run` and fails the leg (c2pool#1539).

```
100% tests passed, 0 tests failed out of 3
  xmr_p2pool_parse_kat ..... Passed
  xmr_p2pool_monitor_kat ... Passed   (437 checks)
  xmr_p2pool_persist_kat ... Passed   (105 checks)
```

## Proven live on workstation-191

**(a) All three LIVE, state written every 2 s** — `--snapshot --seconds 150`, seq 71 on disk:

```
 p2pool monitor  READ-ONLY  chains 3/3  up 00:02:30  peers 12/316  blocks 224  races 2  17.6 KiB out
 coverage 3/3 reporting (the totals above cover those)  MAIN LIVE  MINI LIVE  NANO LIVE
 persist  ON /home/user0/p2pmon-state  seq 71 saved 1s ago  session 1  stall 884 ms  total up 00:02:
 MAIN  port 37889  LIVE  target 10s
   tip      15201303 45b968fb  diff 4.21 G  hash 420.57 MH/s  cum 30.92 P  as of 26s
   cadence  7.48 s / 10 s  [#####---:-------]  over 16 live heights above 15201284
```

**(b) One source taken away** — `--peer nano:127.0.0.1:9 --no-seed nano`. The dead chain says so and shows `--` for every figure; the other two stay `LIVE`; the process stays up and exits 0:

```
 p2pool monitor  READ-ONLY  chains 2/3  1 down  up 00:02:30  peers 8/183  blocks 149  races 0  11.4
 coverage 2/3 reporting (the totals above cover those)  MAIN LIVE  MINI LIVE  NANO DARK
 MAIN  port 37889  LIVE  target 10s
   tip      15201317 117e9ecf  diff 4.21 G  hash 420.78 MH/s  cum 30.92 P  as of 6s
 MINI  port 37888  LIVE  target 10s
   tip      14758830 f6526f09  diff 240.24 M  hash 24.02 MH/s  cum 2.44 P  as of 36s
 NANO  port 37890  DARK 2m30s never connected  target 30s
   !        DARK  never connected, 2m30s into this session -- nothing below was measured
   tip      -- --  diff --  hash --  cum --
   cadence  --  no block held for this chain: nothing to time
   races    contested --  uncles named -- resolved --  monero tpl -- over -- heights
   net      0 up / 0 sock / 0 known / 0 gossip / 0 queued  blocks -- (dup --, bad --, bc --)
```

**(c) `kill -9` and restart.** The killed process printed **nothing** to stdout (0 bytes) — the argument for the file. The file it left is complete, at seq 25, with **no** `"shutdown"` marker, which is how the next start recognises a killed session:

```
$ xmr_p2pool_monitor --read --width 100      # no network dialled
 RESTORED FRAME  rendered from the state file, written 10s ago -- no network was dialled and no numb
 MAIN  port 37889  LIVE  target 10s  [FROM FILE 10s OLD]
   tip      15201325 0297bb4f  diff 4.22 G  hash 421.52 MH/s  cum 30.92 P  as of 14s
   carry    prev tip 15201325 0297bb4f  saved 10s ago  lifetime tip 15201325  monero tpl 3760210
   feed     (the state file carries totals, not the arrival feed)

$ xmr_p2pool_monitor --snapshot --seconds 75  # restart, live again
state: restored seq 25 from /home/user0/p2pmon-state/p2pmon-state.json (43s old, session 3)
       -- previous session did not shut down cleanly
 persist  ON /home/user0/p2pmon-state  seq 36 saved 0s ago  session 3 from 43s-old state  stall 427
 MAIN  port 37889  LIVE  target 10s
   tip      15201336 ebb62c99  diff 4.20 G  hash 420.02 MH/s  cum 30.92 P  as of 9s
   carry    prev tip 15201325 0297bb4f  saved 1m58s ago  lifetime tip 15201336  monero tpl 3760210
```

The pre-kill tip is carried forward on the `carry` row, and the lifetime high-water mark advances monotonically across the restart.

Also proven live: `persist OFF (--no-persist)` and `persist OFF (locked by pid N)` banners, and the all-dark case after the fix (`DARK 8s never connected`).

Build host left clean: no processes, no state directories, nothing else on the machine touched.

---

## 3. The clock those persisted instants are on

*(Added in `c5a801b7`, after the durability verify found it. A separate commit, not an amend, so the fix is reviewable on its own.)*

The durability claim in §1 is one subtraction: `age = now - the instant we wrote down`. Every instant written down came from `std::chrono::steady_clock` — **milliseconds since the writing host booted** — so that subtraction was meaningless the moment the file outlived the boot.

Inside one boot the two clocks are indistinguishable, which is why the format round trip, the `SIGKILL` test and a live soak all passed. After a **reboot** they are not: a 235 s-old file holding `tip_advanced_at_ms = 3000000` is subtracted from a ms-since-*this*-boot `now` of `12000`, the unsigned result clamps at zero, and the monitor renders **"written 0s ago"** with every chain **"LIVE as of 0s"**. That is the exact silent-freshness lie this component exists to prevent, produced in the one case — a reboot — for which the persistence layer already refuses to put the file in `/tmp`. Copy the file to another machine and the same subtraction fabricates the mirror image: an age of days on a file written a second ago.

### The split

`now_ms()` is gone. A name that says "now" and means "ms since boot" is what made this invisible at twenty call sites; a caller now has to name the clock it wants.

| | clock | what it carries |
|---|---|---|
| `steady_ms()` | `steady_clock`, monotonic | every **intra-session duration**: poll cadence, connect deadlines, run window, repaint and save timers, the loop-stall watch, inter-arrival gaps, the feed's "arrived 7 s ago". Already correct, **unchanged** — wall time steps backwards under NTP and must never carry these. |
| `wall_ms()` | `system_clock`, ms since the UNIX epoch | every **persisted instant**: all five `FreshnessView` instants, `written_at_ms`, `session_started_at_ms`, `first_started_at_ms`, the journal's `t`. Freshness that must survive a restart keys off these. |

`FreshnessView` is wall throughout. `tui::view_of()` takes both clocks explicitly with **no default on either** — a caller holding one number has not yet worked out which of the two it has, and that is the mistake this makes unwriteable. `p2pool_state.hpp`'s header promised wall-clock instants from the first commit; the producing code simply did not deliver it.

### A clock that runs backwards is said out loud

`FreshnessView::age()` already clamped a future instant to zero, but a confident `0s` reads as "this just happened" — the same class of lie, in miniature. `FreshnessView::ahead()` now reports the clamp, such ages render `0s+`, and the frame draws a `CLOCK SKEW` row explaining the marker once, at `kAlways` priority.

### Schema `/1` → `/2`

The field *names* did not change; the *meaning* of every instant in them did, and nothing in the bytes distinguishes the two but magnitude. A `/1` file is therefore **refused rather than reinterpreted** — one lost carry-over, once, with `starting fresh (schema is "p2pmon-state/1", expected "p2pmon-state/2")`, instead of a first session whose every panel is dated 20 000 days ago.

### Proven reboot-safe on the real binary

Two monitors wrote a state file **seedlessly** on workstation-191 (every chain `--no-seed`, no `--peer`: nothing dialled, no name resolved). The files alone settle it, against a host uptime of 126 751 s:

```
-- FIXED:                              -- PRE-FIX:
"schema":"p2pmon-state/2"              "schema":"p2pmon-state/1"
"written_at_ms":1789149234796          "written_at_ms":126791073
"tip_advanced_at_ms":1789149214739     "tip_advanced_at_ms":126771013
```

Each file was then read back **235 s later by a fresh process** whose `CLOCK_MONOTONIC` epoch was shifted while wall time was left untouched — exactly and only what a reboot does to a reader (an `LD_PRELOAD` shim offsetting `clock_gettime(CLOCK_MONOTONIC)`; no privileges taken, no host clock changed, nothing left behind).

| reader | PRE-FIX build, `/1` file | FIXED build, `/2` file |
|---|---|---|
| same boot (control) | `written 3m55s ago` | `written 4m15s ago` |
| reboot, monotonic **+1 d** | **`written 1d ago` / `DARK 1d`** | `written 4m15s ago` |
| reboot, monotonic **−120 000 s** | **`written 0s ago` / `DARK 0s`** | `written 4m15s ago` |

The pre-fix build reports a fabricated day and then the silent `0s` for the *same* 235 s-old file; the fixed build reports the real elapsed time in every case, including the control. Reading the `/1` file with the fixed build is refused by name:

```
cannot read state from .../state-old: schema is "p2pmon-state/1", expected "p2pmon-state/2"
```

### KATs

All offline, both CI legs, **no new target and no CMake or `build.yml` edit**:

| KAT | before | after |
|---|---|---|
| `xmr_p2pool_parse_kat` | 253 / 0 | 253 / 0 |
| `xmr_p2pool_monitor_kat` | 438 / 0 | **459 / 0** |
| `xmr_p2pool_persist_kat` | 105 / 0 | **141 / 0** |

**The rendered golden is untouched** — the fixture holds no future-dated instant, so no marker and no skew row appear in it, and no regeneration was required.

The new checks pin the *clock*, not the arithmetic. `persist_kat` section 6 drives a real `MultiWatch` turn — the same `poll_once()` the monitor's loop calls, seedless so it opens no socket — and asserts the instants it recorded are epoch-scale; then a child writes a file and the parent reads it back in a process that shared nothing with the writer but the file, and the rendered age is the time that actually passed. `monitor_kat` adds the 235 s reboot shape, the backwards-clock direction, the `0s+` marker and the `/1` refusal.

Read-only is unchanged throughout: no encoder added, emit set still `{0,1,3,6}`, persistence writes a local host file and nothing else. Absence-vs-zero markers intact.

Build host left clean: binaries and state directories under one scratch tree, no processes left running, nothing else on the machine touched.

---

## Scope

`src/impl/xmr/p2pool/**` plus its CMake and `.github/workflows/build.yml`. **Zero files under `src/sharechain/`.** No consensus is activated, nothing is merged or rolled out.


frstrtr:

PR #1608, opened 2026-09-11
xmr(native): make the M4 ledger survive being killed mid-write

frstrtr:
**DRAFT — deploy-only, stacked on #1594 (`v37/xmr-native-m4harness`). Do not merge ahead of it.**

Found while standing the real 72 h stagenet parity soak up on VM500 (.44), against the synced stagenet daemon. One line of `M4GraduationLedger::save()` is the difference between a soak that can lose three days and one that cannot.

### The defect

```cpp
std::ofstream f(path, std::ios::binary | std::ios::trunc);
```

`SoakDriver` autosaves every `autosave_every` samples — 16 by default. At the soak's P-TPL cadence that is a rewrite of `ledger.json` every couple of minutes, for 72 hours per leg, under a watchdog whose entire job is to kill the node and start it again. For a short window on every one of those saves the file on disk is neither the old ledger nor the new one.

Landing in that window does not cost a sample. `load()` rejects the splice as `"m4 ledger: not valid JSON"`, the next process starts a **new** ledger with both streaks at zero, and the soak carries on looking perfectly healthy — because a streak restarting from zero is exactly what an honest reset looks like, and nothing in the transcript tells the two apart.

### The fix

`save()` writes a sibling `.tmp`, flushes it, `fsync`s it where the platform has `fsync`, and `rename(2)`s it onto the target; the directory entry is fsynced after. The rename is atomic within a directory, so the next reader — the next process, the watchdog, a human with `cat` — sees the whole previous ledger or the whole new one. The fsyncs are what extend that from "survives a crash" to "survives a power cut", and they are `#if`-guarded: without them the rename is still untearable, merely not durable, which is the property the soak actually leans on.

The header stops claiming to be STL-only and says where it is not.

### Scope

`src/impl/xmr/native/parity/xmr_graduation_ledger.hpp`, `save()` and the include block. `to_json()`, `load()`, the four-part key discipline and the re-derivation of state on load are untouched. `xmr_native_m4_soak_kat` covers the persistence round trip through the new path unmodified: **165 checks, 0 failures** on VM500 (gcc 13.4, Release, `XMR_BUILD_RANDOMX=ON`).

This commit is what the live stagenet soak is running (`7ffbac3e`), so leg 1's ledger is keyed to it.

### DIFF
--- a/src/impl/xmr/native/parity/xmr_graduation_ledger.hpp
+++ b/src/impl/xmr/native/parity/xmr_graduation_ledger.hpp
@@ -58,17 +58,29 @@
 // beside it so a reader can see both numbers rather than trust one.
 //
 // SCOPE FENCE: src/impl/xmr/ only. No consensus digest, no src/sharechain/v37.
-// Header-only, STL only, no clock of its own -- the caller supplies unix time,
-// so a KAT drives 72 hours in microseconds.
+// Header-only, no clock of its own -- the caller supplies unix time, so a KAT
+// drives 72 hours in microseconds. STL only apart from save(), which reaches
+// for fsync where the platform has one; see the comment there.
 // ---------------------------------------------------------------------------
 #pragma once
 
 #include <cstdint>
+#include <cstdio>
 #include <fstream>
 #include <sstream>
 #include <string>
 #include <utility>
 #include <vector>
+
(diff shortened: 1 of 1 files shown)


frstrtr:

PR #1609, opened 2026-09-11
xmr(native): the trust anchor counted fees as emission

frstrtr:
**DRAFT — stacked on #1608, which is stacked on #1594 (`v37/xmr-native-m4harness`). Do not merge ahead of them.**

The M4 parity soak was pointed at the synced stagenet daemon on VM500 for the first time. It went to `Revoked` on its first template sample.

## What it saw

Every P-TPL compare differed on `already_generated_coins`, by the same amount, at every height:

```
served (monerod) 18164847101536847281
shadow (native)  18164847090712907281
               =       10823940000
```

`already_generated_coins` is a **sentinel** field, so this did not merely reset the streak — it revoked the graduation key outright. P-TIP, meanwhile, was CLEAN on all nine fields, including cumulative difficulty: the native index had walked the same chain and done the same arithmetic.

## Where the number comes from

It is not arbitrary. It is exactly the daemon's own `fee_amount` for the 848 blocks the anchor generator subtracted when it backed the anchor's `already_generated_coins` out of a later `get_miner_data` reading:

```
get_coinbase_tx_sum(height=2204001, count=848)
  emission_amount 508800000000000     (= 848 × 0.6 XMR, base only)
  fee_amount          10823940000     ← the entire divergence
```

## The premise, and it was written down

```python
# `reward` in a block header IS the coinbase amount (base emission plus fees),
# which is exactly what monerod adds into already_generated_coins per block.
coins = coins_at_md_tip - sum(x["reward"] for x in above)
```

The first half is true. The second is not — `Blockchain::handle_block_to_main_chain` does

```cpp
already_generated_coins = base_reward + already_generated_coins;
```

A fee is existing supply changing hands, not supply being created, and monerod never adds it. Subtracting `reward` removes base+fee per block where only base went in, leaving the anchor short by the fee total — a **constant offset** the native index then carries forward forever, because its own forward accumulation is correct (1556 blocks × 0.6 XMR, to the atomic unit, over the probe run).

## And the cross-check confirmed it

```python
s = emission_amount + fee_amount
if s != sum(x["reward"] for x in above): raise SystemExit(...)
print("coinbase cross-check OK")
```

That is a true identity about two RPCs, and it says nothing about **which of them belongs in the subtraction**. It validated the arithmetic around the wrong quantity and printed OK. It now asserts the decomposition and *uses the emission half*, named as such — a check that would have caught this class rather than blessing it.

## The change

* `xmr_chain_anchor_stagenet.inc`: `already_generated_coins` → `18163913501536847281`, digest recomputed over the corrected body (`e159b23d…`). No format bump, no new field, every other row byte-identical.
* `xmr_anchor_gen.py`: subtract `emission_amount`; the cross-check becomes a decomposition assertion; both provenance comments state the rule instead of the mistake.

## Verification on VM500, against the live synced stagenet daemon

* `xmr_native_anchor_self_check_kat` — **153 checks, 0 failures** (its model bundle is built in-test and is untouched by the value; the embedded case re-derives the digest from the corrected body).
* `xmr_native_m4_soak_kat` — **165 checks, 0 failures**.
* The same soak run, re-run with the corrected anchor: P-TPL CLEAN on all six fields, and the ledger accrues instead of revoking.

This is the first defect the M4 harness found *by being run*, which is the argument for running it against a real chain rather than a fixture.

### DIFF
--- a/src/impl/xmr/native/anchor/xmr_chain_anchor_stagenet.inc
+++ b/src/impl/xmr/native/anchor/xmr_chain_anchor_stagenet.inc
@@ -5,8 +5,10 @@ R"ANCHOR(
 # captured  2026-09-10 19:01Z, daemon tip 2204848, anchor buried 848
 # windows   735 difficulty / 100 short-term / 100000 long-term
 # coins     already_generated_coins from get_miner_data at 2204848 minus the
-#           coinbase of every block above H_a; cross-checked against
-#           get_coinbase_tx_sum over the same range.
+#           BASE EMISSION of every block above H_a -- get_coinbase_tx_sum's
+#           emission_amount, NOT the header `reward`. monerod accumulates
+#           base_reward into already_generated_coins; fees are recycled from
+#           the existing supply and are never added to it.
 # note      monerod publishes no RPC for its compiled-in checkpoints;
 #           `checkpoint` rows are copied by hand at release time and
 #           this capture carries 0 of them.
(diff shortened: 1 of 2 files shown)


frstrtr:

PR #1613, opened 2026-09-12
v37: S-1 live fold-wiring — drive owed_digest emission on live block events [DO NOT MERGE]

frstrtr:
**DO NOT MERGE** — DRAFT. Base `master` @ `ea02b36e`. Opened for review of the wiring and of the one gap it deliberately does not close (S-1c, below).

## The defect

`settle::fold_eb` (`src/c2pool/v37/w4_settlement.hpp`) is the ONE credit-path entry point for the `E_b` entitlement. Until this branch, **nothing live ever called it.**

The only live FOUND site — `XbtcNode::on_block_won` — built *both* the ledger credit and the coinbase payout out of the W5 assembly, and a fresh win (`confirmations == 0`) **withholds** that assembly. So `credit` was empty, FINALIZE did `finalW += {}`, and `owed_digest` sat at the empty `"V37O"` anchor

```
b4db1ded95a73f939975a259f9b48a1d182109f44397ed77e35d624f1a5cf339
```

on every live node, forever, however many blocks the pool won. The pool mined and settled nothing. The two-node regtest below reproduces exactly that at boot, on both nodes, before a single share lands.

## What this branch does

Wires the **caller**. `src/sharechain/v37` is untouched; the `owed_digest` fold body is untouched. The settlement machinery was already live — it simply had no driver.

**`src/c2pool/v37/btc/btc_node.hpp`** — `on_block_won` folds `E_b` out of the lane snapshot the engine has published at the instant of the win:

```
view = engine.snapshot(lane_chain)
f    = settle::fold_eb(reward, *view, /*strict=*/true)
fb.credit = f->credit                             // what the pool now OWES
fb.payout = the W5 outputs the block broadcast    // what it actually PAID
```

`credit` and `payout` are deliberately **different maps**. FINALIZE does `finalW += credit; finalW -= payout`, so `credit - payout` is the carried-forward owed balance — exactly what `owed_digest` commits to. Registering `credit == payout` nets every block to zero; registering the assembly as the credit loses the entitlement. Both shapes are replaced.

**Refuse loud, register anyway.** A real block we mined is a real block: a fold that yields nothing must never cost us the FOUND registration, or the F1 driver never finalizes it and the coinbase it paid is never deducted. No lane snapshot / `fold_eb` refusing a non-ratified geometry / `reward == 0` from a fail-closed backend / an empty lane each stamp the record VALUELESS, log the reason, bump a counter — and hand back an empty credit that is registered regardless.

**Cut witness.** Each win now carries `lane_digest`, lane `version`, lane `incarnation`, `next_pos` (the prefix `P` the fold read at) and `eb_source`. The lane digest is the consensus commitment: two nodes folding at the same lane digest fold the same `E_b`. That identity is the precondition for a convergent `owed_digest`, and an operator can now read it straight off the log.

**`src/c2pool/v37/main_v37_btc_dash.cpp`** — the daemon mirrors the cut and the S1 counters on its FOUND line and on its stop line, beside `owed_digest`. An `owed_digest` still reading `b4db1ded…` after a FOUND is now diagnosable instead of silent.

## KAT

`src/c2pool/v37/test/v37_s1_live_wiring_kat.cpp` — **38 checks, 0 failures**, on a fixed event stream (no clocks, no sockets, no RNG; `MockCoinBackend` + `MemSettleStore`):

1. the empty anchor is what an untouched ledger commits to (baseline)
2. driver-drives-S-1: shares → `on_block_won` → `fold_eb` → FOUND → bury → FINALIZE → `owed_digest` is non-empty and equals a pinned golden `4e13dd3f7472dae1…`
3. determinism/convergence: a second independent stack fed the byte-identical stream reaches the byte-identical digest
4. a fresh win credits `E_b` and broadcasts nothing, so `sum(finalW) == reward`
5. **the defect, reproduced**: the pre-S-1 assembly-as-credit shape → back AT the anchor (this is the regression pin; it fails if the wiring is reverted)
6. the cut witness is the snapshot the fold actually read
7. refuse-loud on `reward == 0`: VALUELESS, counter bumped, FOUND still registered
8. refuse-loud on a non-ratified geometry: `fold_eb` refuses, counter bumped, FOUND still registered
9. a second win over a standing owed balance *does* emit a coinbase — so `credit - payout` carries

Registered with `add_test` **and in BOTH `build.yml` `--target` lists** (the c2pool#1539 / #769 hollow-green rule). `tools/ci/check_test_target_allowlist.py`: `CI drift-guard OK: 301 test target(s) across 11 test tree(s)`.

Local `ctest -R "v37_|f1_finalize"` on workstation-191 (g++-15, RelWithDebInfo): **29/30 pass**. The one failure is `v37_w6_persistence_test (Not Run)` — a **pre-existing** gcc-15 header-hygiene break in `src/c2pool/v37/test/w6_w4_shim.hpp:198` (`std::sort` without `<algorithm>`), in a file this branch does not touch. Flagged, not fixed here.

## Live proof — two-node DASH regtest, real X11 mining

Own rig: one regtest `dashd` pair, two peered `c2pool-v37-btc-dash` daemons (`--p2p-bind` / `--peer`, live `LiveMainchainIndex`), and a real X11 stratum miner driving real shares into each node's stratum. At regtest difficulty every accepted share is also a block, so these are genuine live FOUND events.

```
boot   node A  owed_digest=b4db1ded…cf339      <- the defect, live
boot   node B  owed_digest=b4db1ded…cf339

FOUND  node A  h=141  E_b=0 keys [VALUELESS]   <- first win, empty lane: loud, still registered
FOUND  node A  h=142  E_b=1 keys  cut{v=2 next_pos=1}
FOUND  node B  h=145  E_b=2 keys  cut{v=5 next_pos=4}   <- B credits A's miner too (cross-node lane)
FOUND  node A  h=147  E_b=2 keys  cut{v=7 next_pos=6}
FINALIZE …                                     (d_conf=6, one coin-height per step)

stop   node A  owed_digest=f239774992c0eab9241664733cfc80f7d6cbc41c272287611c22698a05efe58c
               s1{folds=4 valueless=1 refused=0 no_view=0 unresolved=0}
stop   node B  owed_digest=f03bb9b83146cdbe8e1fb52ad2bbaca399f75b7fe45b80c69132832ed00b9ebf
               s1{folds=3 valueless=0 refused=0 no_view=0 unresolved=0}

stop   node A  lane_digest=951ca0fae9bd5f246a628484228a22578562097c7cc69e87e50b0bced4c4e074 version=9 incarnation=1
stop   node B  lane_digest=951ca0fae9bd5f246a628484228a22578562097c7cc69e87e50b0bced4c4e074 version=9 incarnation=1
               (both: raw_total=2048 next_pos=8 identities=2, same two identity keys)
```

Restarting node A on its own settle store replays the durable event log to the **identical** `owed_digest f239774992c0eab9…` — the ledger fold is deterministic from the FOUND/FINALIZE stream.

So, precisely:

* `owed_digest` is **live and non-empty** on both nodes — the S-1 emission works.
* the **lane cut converges byte-identically** across the two peered nodes (`951ca0fa…`, v=9) — the precondition for a convergent `owed_digest` holds.
* the two `owed_digest`s **still differ**, and that is honest: each node registered only its OWN wins (`ledger_seq` 10 vs 6).

## The gap this PR does NOT close — S-1c

Nothing propagates a FOUND event to a **non-winning** peer. `submit_fn` → `BlockEventDriver::on_block_found` → `XbtcNode::on_block_won` fires only on the node whose own stratum accepted the winning share. The carrier relay floods the block-winning *carrier* (`CarrierRelay::append_block_winner`), so peers account the WEIGHT — but nothing turns that into a `FoundBlock` on the receiving side, and the frozen W3-B5 wire carries neither the coin block id nor a settlement-FOUND frame.

Full cross-node `owed_digest` convergence therefore needs S-1c, which has a real design question in it (which lane cut must the receiver fold at, so both sides agree?). The `read-at-version` ring on the engine snapshot surface makes "fold at the version immediately before this carrier" implementable, but it is a new mechanism with its own wire question and its own goldens, and it is not attempted here.

## Safety

* No `src/sharechain/v37` consensus canon touched; no change inside the `owed_digest` fold body. Engine/producer/driver path only.
* No XMR path touched. The sibling XMR defect (`xmr_o2_finalize_connect.hpp:470-472` sets `payout = credit`, netting FINALIZE to zero) is left alone: the XMR lane receives no shares today, so folding `E_b` there would credit nobody, and the change would break a pinned KAT with no live proof behind it. Called out for its own lane.
* `owed_digest` values change wherever a node actually wins blocks — that is the point of the fix, and it is why this is a DRAFT for review rather than a merge.

Diff: 5 files, +223/−12.


frstrtr:
Superseded by c2pool#1618, which carries this fold-wiring commit PLUS the S-1c cross-node convergence fix as one atomic unit (merged to master as 47cb31d5). Landing this PR on its own would put master in the non-convergent S-1 state (the winner folds E_b, peers do not), so it is closed unmerged. Branch v37/s1-live-wiring is kept for history.

PR #1615, opened 2026-09-12
control-plane: live money-gated POST /api/config/apply (#157 Slice A)

frstrtr:
## Slice A (#157) — control-plane apply, dormant → live (money-gated)

Implements **Slice A only** of `docs/design/tx_inject_qt_control.md`. DRAFT — pending independent review + operator tap. Slice B (arm/disarm + submit-raw-tx) is a separate follow-up and is **not** built here.

### What changed
Turns the previously **inert** `POST /api/config/apply` (unconditional 503, pinned by a KAT) into a **fail-closed live write path**.

Fail-closed posture, in layers:
- **Loopback only** (same posture as the GET config endpoints).
- **Dormant by default.** No main installs the apply fn, so production stays `503 {"armed":false}`. Even a wired endpoint stays 503 until an operator registers a per-process **control token** — arming is a reviewed operator-tap step, never a silent default.

Two-tier apply, **atomic** (one bad key rejects the whole batch — nothing applies partially):
1. **Non-money keys** → loopback control token + schema/type/range validation. Unknown / not-applicable / read-only / out-of-range keys are refused with a named cause.
2. **Money-path keys** (any `Mut::MONEY_*` catalog row) → the FULL gate:
   - a **two-phase money nonce bound to the EXACT requested diff** (phase 1 issues a nonce for a specific diff and applies nothing; phase 2 must present the same diff + that nonce — a nonce issued for one diff never confirms another);
   - the server-side **`AddressValidator`** on any address-typed money value (fail-closed on an unparseable address);
   - the **M0 tripwire**, which fires on any attempt to write a money key without a valid confirmed nonce (or with an invalid address).

"Apply" swaps the process-global published `ResolvedConfig` snapshot atomically (the reporting mirror behind `GET /api/config`) and invokes any registered `ParamApplier` setter. It **never touches coinbase/subsidy/PPLNS/payee computation** and **never arms `--embedded-tx-inject`** (that is Slice B, and the catalog marks that flag `RESTART`, not money-path, so this slice does not treat it as money).

### Files
- `src/core/config_endpoint.{hpp,cpp}` — control token, two-phase money nonce (diff-bound, single-use, TTL), tripwire, per-value schema validation, and the armed `apply_config` entry point + JSON request parser.
- `src/core/web_server.{hpp,cpp}` — optional `set_config_apply_fn` seam + `rest_config_apply(body)` (unwired by default → 503).
- `src/core/http_session.cpp` — the `POST /api/config/apply` route: loopback gate → 503 when unwired → otherwise route through the gated apply fn.
- `src/core/test/web_server_config_endpoint_test.cpp` — flips the inert-503 KAT and adds the gated-apply cases.

### Tests (`core_test`, folded into the existing target — no new executable)
Built + run on a build host; **11/11 `ConfigEndpointHttp` pass**. New cases: dormant when unwired; dormant wired-but-unarmed; non-money applies with a valid token (mirror reflects it, `apply_armed` flips true); refuse without token (403); reject unknown key (400, named cause); money write without a nonce issues a confirm and applies nothing (tripwire does not fire on the sanctioned phase-1 request); two-phase nonce must match the diff (a mismatched diff → 409 + tripwire; the matching diff → applied).

### Not done / to polish
- Enum-member and hostport validators fall back to non-empty / shape checks (no per-enum member table exists yet).
- BCH/NMC/BIP110 have no `AddressValidator` blockchain enum, so their addresses fall through the multi-validator (garbage still refused).
- Conflicts textually with #1606 / #1611 in shared files (`http_session.cpp`); rebase later.

DRAFT — do not merge. Requires independent review + operator tap on the money path.

### DIFF
--- a/src/core/config_endpoint.cpp
+++ b/src/core/config_endpoint.cpp
@@ -6,6 +6,12 @@
 
 #include <algorithm>
 #include <cctype>
+#include <chrono>
+#include <cstdint>
+#include <mutex>
+#include <random>
+
+#include "address_validator.hpp"
 
 namespace c2pool::config_endpoint {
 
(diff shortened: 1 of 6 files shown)


frstrtr:
Superseded by #1616 — identical tree (903b70c9), reopened on footer-free commit 894a0b7 to clear the attribution-gate (sanctioned new-branch fix, no force-push). Closing this one.

PR #1622, opened 2026-09-12
[DO NOT MERGE] xmr(native): judge the P-TPL famine constraint by arm identity, not by seat (M4 leg-1 stall)

frstrtr:
**DRAFT — DO NOT MERGE.** Comparator-only fix for the M4 stagenet soak stall; held for integrator review and the operator's redeploy of the soak.

## What the soak showed

The M4 leg-1 soak (`--serve-arm monerod`: serve = monerod, shadow = native; harness from c2pool#1594, ledger fix c2pool#1608, anchor fix c2pool#1609) ran 13 h on a healthy node — synced, 8/8 peers, 0 reorgs, 0 orphans, pool accepting relayed transactions (`acc=190 rej=0`) — and never left `clean_run=0`:

```
by_verdict_this_run  CLEAN 3166 / FAIL 2012 / VOID 9
every FAIL:          seam=TEMPLATE  first_diff="tx_backlog_count served=0 shadow=5"
                     note="1 constraint(s) violated; [P-TPL served_by=monerod]"
                     native_backlog_famine: "native template backlog has been 0 for N
                     consecutive samples while the shadow arm held 5 transaction(s):
                     the native pool is not ingesting relayed transactions"
```

The daemon's mempool was genuinely empty at the time (`get_transaction_pool_stats` `txs_total=0`, `get_miner_data.tx_backlog` absent); the native pool held 5. The FAILs began the moment the native pool first held a transaction the daemon did not (`shadow=1`, 23:46Z) and never stopped.

## Root cause

`ParityOracle::run_template_` (`src/impl/xmr/native/parity/xmr_parity_oracle.hpp`) fed `BacklogFamineGuard` by **seat**, not by **arm**:

```cpp
fi.native_backlog = s.served.tx_backlog.size();        // "native" := whichever arm SERVED
fi.shadow_backlog = shadow.fields["tx_backlog_count"]; // "daemon" := whichever arm was the SHADOW
```

The guard's question (`xmr_backlog_famine.hpp`) is *native pool holds nothing while the daemon holds something, sustained*. Seat and identity coincide only in the serve-native posture. In leg 1 the daemon's empty mempool became "native 0", the native pool's own transactions became "the shadow holds 5", and an ingesting node was refused for "not ingesting". The mirror is worse: a real native famine in leg 1 would have read as *fed* (the served seat's count was non-zero), so the constraint could not catch, in that posture, the regression it was added for.

`tx_backlog_count` itself was never the gate. It is `Regime::Measurement` and is rendered in the diff only; it stays that way. The two counts are different quantities by construction — the daemon's `get_miner_data.tx_backlog` (its own mempool as offered to a template builder) vs. what the native pool would select under the v37 relay rule — and may legitimately differ.

## Fix (comparator only, `src/impl/xmr/native/parity/`)

* `BacklogFamineGuard::Input` names its two counts by arm: `native_*` / `daemon_*` (was `shadow_*`); `native_serving` shapes only the refusal's wording. `BacklogSampleClass::ShadowEmpty` → `DaemonEmpty`; `describe()` says `daemon-empty=`.
* `run_template_` resolves which observation is the native arm and which is the daemon by arm **name** (`is_native_arm_name` / `is_daemon_arm_name`, prefix match on the two real arm names); a pair whose identities cannot be resolved is undecidable, never guessed. Both counts are read from the observations' `tx_backlog_count`, which is `md.tx_backlog.size()` on either arm.
* `COMPARATOR_VERSION` 3 → 4. The table is unchanged; the rule that evaluates its one Constraint row is, so no streak earned under 3 carries. The two exact pins (`xmr_native_parity_kat`, `xmr_native_txpool_parity_kat`) move with it.
* `parity/README.md`: a section on what the backlog measurement and the famine constraint are, and why they are judged by arm.

No change under `src/sharechain/`, no consensus body, no owed_digest. Nothing in the node, the arms, the txpool or the M4 driver/ledger is touched.

## What still catches a real divergence

* The six required EQUALITY rows of P-TPL are untouched and remain sentinels: `prev_id`, `major_version`, `difficulty`, `seed_hash`, `median_weight`, `already_generated_coins` — a template that would build a genuinely different (invalid) block differs in one of these.
* `reward` parity is judged on every block at P-TIP; block acceptance at P-SUB (`daemon_accepted`, `confirmed_on_chain`).
* Per-transaction measurement parity (weight / fee / blob_size keyed by tx id) is the M1 P-POOL seam (`--txpool-parity`), which leg 1 leaves off by design.
* The famine constraint still fires — now in **both** postures — on native 0 / daemon > 0 sustained (K samples and S seconds).

## KAT

`xmr_native_backlog_famine_kat` (already registered: `add_test` + both `--target` lists in `build.yml`) gains **suite G**, the serve-monerod posture:

| | seats | expects |
|---|---|---|
| G1 | monerod serves 0 / native shadow 5 (the live shape) | CLEAN ×20, guard never trips |
| G2 | monerod serves 5 / native shadow 0 (the mirror) | ends FAIL, detail says the native arm is the shadow and does not claim it served |
| G3/G4 | both 0; daemon 50 / native 1 | CLEAN |
| G5 | the same shapes with native serving | as suites C/D |
| G6 | two `native`-named arms; an unknown arm name | undecidable, never a refusal |
| G7 | refusal earned in one seat cleared by a fed sample in the other | guard state is about the pool, not the seat |

With the seat-based rule temporarily restored in the oracle, 10 of the new checks fail (G1 ×2, G2 ×6, G6 ×2); with this rule: **143/0**.

## Verification (soak host, gcc 13.3, Release, `XMR_BUILD_RANDOMX=ON`, scratch clone)

```
xmr_native_backlog_famine_kat   143 checks, 0 failures
xmr_native_parity_kat           833 checks, 0 failures
xmr_native_txpool_parity_kat    210 checks, 0 failures
xmr_native_m4_soak_kat / xmr_native_node_kat / xmr_native_contracts_kat   passed
ctest -R 'xmr_native_(backlog_famine|parity|txpool_parity|m4_soak|contracts|node)_kat'   6/6
```

**Live re-run against the same daemon: not obtainable from a second process on the soak host.** A read-only probe of the fixed `xmr_native_node` (same posture, same `--connect 127.0.0.1:38080`, own ledger) never got a levin peer: the loopback daemon reset every handshake (`read header: Connection reset by peer`), and so did every reachable stagenet peer this box does not already hold. The control settles it: a binary built from the soak's own commit (`b03ce935`, the code that has been connected for 13 h) fails the very same handshakes right now (`OLD vs 127.0.0.1 -> Connection reset`, `OLD vs 194.32.76.244 -> End of file`, `OLD vs 176.9.0.187 -> Connection reset`). This is the environment — the daemon holds 22 outbound stagenet connections, the running soak holds the loopback inbound slot (`127.0.0.1:43726`) and 8 more, and monerod's per-IP inbound limit is 1 — not a regression in the node. The P-TPL evidence on the live daemon therefore comes from the redeploy (the soak's slot frees when the supervisor restarts). What this PR proves offline is exact: suite G1 replays the live sample shape (monerod serving, `tx_backlog` 0 / native shadow 5) through the real `ParityOracle` and it is CLEAN under this rule and FAIL under the shipped one.

## Redeploying the soak

Rebuild `xmr_native_node` from this commit, set `COMMIT` in `env.sh` to it, restart the supervisor. The ledger key moves (commit + comparator v4), so both legs restart from zero — which is where they are. Keep the old `ledger.json` / `samples.jsonl` as the record of the v3 run.

## Out of scope, noted

The native pool holds a handful of transactions the daemon does not (5 at the time of reading). That is expected under the v37 relay rule and is visible as the `tx_backlog_count` measurement; whether those particular transactions would be minable is a P-POOL / P-SUB question, not a P-TPL one, and is not changed here.

### DIFF
--- a/src/impl/xmr/native/parity/README.md
+++ b/src/impl/xmr/native/parity/README.md
@@ -84,6 +84,30 @@ the backlog moved in between, and `SERVED_MISMATCH` — the verdict that catches
 serving something *neither* arm would have produced — would become unprovable.
 The serving arm is still cross-checked, but only while its epoch has not moved.
 
+## The P-TPL backlog: a measurement and a constraint, judged by arm
+
+`tx_backlog_count` is a **measurement** and never scores: the two arms hold two
+different transaction sets by construction. The daemon's number is its own
+mempool as `get_miner_data.tx_backlog` offers it to a template builder; the
+native number is what our pool would select, admitted under the v37 relay rule.
+They differ on ordinary propagation skew, and the native pool may hold
+transactions the daemon has dropped. An equality gate here would refuse a
+healthy node.
+
+The **famine shape** is a constraint (`native_backlog_famine`,
+`xmr_backlog_famine.hpp`): the *native* pool holds nothing while the *daemon*
+holds something, sustained across K samples and S seconds. That is the one
+shape skew cannot explain, and it is the regression the M2h restack was for.
+
+The two counts are the **arms'** own, resolved by arm name, in whichever seat
+(served / shadow) the posture puts them. This is comparator version 4. Version 3
+read "native" off the served seat and "daemon" off the shadow seat, which is the
+same thing only when the native arm serves; in M4's leg 1 (serve = monerod,
+shadow = native) the daemon's empty mempool was judged as a native famine, every
+P-TPL sample on a healthy node came out FAIL, the leg's streak never left zero,
+and — the mirror — a real native famine in that posture would have read as fed.
+A pair whose arm identities cannot be resolved is undecidable, never guessed.
+
 ## Graduation
 
 The ledger is hard-keyed to `{c2pool_commit, comparator_version, monerod_version,
--- a/src/impl/xmr/native/parity/xmr_backlog_famine.hpp
+++ b/src/impl/xmr/native/parity/xmr_backlog_famine.hpp
@@ -56,9 +56,24 @@
 //
 // THE REFUSAL IS STICKY. Once tripped, the check keeps refusing until a sample
 // actually shows the condition resolved -- the native pool holding something,
-// or the shadow holding nothing. A refusal that evaporates on the next
+// or the daemon holding nothing. A refusal that evaporates on the next
 // undecidable sample would be a warning, and this is not a warning: it travels
 // out as a failed CONSTRAINT, which compare_seam turns into a P-TPL FAIL.
+//
+// THE TWO SIDES ARE ARMS, NOT ROLES. The shape is "the NATIVE pool holds
+// nothing while the DAEMON holds something" -- and which of the two arms is
+// serving and which is the shadow is the posture's business, not the guard's.
+// The first cut of this guard read its native count off whatever arm had
+// SERVED and its daemon count off whatever arm was the SHADOW, which is the
+// same thing only in the serve-native posture. In M4's leg 1 (serve = monerod,
+// shadow = native) the two inputs swap seats: an empty daemon mempool became
+// "native backlog 0", the native pool's own transactions became "the shadow
+// holds N", and a node that was ingesting perfectly well was refused at every
+// height for "not ingesting" -- 2,000 P-TPL FAILs on a healthy stagenet node,
+// the graduation streak reset on each one, and (the mirror image) a REAL
+// famine in that posture would have read as fed. So the inputs are named by
+// arm identity, the oracle resolves them by arm NAME, and a pair whose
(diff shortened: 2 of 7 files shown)


frstrtr:
Closing as superseded: the same fix (judge the P-TPL famine constraint by arm identity, not by seat) was merged in #1624 (v37: M4 parity oracle), implemented in src/impl/xmr/native/parity/xmr_backlog_famine.hpp. The branch is kept; nothing is deleted.

PR #1625, opened 2026-09-12
v37: XMR live settlement — X2 lane share-push + S-1b fold_eb (owed non-empty + convergent) [DO NOT MERGE]

frstrtr:
> ## UPDATE — commit `b819937`: the `first_eligible` ARMING RULE, and why S-1b still forked
>
> S-1b made `owed_digest` non-empty and the wire-carry made two nodes agree on
> the money. They still forked on the **age column**. `owed_digest` commits
> `"V37O" || per finalW row : key || i64 finalW || u64 first_eligible`, and
> `rearm_first_eligible()` arms a key at the bin its `EffectiveOwed = finalW −
> Σ(pending payout)` first goes positive. The pending set is deliberately **not**
> synchronised across nodes — tolerating that relay-hop skew is why the v0x03
> wire-carry exists — so two honest nodes handed the **same finalize stream** arm
> the same key at **different bins**, and because arming is sticky their digests
> then differ **forever** at byte-identical balances. Ruled: **arm on the
> FINALIZED HALF** (`finalW > 0`), which has a byte-identical finalize stream
> cross-node.
>
> The edit is inside the consensus fold body, so it is **not taken here**.
> `b819937` lands the evidence and the turnkey. **Read PART IV at the bottom.**

---

**DO NOT MERGE — DRAFT.** Base `master`. Consumer tree only; `src/sharechain/v37`, the `owed_digest()` fold body, `fold_eb` and `settle_block` are byte-untouched.

## The two defects

The Monero daemon mined, submitted and finalized real blocks while its settlement layer stood completely still. Two independent defects held it there.

**X2 — the lane received no shares.** The XMR stratum path was a pure pass-through: `StratumListener::SinkTap` counted an accepted share and logged it, `GatedShareSinkT` decided whether the bytes also cleared the exact 128-bit network rule, and the publisher published. Nothing ever appended a record to the lane. `AddLane` was the only engine write the daemon made in its whole lifetime, so the Monero lane stood at `raw_total = 0` with an empty identity set, and no miner ever acquired a payout identity.

**S-1b — the fold netted to zero.** `xmr_o2_finalize_connect.hpp` registered every win with `credit == payout == { payee : reward }`, so FINALIZE did `finalW += credit; finalW -= payout` and cancelled exactly. `owed_digest()` never left the empty `sha256d("V37O")` anchor `b4db1ded…` — and that same constant is what the option-B coinbase serialises into `tx_extra 0x03` as the merge-mining leaf.

Everything followed from those two facts: `E_b` folded to `{}` at every prefix, the OWED ledger had nothing to commit to, and a peer had nothing to converge with.

## What this does

Mirrors the merged DASH S-1 / S-1c stack (`btc/btc_node.hpp`, `main_v37_btc_dash.cpp` on master) in the XMR consumer tree.

| file | role |
| --- | --- |
| `xmr/xmr_carrier_share_sink.hpp` **(new)** | X2. `XmrLaneShareSink`, a forwarding `IShareSink` inserted **outside** the RandomX exact gate, so the network-block decision is byte-identical while every ACCEPTED share becomes a carrier mint request. Plus `XmrCarrierIndex`, the v37 `IMainchainIndex` over the Monero chain this node follows, with a context horizon. |
| `xmr/xmr_carrier_stack.hpp` **(new)** | The carrier layer the DASH daemon stands up inline, factored so the two families cannot drift: tracker, index, ingest, peer node, relay, send queue, the W3-B5 wire-freeze boot self-check, `--p2p-bind` / `--peer`, and the S-1c receive seam. |
| `xmr/xmr_s1_fold.hpp` **(new)** | S-1b. `fold_at_tip()` / `fold_at_peer_cut()` — **callers** of `settle::fold_eb`, the one W4 S8 credit-path entry point. No fold, no ledger arithmetic, no digest is defined here. |
| `xmr/xmr_o2_finalize_connect.hpp` | Registers `credit = E_b`, `payout = {}`. Adds `offer_peer_win()` (thread-safe, drained on the main thread with a bounded cut-miss retry), the double-drive guard, the on-registered-found hook, and sidecar **v2** carrying the cut. |
| `xmr/xmr_o2_settlement_fixture.hpp` | `IXmrOwedSource` + `XmrLiveOwedSource`: option B's coinbase now proposes over, and commits to, the node's **real** `OwedLedger`. The proof fixture stays for `--owed-demo-amount`, and which one is in use is printed. |
| `carrier_send.hpp` | `OwnWinRequest` gains an optional explicit `PayoutDescriptor`, honoured ahead of `payout_script`. Unset — every existing caller, the whole DASH path — leaves `emit_now` byte-identical. |

### The cut rule

`E_b = fold_eb(reward, view@P)` is a pure function of `(reward, view.payout, view.identities)` at ONE lane prefix. A receiver that folded at its own tip would fold at a strictly later prefix — its tip already includes the block-winning carrier it just admitted — and would credit a different `E_b`. So `fold_at_peer_cut` reads the winner's `(next_pos, spine_digest)` back out of **its own** settlement ring and REFUSES if it cannot find that exact prefix under that exact commitment. Refusing leaves `A != B` visible and counted; folding at a neighbouring prefix would hide a settlement fork behind a plausible number.

### The payout leg, stated

A freshly found Monero block broadcasts **no settled-owed output keyed to a lane identity**: option A's coinbase pays monerod's own `--payout-address`, and option B's K_fair coinbase is assembled before the block is buried. So `payout` is empty by construction — the same place DASH's burial gate lands at depth 0 — and the whole entitlement carries forward as owed. That carry is exactly what `owed_digest` commits to.

### KAT — `v37_xmr_s1b_x2_kat`, 39/0

Registered with `add_test` **and** on BOTH `build.yml --target` lists (the c2pool#1539 hollow-green rule). Stdlib + Threads only: no coin lib, no crypto backend, no sockets, no clocks.

It pins the fix **and reproduces both defects**, so it goes red if either returns:

- `2h` the pre-S-1b shape (`credit == payout`) nets `finalW` to zero and lands back on the anchor
- `3g` a peer with no cut descriptor stays at the anchor while the winner has moved — two nodes, two owed ledgers
- `4b` the cut rule is load-bearing: `fold(P) != fold(P+1)` on this stream
- `5b`/`5c`/`5d`/`5e` each fail-closed refusal as its own bit (cut_miss, cut_mismatch, non-ratified geometry, reward == 0)
- `6a`-`6d` the X2 sink wiring, including that `submit_network_block` is passed through untouched
- `7a`-`7e` the Monero parent is not byte-reversed, and the context horizon bounds it

`--mock-smoke` is 32/32 with its S-1b assertions rewritten: `FC10` now asserts the entitlement reaches `finalW` where it used to assert the netting to zero.

## Honest boundaries

- **One identity, not per-miner.** There is no base58 address decoder in this tree (`xmr_stratum_messages.hpp` says the login address is "decoded downstream" and nothing downstream decodes it). V1 credits every share to the operator's own `--payee-spend-hex`/`--payee-view-hex` pair — which is also the key the FOUND/FINALIZE records use, deliberately the same object so the two cannot drift. Per-miner identity is a follow-on; `AcceptedShare::address` is carried into the carrier tag so the mapping is auditable the moment that decoder exists.
- **Synthetic carrier PoW.** The minted carrier still carries the synthetic RDWR envelope ground to `consensus_lz(parent)`, not the miner's RandomX share: a peer cannot verify the Monero PoW from the frame. What is real is the wire (frozen v0x02), the index (the live Monero chain), the identity, the lane weight, and the cross-node flow.
- **E_b is irrecoverable after a crash in one narrow window.** If a process dies between the sidecar write and the FOUND write-ahead, the next boot has a fresh settlement ring that does not retain the winner's prefix. That block is re-registered with an EMPTY credit and counted as `eb_irrecoverable` with a loud line, rather than re-folded at a foreign cut.
- **Wire flag day.** This build EMITS v0x02 and ACCEPTS `{v0x01, v0x02}` — a v0x01-only peer will reject its frames. Loud on purpose (`wire_freeze::flag_day_id()`), because a mixed fleet must not be able to silently strip a block-winner cut descriptor at a v0x01 hop.

---

## `canonical_coinbase_matches` = (a) — the peer-recompute (this completes the PR)

Rebased onto `master` @ `309496e2` (build.yml + `src/c2pool/v37/test/CMakeLists.txt` resolved as unions, keeping the S-1 / owed-event-MMR / M4 target names).

### The defect that was left open

Everything above converges the two nodes — **until the first block whose option-B coinbase actually SETTLES owed**. From that block the winner deducts a real payout map at FINALIZE (`finalW -= payout`) while the receiver deducts nothing, because the map is unbounded and is deliberately NOT on the frozen v0x02 wire. The receiver's only honest move was to REFUSE the block, so `owed_digest` forked at the first settlement and `refused[payout]` climbed for the rest of the run.

### The ruling, and what this is

The operator ruled **(a)**: the peer RECOMPUTES the winner's K_fair payout from its OWN owed ledger. K_fair is `OwedLedger::propose_coinbase` — oldest-owed-first over `EffectiveOwed`, taken O(K) off the incremental `EffectiveOwedIndex` ordered view (`w4_settlement.hpp` §4.6) — and it is deterministic.

So `xmr/xmr_peer_payout_recompute.hpp` **(new)** does not compute a payout. It **remembers** one. The node already runs `propose_coinbase` once per height for its OWN option-B template (`xmr_o2_settlement_source.hpp`, `KFairSource::W4Propose`), and the Owed-role outputs of that assembled template ARE the payout it would have broadcast had it won that height. `XmrKFairPayoutCache` records that map per height at the instant it was built; `recompute_peer_payout()` hands it back when a peer's win for that height arrives. **A second CALLER of the node's own K_fair run — never a second K_fair rule.** `src/sharechain/v37`, the `owed_digest()` fold body, `fold_eb` and `settle_block` are byte-untouched; nothing is added to the wire.

### Why the cached height, and not a fresh call at receipt

`propose_coinbase` is a pure function of the ledger STATE, and the two nodes share that state at **one instant per height**: both build `h` when their tip is `h-1`, after the finalize of bin `h-1` and before the finalize of bin `h`. A receiver that called `propose_coinbase` when the descriptor lands would call it at whatever state it had reached by then — and on a live rig that is routinely **one finalize ahead**, because it learns of `H_b` from the chain a poll away while the descriptor is a poll plus a relay hop away. FINALIZE moves `EffectiveOwed` by `+credit` and re-arms `first_eligible`, so the late call returns a DIFFERENT map wearing a plausible number. KAT case 6 pins that.

### The guards (fail-closed; a refusal is never a silent credit)

| code | refused because |
| --- | --- |
| `no-record` | we built no template for `H_b` (not serving / rebuild held / aged out) |
| `ambiguous` | we built more than one DISTINCT map for `H_b` |
| `reward` | our template's reward ≠ the winner's → different parent/fees → different budget |
| `digest` | our `owed_digest` at our build ≠ the winner's at the win |
| `empty` | the winner settled and our own K_fair proposed nothing |
| `shape` | a non-positive amount, or Σ payout > reward |
| `reservation` | we learned of ANOTHER block in the window AFTER the build |

### The reservation witness, and why it had to exist

K_fair draws on `EffectiveOwed = finalW − Σ_pending payout`, while `owed_digest` commits to the **finalized half only**. Two nodes can therefore hold the SAME digest and still propose different maps, because one has more blocks RESERVED.

This was not theoretical. On the 2-node regtest the winner mined **two rival blocks at one height and reserved both**, while the receiver had drained only the first of the two descriptors when it built the next height: equal digests, and a proposal **47× larger**. Booking it would have been a silent divergence — the one outcome this path must never produce.

So each record also witnesses its reservation (the finalize cursor at the build, and how many blocks were known in that window) and the lookup re-asks; a count that has GROWN is refused. Two ordering fixes keep it from firing constantly: the existing R-7 "do not build a template over an unbooked FOUND" hold now also covers a **peer** win that has arrived and not been booked, and the rebuild is held — bounded — while a chain height in `(cursor, tip]` still has no descriptor here.

### Live proof — isolated 2-node XMR regtest

Own `monerod` (`--regtest --offline --fixed-difficulty 20000`, own ports and data dir), own `xmrig`, `--coinbase v37`, `d_conf=3`. Node **A** mines; node **B** is a pure peer on the carrier relay.

**Node B reproduced the winner's payout BYTE-FOR-BYTE on nine consecutive settling blocks** (h=6..14), e.g.

```
A: FOUND registered 9a2dcc432c87… h=6  payout=1 keys/35184002991678 pico
B: S-1c PEER FOUND registered 9a2dcc432c87… h=6  payout=1 keys/35184002991678 pico [RECOMPUTED K_fair]
A: FOUND registered d9bac9e2261b… h=11 payout=1 keys/35183667452158 pico
B: S-1c PEER FOUND registered d9bac9e2261b… h=11 payout=1 keys/35183667452158 pico [RECOMPUTED K_fair]
```

**Per-finalize-cursor `owed_digest`, A vs B — EQUAL at cursors 0–5 and 8–14, all of them POST-SETTLEMENT:**

```
cursor=8   EQUAL  A=7c430cf9f53cae7f876e422cdf8695c585cf115e85be012b6440c668c4532c1f  B=7c430cf9…
cursor=9   EQUAL  A=27e6c14b2c06a5a3699046001bb0a798defe4a3a3fcbb02c273af0cdb330a839  B=27e6c14b…
cursor=10  EQUAL  A=d1bc6f972df8dc998b561af0d587d5ee3bacd2f15bc660f10183bed6806922e4  B=d1bc6f97…
cursor=11  EQUAL  A=11105d7e2e54f160b2a492f977d7c46421905bda286f1e1599f20693f78d7a6b  B=11105d7e…
cursor=12  EQUAL  A=2b940afa8e041959a588f1db1652b094deb41dcc5e72e76dcdc796579cbe1472  B=2b940afa…
cursor=13  EQUAL  A=a9d5c66f9e72851fd115b10338890e554f08c53d9c5f490200aae76e29bc3aeb  B=a9d5c66f…
cursor=14  EQUAL  A=9af29e02b5d7b7cca84cb91e860a2fc50b11e6d8171bdfadf8c0bd53c9a4a2c7  B=9af29e02…
```

Node B's counters at that point:

```
K_fair peer-recompute: armed  APPLIED=12  refused[ambiguous=0 reward=0 empty=0 shape=0 reservation=0]
R-7 payout leg: booked=12 | floors: SOLVENCY(finalW) min=0  reservation(EffectiveOwed) min=0
```

`SOLVENCY(finalW) min=0` on **both** nodes is the no-double-pay claim: no block finalized more payout than entitlement.

### Honest boundaries

- **The run then diverges at cursor 15, loudly.** B refuses h=15 on the `digest` guard because its template for that height was built one finalize behind the winner's. That is the designed failure — a refusal, never a silent credit — but it is also **terminal**: ruling (a) has no recovery path, so one unreproducible block forks the two ledgers until an operator resyncs.
- **The last gap is not closable from this side.** `owed_digest` pins the finalized half; the reservation witness pins the part of the reserved half this node can see. A winner that reserves a rival **this node never hears of at all** would not move the count, and `empty`/`shape` are all that stand between that and a wrong map. Closing it needs one more bit from the winner: a commitment to its payout map on the wire (a v0x03 trailer), or the receiver reading the winner's coinbase amounts off the chain it already follows. Both are outside ruling (a) as written and are surfaced here rather than invented.
- **Fleet-identical caps are assumed.** `output_cap` / `h_min` / `sink_min` / the fixed-output set are config, and nothing on the wire pins them.
- **Armed only where it is true.** `--owed-demo-amount` builds the coinbase from a separate proof fixture, so the recompute is DISARMED there and the pre-(a) fail-closed refusal stands.

### KAT — `v37_xmr_peer_recompute_kat`, 40/0

Registered with `add_test` **and** on BOTH `build.yml --target` lists (the c2pool#1539 hollow-green rule). Stdlib + Threads only. It pins: the reproduction is byte-exact and order-exact; four blocks with three of them SETTLING converge through every settlement with `finalW` solvent on both; the pre-(a) defect reproduced in BOTH shapes (refuse, and credit-with-empty-payout); no double-pay across the settlement boundary; every guard as its own bit with a non-vacuous positive control; the cut INSTANT being load-bearing; the reservation witness; the bounded cache; and the NON-settling path unmoved. `v37_xmr_s1b_x2_kat`, `v37_s1c_convergence_kat`, `v37_s1_live_wiring_kat` and `v37_owed_event_mmr_kat` are green beside it.

---

## ★★ SUPERSEDED ON THIS BRANCH: the recompute is replaced by WIRE-CARRY (wire v0x03)

Head `5460c346`, rebased onto `origin/master` (`a0d12429`). **Still DRAFT — do not merge.**

The section above describes the peer-RECOMPUTE leg and closes by naming its own terminal boundary: *"one unreproducible block forks the two ledgers until an operator resyncs… closing it needs one more bit from the winner: a commitment to its payout map on the wire (a v0x03 trailer)."* The operator ruled that bit. This commit builds it.

### What changed

The winner already knows its K_fair owed-deduction map exactly — it is the Owed-role output set of the template it mined (Fixed and residual-Sink outputs are not ledger keys and never enter it), which is the very map its own ledger booked as `PendingRec::payout`. So it **emits that map on the v0x03 carrier trailer**, and the peer **folds it verbatim** through the existing `on_network_block_won -> on_block_found/on_block_finalized` path.

No recompute. No template build to compare against. **No same-instant requirement anywhere.** The two-clock dependency is removed rather than guarded, which is why convergence is now *sustained* instead of eventual.

The recompute stays, demoted to a **non-authoritative local cross-check**. It can only ever say whether this node's own K_fair run for that height happened to agree; nothing refuses on its verdict. That turns the old failure mode into a number an auditor can watch (`WIRE-CARRY xcheck differ` / `no-map`).

### Wire v0x03 — a SECTIONED trailer

v0x03 is the v0x02 frame **verbatim** plus a fixed-order list of optional sections, each introduced by its own present byte, appended after the cut descriptor. Every v0x01 and v0x02 offset stays valid over the v0x03 prefix, exactly as v0x02 kept v0x01's.

```
frame_v3 := v0x02 frame ; drops section ; payout section
drops    := u8 present ; if 1: u16 count ; count x { b32 payee ; i64 credit }
                              ; b32 enrollment_digest
payout   := u8 present ; if 1: u16 count ; count x { b32 identity ; u64 amount }
```

Section 1 is **c2pool#1627's DROPS credit map**, carried here byte-for-byte with its codec unchanged.

> ### ★ WIRE-INTEGRATION POINT (where c2pool#1625 and c2pool#1627 reconcile)
>
> c2pool#1627 (`v37/drops-activate` @ `4a71a635`) defined v0x03 as the v0x02 frame plus the DROPS trailer **alone**, with nothing after it. This branch keeps that section in position 1 and appends the K_fair payout section after it. **The merged layout is the union above**, and the only work at merge is re-pinning #1627's three v0x03 goldens, which each gain exactly **one trailing zero byte** (the absent payout section).
>
> No field moves, no field widens, no section is re-ordered, and v0x01/v0x02 are untouched by either branch. Until that merge a #1627 build and a #1625 build are each other's `REJECT_TRUNCATED` — loudly, which is the correct flag-day behaviour and never a silently half-read settlement trailer.
>
> Sections rather than a TLV, deliberately: settlement data is exactly the thing a node must **not** skip, so an unknown shape stays a whole-frame rejection at a visible version bump. This build has no DROPS ledger leg, so it never populates section 1 and **refuses** a peer win that carries one, rather than crediting E_b while ignoring a credit the winner applied.

### Fail-closed — and what is never a refusal

Refused: a settling descriptor with **no** carried section; a map with no cut; a map on a descriptor whose `payout_emitted` is 0; unordered or duplicate identities; a zero or non-positive row; more rows than `W3_PAYOUT_MAX_ENTRIES` (4096); and a total above the reward the **same frame** carries. Each is a whole-frame or whole-block refusal — never a truncation, never a partial fold.

**A timing skew is never a refusal on this path.** That is the property being bought.

### Live proof — isolated 2-node XMR regtest

Own `monerod --regtest --offline --fixed-difficulty`, own `xmrig`, own ports and data dirs, `--coinbase v37`, `d_conf=3`. Node A mines; node B is a pure peer on the carrier relay.

```
node A (winner)  carrier send (v0x03 section 2): K_fair maps carried=42
                 settling-wins-without-one=0 | R-7 payout leg booked=42
node B (peer)    WIRE-CARRY (v0x03 K_fair): carried=41 refused[absent=0 shape=0]
                 S-1c offered=45 credited=45 cut_miss=0 cut_mismatch=0
                 refused[fold=0 payout=0 late=0]  admit_rejected=0
                 local xcheck agree=37 differ=0 no-map=4
```

`owed_skew=42` — the winner's `owed_digest` at its win differed from B's at receipt on 42 occasions, and **not one** caused a refusal. On 4 settling blocks the local recompute had no comparable map at all and would have refused; wire-carry folded them and converged.

Per-finalize-cursor `owed_digest`, A vs B: **26 cursors compared, 26 EQUAL, 0 DIFFER**, no first divergence anywhere in the run — including past cursor 15, where the recompute run forked. Deepest common cursor 42:

```
A = c575073f2566f04b8fe8645e9cfc0974d5cd367c9bae4d827beb3ff968067352
B = c575073f2566f04b8fe8645e9cfc0974d5cd367c9bae4d827beb3ff968067352
```

**No double-pay:** across 44 settling blocks the rows the winner carried equal the rows the peer folded, block for block, 0 mismatches. `SOLVENCY(finalW) min=0` and `reservation(EffectiveOwed) min=0` on **both** nodes for the whole run.

### Honest boundaries

- **A first run on a heavily loaded host diverged at cursor 25 — not on this leg.** Its wire-carry counters were `refused[absent=0 shape=0]` there too; the break was a cut **miss** in the E_b fold (the winner's lane prefix was never published on B after nine inbound carriers were admit-rejected under load). Re-run on a quiet host: `admit_rejected=0`, `cut_miss=0`, and the 41-block run above. The S-1b cut path's behaviour under executor coalescing is a separate, pre-existing question and is left where it is.
- **The carried map is the winner's assertion about its own coinbase.** A peer cannot verify it against the chain from this frame alone; what it enforces is shape and budget (bounded rows, strictly ascending keys, every amount > 0, Σ ≤ the reward on the same frame). Verifying the map against the on-chain coinbase outputs is the named next fence and is deliberately not invented here.
- **Fleet-identical caps are still assumed** for the *cross-check*, not for the fold — the fold no longer depends on them at all.

### KAT — `v37_xmr_wirecarry_kat`, 41/0

Registered with `add_test` **and** on BOTH `build.yml --target` lists (the c2pool#1539 hollow-green rule). Stdlib + Threads only. It pins: the wire leg (exact round-trip, and a refusal for every shape that could fork a ledger); **≥ 20 consecutive settling blocks** converging byte-equal after *every* block with the peer one descriptor behind the winner; the skew being **real** on 18 of them; the **control**, in which the same stream booked recomputed-at-receipt maps, forking and staying forked; no double-pay; the fail-closed half; and the coexistence of both v0x03 sections in one frame.

Existing suites re-pinned rather than relaxed: `v37_w3_wire_freeze_kat` now asserts v0x03 costs exactly three bytes over v0x01 and moves no v0x01 body byte, and the frozen-wire selfcheck grew to **700 checks** with three v0x03 fixtures (nothing-to-say / cut + both sections / cut + payout only). Full v37 `ctest`: **35/35**.

*Canon untouched: nothing under `src/sharechain/v37`; `owed_digest()`'s fold body, `fold_eb` and `settle_block` are byte-unchanged. Frozen v0x01/v0x02 goldens are byte-unchanged.*

---

# PART IV — `first_eligible` FINALIZED-HALF arming (commit `b819937`)

**Nothing in this part is applied.** The change lands in
`OwedLedger::rearm_first_eligible()` in `src/c2pool/v37/w4_settlement.hpp` — the
consensus fold body — and moves `owed_digest` for any ledger that carries a
pending payout across a finalize. That is a consensus activation and a fleet flag
day, and it is the operator's hand. The exact two-hunk diff, the one sibling
oracle that must move with it, and the apply order are in
[`src/c2pool/v37/S1B-ARMING-TURNKEY.md`](src/c2pool/v37/S1B-ARMING-TURNKEY.md).

## 1. The defect, stated exactly

```
owed_digest = sha256d( "V37O" || for each finalW row, key ASC :
                       key(32) || i64 finalW (LE) || u64 first_eligible (LE) )
```

`rearm_first_eligible(bin_height)` runs once per `on_block_finalized` and arms a
key at the bin its `EffectiveOwed(k) = finalW(k) − Σ_pending payout(k)` first goes
strictly positive. Arming is **sticky**. `EffectiveOwed` reads `m_pending`, and
`m_pending` is per-node by design.

Observed live on the 2-node XMR regtest rig: after an independent restart both
nodes replayed `events=269` from their own W6 store, both reported one owed key
at the same balance at `hw=47 cursor=44`, and the digests still read
`6bd47268…` (A) and `ec91e352…` (B).

## 2. The live fork, reproduced from the rig's own data — and closed

`v37_s1b_arming_kat` replays the **literal** event stream from `rig/settleA/settle.img`
— real identity key `b9d5bd32d700531c…`, real credits, real payouts, real bins —
against a second node handed the **same seventeen events**, with the same finalize
substream in the same order at the same bins, and only two payout-bearing peer
FOUNDs arriving one relay hop earlier.

```
finalW = 70368609960000 piconero, one row, one key, ledger_seq 17 — on BOTH nodes

OLD  A  fe=4  d004509fe8bc58afe6aebe5f861a31ea1ea30785b69996369914e922d3d463d9
OLD  B  fe=5  689d08268d5dd8b6683bc8ff8a2293b66af692f8bbcfebdf2eebccd5c0350b9c   FORK
NEW  A  fe=4  d004509fe8bc58afe6aebe5f861a31ea1ea30785b69996369914e922d3d463d9
NEW  B  fe=4  d004509fe8bc58afe6aebe5f861a31ea1ea30785b69996369914e922d3d463d9   EQUAL
```

Both OLD-rule digests were independently confirmed by driving the **real,
unmodified** `OwedLedger` with the same two arrival orders: the shipped code forks
on this stream today. Across 3000 randomised skews the OLD rule forks **~39 %** of
honest node pairs; the NEW rule forks **none**.

> The literal pair `6bd47268…` / `ec91e352…` is **not** re-derived byte-for-byte:
> the 269-event stores that produced them were overwritten by the rig's next run,
> and the surviving logs record `EffectiveOwed`, not `finalW`, so the preimage is
> not recoverable. What is reproduced is the same defect on the same rig's real
> data, with `first_eligible` the only differing preimage column.

## 3. A from-spec shadow, validated as an ORACLE

`src/c2pool/v37/test/v37_s1b_arming_kat.cpp` implements the owed ledger **from
spec** — its own SHA-256, its own FOUND/FINALIZE/ORPHAN state machine, its own
preimage — and carries **both** arming rules side by side, so the ruling is minted
and proved with **no canon file edited**. Stdlib only: no engine, no coin backend,
no sockets, no clocks, no threads.

Under the OLD rule it reproduces **every** published owed-event-MMR value
byte-for-byte at every cut — `owed_digest`, `ledger_seq`, and the w5
StateCommitment root with the gate OFF **and** ON. That is what makes the re-mint
a measurement rather than an opinion.

27 checks, byte-identical output at `-O0`, `-O1`, `-O2`, `-O3` and under
`-fsanitize=address,undefined` with `detect_leaks=1`.

## 4. The re-mint

Fixture: `oes_v1::schedule_v1()` (`owed_event_mmr_golden_v1.hpp`), chain 7. It is
the **only** shipped owed fixture that carries a pending payout across a finalize,
which is the only way the two rules can differ.

| cut | value | NEW | SUPERSEDES |
|---|---|---|---|
| 21 | `owed_digest` | `dff20133e4824072c224469a1aed460620ddf7d55275aff40e944ff8f0b6bdf7` | `5e3c0cd2ff4db05b…` |
| 21 | `sc_root_gate_off` | `a436d0590f962719165869bc4ea09ed30312755dc9fbb042696b9609159cefec` | `309662e29e0bec55…` |
| 21 | `sc_root_gate_on` | `0379deb8676d1da6b2022d158277bb06363591c8a34bddf7150924b2b1ce8444` | `d45517c7b7187889…` |
| 24 | `owed_digest` | `dff20133e4824072c224469a1aed460620ddf7d55275aff40e944ff8f0b6bdf7` | `5e3c0cd2ff4db05b…` |
| 24 | `sc_root_gate_off` | `3f030a249bd4b36fbca505aba6add56a1cdd92072b2197182b6ec948aac4e487` | `55ff6625312a8d66…` |
| 24 | `sc_root_gate_on` | `cd57f501ae85c7c6fb636c2e49835162fa6dd3c1c91585e1b854fcec1298cbdf` | `b8380f38c64e1c2d…` |

Cuts 0, 6 and 15 do not move. The StateCommitment roots move **because** the
`"V37S"` summary leaf carries `owed_digest` verbatim; the MMR columns do not move
(the record log commits event payloads, which carry no `first_eligible`).

## 5. ★ What does NOT move, proved rather than assumed

**The shape theorem.** If every FOUND carries an **empty payout** and its FINALIZE
is adjacent to it, then `m_pending` holds no payout at any rearm,
`EffectiveOwed ≡ finalW`, and the two rules agree **pointwise** — for any credit
maps, including negative DROPS REPLACE deltas. Case `D1` proves it on **20 000**
randomised streams: zero digest divergences, zero `fe` divergences.

That is the shape of the V37.1 ridge schedule, of **both** DROPS activation mints,
and of the S-1 live-wiring stream. So all of these stand unmoved:

`b4db1ded…` (empty anchor — no rows, so no `fe` column at all; asserted under
**both** rules) · `9cfaf97d…` · `87c5249a…` · `984c7753…` + fold `b03abb1f…` ·
`ae44add2…` + fold `50b5d7d1…` · `4e13dd3f…` · `2479d5b6…` (lane geometry, no
owed input).

> ### ★ A correction worth stating plainly
> `9cfaf97d…` and `87c5249a…` are **owed_digest** values, **not** lane digests —
> `v37_1_ridge_activation_test.cpp` mints both from `OwedLedger::owed_digest()`.
> They are unmoved because of the shape above, **not** because they sit outside
> the owed fold. (The value that KAT reports as a lane digest is `9bad8ea0…`, and
> it is reported rather than pinned.) A reviewer who assumes they are outside the
> fold would have skipped the only argument that actually protects them.

## 6. Registration

`add_test(NAME v37_s1b_arming_kat …)` in `src/c2pool/v37/test/CMakeLists.txt` and
`v37_s1b_arming_kat` on **both** `build.yml` `--target` lists (the Linux x86_64 leg
and the ASan+UBSan leg), so CI compiles and **runs** it instead of registering a
NOT_BUILT CTest sentinel. `tools/ci/check_test_target_allowlist.py` green (308
targets).


frstrtr:
Closing as superseded: the X2 share-push / S-1b fold_eb design was re-made in #1691 and then replaced by the RECON design merged in #1697 (coinbase-authority payout + on-chain credit cut) and #1704 (K_fair arming finalized-only, owed_digest V37O->V37Q). On master, shares reach the lane through the GAP-2 relay (src/c2pool/v37/xmr/relay/xmr_receipt_mint.hpp) and fold_eb runs at the on-chain cut. The branch is kept for reference; nothing is deleted.

PR #1650, opened 2026-09-16
[DO NOT MERGE] xmr(native): wire anchor gate 4 -- network-confirm the bundle before serving

frstrtr:
**[DO NOT MERGE] DRAFT.** Consumer tree only (`src/impl/xmr/native/**`). No consensus digest, no `src/sharechain/v37`.

## The gap

`load_anchor()` (`src/impl/xmr/native/anchor/xmr_anchor_load.hpp`) runs three gates, and every one of them judges an `AnchorBundle` **against itself**: source, form, meaning. Its own banner says so, and says the fourth gate is the caller's:

> `load_anchor()` proves the bundle is well-formed and self-consistent; the caller MUST still fetch the block at `bundle.height` from peers and refuse to start unless it hashes to `bundle.id` (`anchor_confirmed_by_network`).

`anchor_confirmed_by_network()` has existed for that purpose. Nothing called it. Neither `ChainBoot::boot_from_anchor()` nor `NativeNode::start()` asked the network anything, so an anchor was trusted by **provenance alone** — fine for the release-pinned stagenet bundle this project minted itself, a real gap for a mainnet bundle handed to a pool operator on `--native-anchor-path` with value at stake. A fabricated chain, correctly encoded, is indistinguishable from the real one until somebody asks the network.

## What this wires

New `src/impl/xmr/native/node/xmr_anchor_confirm.hpp`:

* **`AnchorNetworkConfirm`** — the judgement. The levin p2p protocol has no "give me the block at height H", so the request is for `bundle.id` over `NOTIFY_REQUEST_GET_OBJECTS` (the same 2003 the genesis boot and the sync driver already use), and both halves of the question are answered from the bytes that come back:
  * the id is **recomputed** from the blob with `consensus/xmr_block_id.hpp` and compared by `anchor_confirmed_by_network()` itself — nothing is taken from the peer's word;
  * the block's **own height** is read out of its coinbase `txin_gen` and compared to `bundle.height`. This second half is why gate 4 is more than one comparison: with a tampered height the id still matches, so the pinned function alone says yes. The KAT asserts exactly that.
* **`AnchorConfirmDriver`** — the bound. At most N distinct handshaked peers, one at a time, each with its own turn, under one wall deadline. Mismatch refuses immediately; an exhausted peer set refuses; an expired deadline refuses. Fail-closed in every direction.

`ChainBoot::boot_from_anchor()` now installs the bundle **and** arms the gate. While it is pending the class withholds everything a peer could act on — inbound blobs are offered to the confirm and then dropped rather than connected onto an unconfirmed trust root, and `get_block_entry` / `find_supplement` answer with nothing. `our_sync_data()` stays truthful and live, because a levin handshake is a precondition of the very fetch the gate needs. `NativeNode::start()` drives the gate after the peer pool has peers and **does not return true until it settles**, so nothing above the node — stratum, templates, settlement — is ever offered an unconfirmed chain.

The existing self-consistency checks are untouched. This is an **additional** confirm.

## Operator surface

`--anchor-confirm-peers` (default 4) / `--anchor-confirm-peer-ms` (12000) / `--anchor-confirm-timeout-ms` (60000) widen the window on a slow link. There is deliberately **no switch that turns the gate off**, and `--probe-only` is no longer an exception to it (an anchor-booting probe now requests exactly one block — the anchor itself); both facts are stated in the code rather than left to be discovered. A refused start drains the node log to stderr, so the operator sees which gate fired.

## KAT

`src/impl/xmr/native/test/xmr_anchor_gate4_kat.cpp` — **115 checks, 0 failures**. Registered with `add_test` and in **both** `--target` lists in `.github/workflows/build.yml` (the #1539 "***Not Run" lesson). Driven by a mock levin GET_OBJECTS peer; no daemon, no sockets.

| case | outcome |
|---|---|
| (a) bundle height/id match a served block | CONFIRMS, and only then is the node ready to serve |
| (b) tampered `bundle.id` | REFUSED (the served block hashes to something else; peer set exhausts) |
| (c) tampered `bundle.height` | REFUSED — and the pinned id comparison is shown **not** to catch it alone |
| (c') peer serves a different-id block | counted foreign, never confirms |
| (d) no peer can serve within the window | REFUSED by exhaustion **and** by deadline — fail-closed, not silent-trust |

Plus: the fetch is bounded (3 requests for a 3-peer budget with 5 peers available; a second poll inside a peer's turn does not re-ask; every request carries exactly the anchor id, unpruned; no chain request is issued), a peer that disconnects loses its slot, a refusal is terminal, and the genesis boot path is untouched (gate 4 stays `Disarmed`).

**Falsifiability check:** with `AnchorNetworkConfirm::ready()` mutated to `return true` (gate disabled), the KAT goes to **22 failures**. It is not vacuous.

## Live evidence — read-only, public stagenet peer

Isolated node from this scratch build, dialling `195.66.213.39:38080` (a public stagenet peer). The `.44` soak node and daemon were not touched.

1. The real `ANCHOR_EMBEDDED_STAGENET` (H_a `2204000`, id `c55f08bce08d3514f520611d6262afb15247e506d58453a73fd48067813dcdf5`) **CONFIRMS**:

```
[GATE-4] anchor confirmed: height 2204000 id c55f08bc...dcdf5 (asked 1, spent 0, blobs 1)
  -- the network served block c55f08bc...dcdf5 at height 2204000,
     which is the anchor this node booted from
```

2. The same bundle with `bundle.id` flipped by one bit **and the digest recomputed** (so gates 1..3 accept it — verified) is **REFUSED**:

```
[GATE-4] anchor refused: height 2204000 id c45f08bc...dcdf5 (asked 1, spent 1, blobs 0)
  -- asked 1 peer(s) for the block the anchor bundle pins and not one served it;
     refusing to start (fail-closed: an anchor nobody on this network has is
     exactly what a fabricated anchor looks like)
```

3. The same bundle with `bundle.height + 1` and the digest recomputed is **REFUSED**:

```
[GATE-4] anchor refused: height 2204001 id c55f08bc...dcdf5 (asked 1, spent 0, blobs 1)
  -- the block that hashes to the anchor id is at height 2204000 on this network,
     but the bundle pins 2204001; refusing to start
```

Independently corroborated out of band: a public stagenet RPC node reports `get_block_header_by_height(2204000).hash == c55f08bce08d3514f520611d6262afb15247e506d58453a73fd48067813dcdf5`.

## Build / test

`workstation-191`, g++-15, `-DXMR_BUILD_RANDOMX=ON`, own scratch clone and build dir. `ctest -R "xmr_native|xmr_levin|xmr_epee"` — **30/30 passed, 0 failed** (including the new target). `c2pool-v37-xmr` and `xmr_native_node` build clean.

## Bound of the claim

Gate 4 proves **the network agrees this block exists at this height**. It does not prove most-work: fork choice and proof of work stay where they are, in C2c, above the anchor. A peer serving a block at all means the block is in its main-chain database (monerod's `handle_get_objects` reads the blockchain and reports anything it does not hold in `missed_ids`), and a peer cannot forge agreement because the id is a hash of the bytes it sent.

---

# Follow-on (same branch): mainnet-readiness plumbing for `c2pool-v37-xmr`

Gate 4 landed on the embedded node. These two commits make the node's mainnet-relevant knobs reachable from the **pool binary**, which is what a mainnet bring-up actually runs. None of it is consensus: it is config and entrypoint plumbing, plus one real bug in the seed path.

**`e62917c9` — plumb the gate 4 window, `--seeds` and a chain snapshot into `c2pool-v37-xmr`**

1. **Gate 4's window.** `--anchor-confirm-peers` / `--anchor-confirm-peer-ms` / `--anchor-confirm-timeout-ms` now exist on the pool binary and are forwarded into `NativeNodeConfig.anchor_confirm`. They only **widen** the gate — an operator on a slow or filtered mainnet link needs to be able to say so, otherwise a peer that answers in fifteen seconds is indistinguishable from a fabricated anchor and the refusal reads as a broken build. There is deliberately **no** knob that turns the gate off, and two guards say so: the parser refuses a zero, and the forward refuses to produce a confirm config that is not `sane()`. Defaults are unchanged when the flags are absent, and a KAT pins the pool binary's defaults equal to `AnchorConfirmConfig`'s own so the two sets cannot drift.
2. **`--native-seeds` (alias `--seeds`).** `p2p::dns_seeds()` existed, was pinned by the locator KAT, and was called by nobody. The node now runs a bounded DNS seed round before gate 4 (the gate needs handshaked peers to ask), takes IPv4 results only, and plants them in the peer store as `PeerSource::DnsSeed` so the dial plan's netgroup caps and rotation still apply. A resolver that answers nothing or hangs falls back to the compiled-in IP seeds; a cold start cannot hang on it. `--seeds` and `--native-connect` work together, and either one now satisfies the p2p-first address-source precondition. Field note: as of this writing all four mainnet DNS seed hosts answer NXDOMAIN, so the IP-seed fallback is load-bearing, not decorative.
3. **`--native-snapshot-path` (+ `--native-snapshot-every`).** The chain index has carried `save_snapshot()`/`load_snapshot()` since C2 and nothing ever wrote one to disk, so every restart re-walked from the anchor. The new `node/xmr_chain_snapshot_store.hpp` is the disk half: an envelope that **binds the image to the anchor identity gate 4 just confirmed** (magic, version, network, anchor height, anchor id, length, sha256 over all of it), written through a sibling temp and a rename. It is not a second trust root — the image is read only **after** the confirm settles, and every failure (absent, corrupt, truncated, foreign network, foreign anchor, wrong magic or version, a length that lies) falls back to the anchor boot rather than to a weaker check. A start gate 4 **refused** arms nothing, so it cannot overwrite a good image either.

**`171fc3c2` — let `--seeds` actually dial in a probe, and pin which seed table mainnet selects**

- `NativeNode::start()` assigned `pc.use_seeds` **after** the `probe_only` arm, silently making that arm's own `pc.use_seeds = false` dead — and the surviving behaviour was the worst of the two: seeds on, but a dial budget sized by the number of **pinned** peers, so `--net mainnet --seeds --probe-only` with no `--connect` did zero dials and reported what looked like an unreachable mainnet. That probe is the pre-flight before a mainnet bring-up, so it now gets a small fixed seed allowance on top of its pinned peers; the allowance is constant and the plan's netgroup caps and rotation still apply, so a probe still cannot wander into peerlist strangers.
- KAT check `C5` joins the two halves nothing joined before: `resolve_nets(...).wire` is the net handed to `dns_seeds()` and to the peer store, so a mainnet pool seeded off a test network's table would have left every other check in the file green. It also pins the regtest consequence — regtest shares the **mainnet** wire net, which is the standing reason `--seeds` is opt-in and off by default.

## Tests

`v37_xmr_mainnet_readiness_kat` — **159 checks, 0 failures**, registered with `add_test` and in **both** `build.yml` `--target` lists (the "***Not Run" lesson). Five suites: the flags parse and a malformed one refuses; every knob reaches `NativeTemplateConfig` and no configuration disarms gate 4; the mainnet seed sets, the address-source precondition and the table `--net mainnet` selects; the snapshot's disk round trip to the same tip, frontier and template inputs plus each corruption refused; and gate 4 still fail-closed through the pool binary's own config. Mutating the forward (`use_seeds`, the peer bound) or the seed selection reds exactly the checks that name them, with exit 1 and no crash.

Also still green on this branch: `xmr_native_anchor_gate4_kat` 115/0, `xmr_native_chain_index_kat` 1717/0, `xmr_native_chain_locator_kat` 109/0, `xmr_native_anchor_self_check_kat` 153/0.

**Gate 4 through the pool binary, re-proved after the plumbing:** with an anchor bundle and one unreachable peer, `c2pool-v37-xmr` announces the forwarded window (`2 peer(s), 1000 ms each, 4000 ms in total`), asks nobody because nobody handshakes, and **refuses to start** — *"fail-closed: an unconfirmed trust root is not a trust root"* — rather than serving.

Still **[DO NOT MERGE]**: this is a draft for review, and the mainnet dry-run is the operator's call.

---

## `c36ff68e` — resume the snapshot THROUGH the boot, so a resumed node stops re-IBDing

The persistence added above **worked and then undid itself**, and the log said so without explaining it. `NativeNode::start()` loaded the image straight into the chain index, announced READY, and left `ChainBoot` with `booted_ == false` — so the **first inbound blob** went to `try_seed_()`, which seeds row zero with `ChainIndex::seed_direct`, and `seed_direct` **RESETS the index**. A live regtest node printed

```
[snapshot] RESUMED at height 600 ... no re-IBD from the anchor
[tip] height=1 ... [tip] height=600
randomx hashes=601 verified=600
```

— the resumed chain destroyed, re-fetched and re-hashed block by block, with the pool status reading `native tip feed: events=300 best=0` a moment after READY. The same un-booted gate also made the serving half advertise `current_height 1` / `top_id = genesis` no matter what the index held, which is what got the whole chain handed back to us.

**The fix is structural, not an ordering tweak.** `ChainBoot` gains `resume_from_snapshot()`, which installs the image **and** records that the boot is discharged; `load_snapshot_()` calls that instead of `index_.load_snapshot()`, still on the verify thread. Because `on_objects()` and `on_chain_entry()` test `booted()` first, `try_seed_()` is then **unreachable**, the serving reads forward to the index (advertising the resumed tip and a resumed locator), and `SyncDriver`'s booted predicate flips, so its first ask carries the resumed locator instead of asking for genesis.

**Gate 4 is not weakened.** The resume still runs strictly after `run_anchor_confirm_()`, and the check is repeated *inside* `ChainBoot` rather than left to `start()`'s line order: an anchor that is installed and **not yet confirmed** refuses the resume. A refused gate 4 still returns before the load, so `snap_armed_` stays false and the session that could not confirm its anchor cannot write its index over a good file. Absent, corrupt, foreign-network and different-anchor images all still fall back to the ordinary boot path with `booted_` untouched.

`try_seed_()` also gains a **belt**: it refuses to seed row zero over an index that already holds rows. The real pre-boot path has an empty index, so it is never taken; it is there so a future misordering is non-destructive rather than silent. It is deliberately **not** in `ChainIndex::seed_direct`, which the KATs and the parity replay re-seed on purpose.

### Tests

New target `xmr_native_snapshot_resume_kat` — **141 checks, 0 failures** — registered with `add_test` and in **both** `build.yml` `--target` lists. It is the only test target that starts the whole `NativeNode` (no peers, no seeds, no daemon, no RandomX linked), which is what lets it make the two claims that only the assembly can support:

- a genesis boot over a good snapshot comes up **at the snapshot's frontier** and **not owed a genesis seed** — the "never announce READY over an index the next blob would wipe" pin;
- a **gate-4 refusal arms no saver**: the good file on disk is **byte-identical** after the refused start, and the KAT checks that it was gate 4 that refused (anchor installed, confirm state `Refused`) rather than an earlier gate.

At the `ChainBoot` seam it measures the re-IBD itself: a model verifier's hash counter must read **zero** across the resume, across a re-delivered genesis blob, and across the whole re-delivered prefix (every block comes back `Duplicate`), then **one** for the first genuinely new block. The **control** lives in the same file and keeps the old wiring visibly broken (boot still owed its seed, still advertising height 1), so none of the above can pass for the wrong reason. Plus the five fail-closed refusals (empty, corrupt, truncated, foreign network at the index; foreign network, foreign anchor and a tampered digest at the envelope) and the gate-4 ordering, proven both ways — refused while pending, resumed after the confirm, at a frontier **above** the anchor height.

Mutation-checked: reverting the node wiring reds 2 checks, removing the gate-4 guard reds 3, and not discharging the boot on a successful load reds 7 — each with exit 1 and no crash.

Also still green with this change on the same build: `xmr_native_node_kat`, `xmr_native_anchor_gate4_kat`, `xmr_native_anchor_self_check_kat`, `xmr_native_chain_index_kat`, `xmr_native_reorg_follow_kat` (6/6 via ctest), and both consumers still build: `xmr_native_node` and `c2pool-v37-xmr`.

Consumer tree only: `src/impl/xmr/native/**`. No consensus digest, nothing under `src/sharechain/v37`. Still **[DO NOT MERGE]**.


frstrtr:
Superseded by #1692 — a clean recommit of these three pieces (gate-4 network-confirm run-path, periodic snapshot/resume through ChainBoot, mainnet --seeds) onto current master. This branch is ~75 commits behind; closing in its favour. Branch retained (not deleted).

PR #1652, opened 2026-09-16
build(test): link BlockchainAddressValidator into test_config_endpoint

frstrtr:
## Base repair -- unblocks #1649 (LTC) + #1651 (DGB)

**Symptom.** `test_config_endpoint` fails to link on Linux x86_64 with undefined references to `c2pool::address::BlockchainAddressValidator`, `DecodeBase58Check`, and `core::log::Logger::m_categories`. This one base break reds two lanes (#1649 LTC, #1651 DGB) plus `dash-daemonless-smoke`.

**Origin of the reference (attribution).** The target compiles `core/config_endpoint.cpp`, which since **894a0b79** (`control-plane: make POST /api/config/apply live and money-gated`, **#157 Slice A**) instantiates `BlockchainAddressValidator` at `config_endpoint.cpp:675`. The `test_config_endpoint` target added at **854336be** (#1648) never carried the validator translation unit or its transitive symbols. This is a base-repair of that gap, not a lane change.

**Fix (test-build only).** In `src/c2pool/CMakeLists.txt`, for the `test_config_endpoint` target only:
- add sources `core/address_validator.cpp` (defines the validator) and `core/log.cpp` (defines `Logger::m_categories` used by the validators category `LOG_*` path);
- carry `btclibs` (for `DecodeBase58Check`) + `Boost::log`/`Boost::log_setup` explicitly, matching the established "does not link core (web_server/stratum SCC tangle), carry Boost.Log explicitly" idiom already used by the object-lib targets in this file.

No production code, no coin-tree files touched.

**Verification.** Linux x86_64: target links; `ctest -R test_config_endpoint` -> **1/1 Passed**. Commit GPG-signed.

_Note: the diagnosis handed to me was sources-only (add `address_validator.cpp`); that alone still fails to link because the validator pulls in base58 + category-log symbols, hence the added `log.cpp` source and the btclibs/Boost.Log link deps._


frstrtr:
Superseded by 68fe18c6d ("build: give the config-endpoint KAT the address validator it actually calls"), already on master via the #1658 merge (0f36fe311) — the BlockchainAddressValidator link this PR added is present on master, so this change is redundant and now CONFLICTING. Closing as superseded; branch ltc-doge/fix-test-config-endpoint-link left intact per the no-delete rule.

PR #1653, opened 2026-09-17
[DO NOT MERGE] xmr(settle): CARROT/FCMP++ version gate on the W5 coinbase (SCAFFOLD; fence stays up)

frstrtr:
**[DO NOT MERGE] — DRAFT.** Held for operator review. Consumer-tree only; canon untouched; the CARROT fence is **not** lifted.

## What this does

The W5-XMR coinbase derives its output one-time keys with the **pre-CARROT** Monero recipe (deterministic `r`, `R = r*G` in `tx_extra` `0x01`, `P_i = H_s(8 r A_i || i) G + B_i`, 1-byte view tag, exact-sum with a residual sink). CARROT rewrites that derivation, so the recipe would produce an **invalid, unspendable** coinbase under the CARROT fork. Until now that was a flat fence: `build_coinbase()` refused any `major_version` above `XMR_PRECARROT_MAX_MAJOR_VERSION`.

This replaces the flat fence with a **version gate**, and nothing else:

| `monero_major_version` | regime | arm |
| --- | --- | --- |
| `<= 16` | `CoinbaseRegime::PreCarrot` | `build_coinbase_precarrot()` — the old body, unmoved |
| `>= 17` | `CoinbaseRegime::Carrot` | `build_coinbase_carrot()` — **scaffold, refuses** |

Both arms re-check the regime themselves, so calling an arm directly cannot bypass the gate, and the dispatcher fails closed on any value outside the exhaustive enum.

**Net behaviour change for a v17 block: none.** It was refused with `CarrotFence` before and it is refused with `CarrotFence` now — no outputs, no keys, no prefix. Only `BuiltCoinbase::detail` is richer, so an operator reading a log learns *why*.

**Net behaviour change for a v16 block: none, to the byte.**

## Why SCAFFOLD and not an implementation

Verified against **primary sources** on 2026-09-17 (several aggregator articles claiming FCMP++ "activated in Q1 2026" are contradicted by the code and the release history and are simply false):

* `monero-project/monero` `master` **and** `release-v0.18`: `src/hardforks/hardforks.cpp` mainnet table ends at `{ 16, 2689608 }` (v16, 2022-06-30). There is no v17 row.
* `master` has **no** `src/carrot_core` and **no** `HF_VERSION_CARROT` / FCMP++ constants in `src/cryptonote_config.h`.
* Newest upstream release is the v0.18.5.x line (CLI 0.18.5.1, 2026-07-08). No v0.19 / FCMP++ release.
* CARROT + FCMP++ live on the `seraphis-migration/monero` fork (`fcmp++-stage`), where `HF_VERSION_FCMP_PLUS_PLUS = HF_VERSION_CARROT = 17` — but the mainnet rows `{ 17, 2689609 }` / `{ 18, 2689610 }` are explicitly commented *"Mock values for tests"*.
* Upstream integration is open or draft: `carrot_core` `monero-project/monero#9559`, FCMP++ integration `monero-project/monero#9436`, `carrot_impl` `monero-project/monero#9697`, RandomX V2 + PoW commitments in v17 `monero-project/monero#10038` (contested).
* No mainnet fork height and no fork date announced.

Decisively: **there are no published reference vectors for a CARROT *coinbase* enote**, so no derivation could be pinned as conformant. Implementing one now would mean inventing conformance — and an approximately-right coinbase does not fail a test, it pays the block reward to keys nobody can spend.

So the CARROT arm ships as a typed, testable seam that **cannot emit an output key by accident**: `src/impl/xmr/settle/xmr_carrot.hpp` carries the transcribed constants, the derivation steps that *are* captured from the reference implementation, an explicit seven-item gap list of what is *not*, and refusing stubs guarded by `carrot::DERIVATION_IMPLEMENTED = false` plus a `static_assert` that fires if anyone flips the flag while the stubs still refuse. The seam is header-only on purpose: seven CMake targets already compile `settle/xmr_coinbase.cpp`, and a new TU would have to be added to every one of them.

## Canon untouched

`XMR_PRECARROT_MAX_MAJOR_VERSION` still lives in `src/sharechain/v37/v37_descriptor_xmr.hpp` and still reads `16`. The consumer-side constant is *derived* from it, and the KAT asserts canon itself has not moved. Lifting it is an operator ruling this PR does not make.

## Proof the v16 arm did not move a byte

`src/impl/xmr/test/xmr_carrot_gate_kat.cpp` (`add_test` name `xmr_carrot_gate_kat`, registered on **both** `build.yml` legs) pins a complete v16 coinbase — `r`, `R`, every one-time key `P_i`, every view tag, the merge-mining root, `tx_extra`, the whole serialized `miner_tx` prefix, the prefix hash and the coinbase tx hash — against goldens captured from the pre-gate tree. Mainnet Monero is at `major_version` 16 today, so that is a **today-mainnet** coinbase, not a hypothetical one. Payout targets are the official `monero-project` `tests/crypto/tests.txt` consensus points.

Independently of the KAT, the same input vector was run through `origin/master`'s unmodified `build_coinbase()` and through the gated tree. The two dumps are identical:

```
sha256(pre)  = 0b4a10a81cef4a014c697575352a26284db6e0a6882e1edcad8c014929073887
sha256(post) = 0b4a10a81cef4a014c697575352a26284db6e0a6882e1edcad8c014929073887
```

The KAT also pins the other half: every `major_version` 17..255 refused with `BuildError::CarrotFence` and nothing emitted (239/239); each arm refuses on its own; the seam stubs refuse and **wipe** their outputs rather than leaving a half-built enote; the canon fence is still 16.

## Verification

workstation-191, g++-15, standalone targeted compile of the coinbase + KATs (pure crypto — no RandomX, monerod, libsodium or boost):

* pre-gate vs gated coinbase dump — **byte-identical**, sha256 equal
* `xmr_carrot_gate_kat` — **OK, 0 failures**
* `xmr_coinbase_kat` (pre-existing regression, incl. its own fence-trip case) — **OK, 0 failures**
* `-Wall -Wextra` clean

## Operator seam — the decisions this PR does NOT make

1. **The deterministic anchor.** CARROT's `anchor_norm` is sender-chosen randomness; `monerod` picks it at random, which a pool cannot, because every node must re-derive the identical coinbase byte for byte. This is the CARROT-regime analogue of `derive_tx_secret_key()` and it is a **v37 design ruling, not a transcription**. Choosing it wrongly is a silent privacy failure (Janus-style linkage), not a build error. P2Pool's own choice is recorded in the seam doc for reference.
2. **Lifting the canon fence** `XMR_PRECARROT_MAX_MAJOR_VERSION`.
3. **Flipping `carrot::DERIVATION_IMPLEMENTED`** — only with (a) a derivation matching a tagged upstream release, (b) pinned reference vectors, and (c) 1 and 2 above.
4. **Adding the seam's `.cpp`** and the seven source lists that will then need it.

Full write-up: `docs/xmr-lane/carrot-coinbase-seam.md`.


frstrtr:
### CI note — the red Linux leg is pre-existing on the base commit, not this branch

`build.yml` / **Linux x86_64** fails at the **Build tests** step with:

```
config_endpoint.cpp: undefined reference to `c2pool::address::BlockchainAddressValidator::BlockchainAddressValidator()`
config_endpoint.cpp: undefined reference to `c2pool::address::BlockchainAddressValidator::validate_address_multi(...) const`
gmake: *** [Makefile] test_config_endpoint Error 2
```

The identical failure, in the identical target, is on the **base commit** `854336be` (master) — run `35134255540`, job `104922535662`. Nothing in this branch touches `config_endpoint`, `BlockchainAddressValidator` or any linkage near them; the diff is `src/impl/xmr/settle/`, `src/impl/xmr/test/`, `docs/xmr-lane/` and two `--target` lists.

Because the build dies at `test_config_endpoint`, CI never reaches `xmr_carrot_gate_kat`, so the CMake wiring is proven locally instead — a real configure of this tree with the CI's own conan profile and options (`ci/conan/linux-gcc13.profile`, `-DCMAKE_BUILD_TYPE=Release -DCOIN_DGB=ON -DXMR_BUILD_RANDOMX=ON`), then a targeted build and `ctest` of exactly the two affected targets:

```
[100%] Built target xmr_carrot_gate_kat
[100%] Built target xmr_coinbase_kat
1/2 Test #1599: xmr_coinbase_kat ......... Passed
2/2 Test #1600: xmr_carrot_gate_kat ...... Passed
100% tests passed, 0 tests failed out of 2
```

`tools/ci/check_test_target_allowlist.py` is green too (304 targets across 11 test trees, all present in the `build.yml --target` allowlist), and the **Test allowlist drift-guard** and **attribution-gate** checks both passed on this branch.

Green checks on this branch: attribution-gate, Test allowlist drift-guard, PR path filter, Param Catalog Tripwire, heavy-leg-lock-test, test-floor-guard-honesty.

PR #1656, opened 2026-09-17
[DO NOT MERGE] v37 Stage 1 supply: unified FrameVault + carrier repair control channel (serve / fetch / verify)

frstrtr:
**[DO NOT MERGE] — DRAFT.** Stage 1 of the convergence-repair track: **SUPPLY**.

**STACKED.** Base is `v37/carrier-relay-robustness` (frstrtr/c2pool#1655, head `5636d8f1`), which itself sits on master `22ef1a11`. Review/merge frstrtr/c2pool#1655 first; this PR's diff is only its own commit.

---

## The gap this closes

No node in this tree can answer *"which carrier sits at lane position P, and what were its bytes?"*. Verified on this checkout:

| structure | file | has an ordered (pos -> carrier hash) row? |
|---|---|---|
| `::v37::L0Slot` = `{pos, w_raw, w_scaled, miner, flags, version, origin_bin}` | `src/sharechain/v37/v37_lane.hpp` | no hash at all |
| `::v37::LaneRecord::Push` = `{chain, desc, w_raw, flags}` | `src/sharechain/v37/v37_lane_executor.hpp` | no hash at all |
| `SettlementView` / `LaneSnapshot` | `src/c2pool/v37/v37_engine.hpp` | `next_pos` + `digest`, no hashes |
| `RelaySeenSet` | `src/c2pool/v37/w3_relay.hpp` | unordered `set<bytes32>` |
| `MemShareTracker` | `src/c2pool/v37/carrier_ingest.hpp` | `identity -> set<hash>`, unordered |
| w6 `CarrierRec` | `src/c2pool/v37/w6_persistence.hpp` | keyed by share hash, ordered only by a backward `prev_hash` walk — **and `main_v37_btc_dash.cpp` never instantiates W6** |
| frstrtr/c2pool#1655 re-offer buffer | `src/c2pool/v37/w3_relay.hpp` | hash-keyed, ~60 s, fed only from `do_broadcast()` |

So a node that has fallen behind cannot even **name** the carriers it is missing.

## What this adds

**1. `src/c2pool/v37/frame_vault.hpp` — one unified bounded FrameVault.** Verbatim CarrierWire frames, written once per admitted carrier, with two indices: **by lane POSITION** (the `(chain, pos) -> carrier hash` row that exists nowhere else) and **by carrier HASH**. Retained back to the lane **WINDOW horizon** (W, default 8640 positions) so an ordered prefix `[a, P)` can be served. Memory bounded three ways, explicitly — `max_entries` 16384, `max_bytes` 32 MiB, `horizon_positions` 8640; whichever binds first, binds, and a frame larger than `max_bytes` is never retained. The frstrtr/c2pool#1655 re-offer now rides this store as a **shallow recent sub-view**: same bytes, one copy, every re-offer option and counter keeping its existing meaning.

**2. The `>= 0x80` control-opcode channel.** `frame[0]` is split permanently — `0x01..0x7f` CarrierWire versions, `0x80..0xff` control opcodes — and `carrier_net.hpp` demuxes `frame[0] >= 0x80` **before** `handle_inbound`.

| opcode | frame | meaning |
|---|---|---|
| `0x80` | `GETORDER` | your ordered carrier ids for `[a, P)` at a cut `(chain, P, spine_digest)` |
| `0x81` | `ORDER` | the answer: `{pos, id}` pairs in strictly increasing position order, plus the server's asserted spine digest at the prefix it actually served |
| `0x82` | `GETFRAMES` | bytes for a bounded id set |
| `0x83` | `FRAMES` | the answer |

`ORDER` is the structurally necessary response half of `GETORDER`; the three opcodes the brief named are all present.

**Backward tolerance, both directions.** An old peer runs a `0x8x` frame into `CarrierWire::decode`, gets `REJECT_BAD_VERSION` — a *verdict*, not an error — and **keeps the socket**; `reader_loop` breaks only on a short read or an over-long length prefix. A repair-aware node that meets an opcode it does not implement ignores-and-counts it. Neither side can be disconnected by a control frame it does not understand, so this deploys piecemeal with no flag day.

**3. Verification, fail-closed.** The requester recomputes each served frame's carrier hash **from the bytes** and requires it to equal the id it asked for. Any hash mismatch, missing id, duplicate id, id never asked for, undecodable frame, or no answer inside `request_timeout` rejects the **whole** response: nothing is handed on, the failure is counted per peer, the peer is not dropped.

**4. Per-peer DoS bounds.** Token bucket per peer on the serve side (over budget gets **nothing** back — zero amplification); exactly **one outstanding request per peer** on the fetch side, so an answer can only ever satisfy a question actually asked; hard caps on every reply (`kCtrlMaxIdsPerOrder` 4096, `kCtrlMaxIdsPerFetch` 64, `kCtrlMaxReplyBytes` 4 MiB) that an asker can only lower; LRU-bounded peer tables on both sides. All peer state keys on a `PeerId` that is fresh per connection and never reused.

**5. Flap safety.** The re-offer attempt cap used to **erase** the entry, so a peer that connected and dropped repeatedly armed forced sweep after forced sweep, burned every entry's attempt budget, and deleted the only retained copy of carriers a *different*, well-behaved peer still needed. It now retires from the **sub-view only**; the bytes stay in the vault. A peer cannot insert into the vault at all, so it cannot push another peer's bytes out by volume either.

**6. `SO_SNDTIMEO` + drop-slow-peer — closing frstrtr/c2pool#1655 Defect-B.** That PR gave each connection its own write lock but did not bound the write itself. Broadcast could tolerate that; the serve path cannot — a peer can request a multi-frame reply and then never read, pinning a reader thread. Every connection now gets a bounded write time (10 s default). A write that times out mid-frame leaves the stream **desynced**, so the peer is dropped **hard** (`shutdown`, then removed) rather than left live with half a frame on it.

## What this deliberately does NOT do

It **supplies and verifies bytes**. It does not apply them: no admit of a fetched frame, no `settle::fold_eb`, no `btc_node.hpp` edit. Under ruling A the served order is a peer's verifiable **assertion**; making it trustless by replay-to-digest is **Stage 2**. `CS-2` and `CS-8` assert the requester's lane stays at `next_pos == 0` while holding verified bytes.

## Zero consensus movement — proven

- `git diff --name-only` against **both** master `22ef1a11` and frstrtr/c2pool#1655 head `5636d8f1`, restricted to `src/sharechain/v37`: **0 files**. `w3_wire_freeze.hpp` untouched.
- The opcodes are **not** wire versions: `kAcceptedVersions` is still exactly `{0x01, 0x02}`, and `CS-8` asserts every value `>= 0x80` is refused by both `version_accepted()` and `CarrierWire::decode`.
- Frozen wire goldens byte-identical: `v37_w3_wire_freeze_kat` **409/0**.
- owed_digest goldens byte-identical: `v37_1_activation_test` **8/8** (gate-OFF `owed_digest` == the pristine-master golden; gate-ON == the V37.1 golden; golden stamp `cffdfd30…` verified), `v37_w4_owed_incremental_test` **14657/0**, `v37_owed_event_mmr_kat` **174/0**, `v37_w4_settlement_test` **899/0**, `v37_s1_live_wiring_kat` **38/0**, `v37_s1c_convergence_kat` **51/0** (which re-runs the full v0x01 + v0x02 golden set, 409 checks, inside itself).
- A full serve -> fetch -> verify round-trip moves neither lane digest nor `owed_digest` on either side.

## Tests

New suite `src/c2pool/v37/test/v37_convergence_supply_kat.cpp` — **374 checks, 0 failures**, green plain and under **ASan+UBSan**, five consecutive clean repeats.

- **CS-1** the unified vault: position index in lane order, hash index to verbatim bytes, the re-offer as a sub-view (one copy, not two).
- **CS-2** `GETORDER` -> `GETFRAMES` -> verify round-trip over real sockets: exact ordered ids, exact bytes, and the requester's lane unmoved.
- **CS-3** a **lying peer** (right id, wrong bytes) is fail-closed on the recomputed hash — and the honest frame in the same reply is discarded too.
- **CS-4** a **missing id** is fail-closed; a **silent peer** is bounded by `request_timeout` and is *not* dropped.
- **CS-5** vault entry / byte / window-horizon caps, an `a` below the horizon refused rather than answered with a gapped prefix, the per-peer token bucket, one-outstanding-per-peer, and an unsolicited answer discarded.
- **CS-6** a flapping peer exhausts its own re-offers but **cannot evict another peer's retained bytes**; the order and the frames are still servable afterwards.
- **CS-7** backward tolerance both ways, including an **old peer reproduced as the verbatim pre-split reader loop over a raw socket**: it meets two opcodes, keeps the socket, and admits the very next carrier frame on the same connection.
- **CS-8** zero consensus movement (above).

Registered with `add_test` **and** on **both** `build.yml --target` lists (x86_64 leg and ASan+UBSan leg), so CI compiles and **runs** it rather than registering a `NOT_BUILT` CTest sentinel.

## One existing test changed, and why

`v37_carrier_relay_robustness_kat` **RR-5** tagged its synthetic frames with bytes `0xA0..0xA3`, which the new demux correctly routes to the control channel instead of `set_inbound`. That is the namespace split working, not a transport regression. The tags move to the body range (`0x40..0x43`) with a `static_assert` pinning them below `0x80`; the property RR-5 pins is unchanged. **145/0**, plain and under ASan+UBSan.

## Files

- new: `src/c2pool/v37/frame_vault.hpp`, `src/c2pool/v37/carrier_supply.hpp`, `src/c2pool/v37/test/v37_convergence_supply_kat.cpp`
- modified: `src/c2pool/v37/carrier_net.hpp`, `src/c2pool/v37/w3_relay.hpp`, `src/c2pool/v37/test/v37_carrier_relay_robustness_kat.cpp`, `src/c2pool/v37/test/CMakeLists.txt`, `.github/workflows/build.yml`

## Known limits (honest)

- The served **order** is an assertion this stage only *verifies at the byte level* (`hash(bytes) == id`). A peer can still assert an order that is not its real lane order; catching that is Stage 2's replay-to-digest, by design.
- A hash-only vault entry (a frame broadcast without being admitted — the `append_block_winner` path) carries no position and, sitting at the front of the deque, stops the horizon sweep until the entry/byte caps evict it. Memory stays bounded by those caps; it is documented in `frame_vault.hpp`.
- The `ORDER` reply's `spine_digest` comes from an injected probe. When the daemon has no published projection at the served prefix, `have_spine` is 0 and the field is zeros — never a guessed digest.

---

## Follow-up commit `be0415cd` — the reply ceiling, and how a big fetch now finishes

**The defect, measured on this branch at `f1286d18` before the change.** `kCtrlMaxReplyBytes` was **4 MiB** while `carrier_net.hpp` refuses any frame over `kMaxCarrierFrame` = **1 MiB** by *breaking the socket* — it has to, because the length prefix is how it finds the next frame boundary. So an honest `FRAMES` answer that happened to be large did not merely fail; it cost the honest server the honest peer that asked. A probe over real sockets on the pre-fix head:

```
kMaxCarrierFrame   = 1048576
kCtrlMaxReplyBytes = 4194304
20x64KiB FRAMES encoded = 1311452  (over ceiling: YES)
10x64KiB FRAMES encoded =  655732  (over ceiling: no)
send 10x64KiB (655732 B): ok      -> B.n_peers = 1
send 20x64KiB (1311452 B): ok     -> B.n_peers = 0   <== honest server hard-dropped
```

It is reachable on ordinary traffic: the canon's F-2 aux slot takes up to `0xffff` refs and the descriptor is carried in the frame, so a carrier frame in the tens of KiB is well-formed, and the PoW preimage does not cover the descriptor, so such a carrier costs nothing extra to mine.

**1. The budget is derived, not chosen.** `kCtrlMaxReplyBytes` is now `kMaxCarrierFrame` minus the `FRAMES` fixed header, minus the worst-case per-frame overhead (36 B x `kCtrlMaxIdsPerFetch`), minus the worst-case un-servable-id list (32 B x `kCtrlMaxIdsPerFetch`), minus a 4 KiB margin held back so a future field cannot silently eat into it — **1,040,111 B**. Two `static_assert`s pin it, one for `FRAMES` and one for `ORDER`, so a later bump of either id cap fails to compile instead of walking a reply over the ceiling in silence. The serve path reads all three bounds through clamps (`clamped_reply_budget` / `clamped_frames_per_reply` / `clamped_ids_per_order`), so `SupplyServeOptions` can only ever make a reply **smaller**; the old 4 MiB value cannot be restored through the options struct.

**2. A big fetch completes by continuation.** A budget alone would have turned a dropped peer into a stalled fetch, so `FRAMES` now carries `status` and a `cursor`: how many of the requested ids this answer consumed. The server serves whole frames in the order asked until the budget is spent, then answers `TRUNCATED` with the cursor it served through; the requester verifies that prefix **exactly as before** — every id in it delivered and hash-bound, or the whole prefix is refused — and automatically re-asks from the cursor until the ask is exhausted. Truncation narrows *what is claimed*; it never softens the check.

**3. A frame that alone exceeds the ceiling is named.** No chunking can ever deliver it, so omitting it would read as `MISSING_ID` and stall the caller for ever. It comes back in an explicit `unservable` list, reported through both a dedicated callback and `SupplyFailure::UNSERVABLE_ID` — distinct from a missing id, not a stall, and not a peer drop. The rest of the same answer is still delivered.

**The loop cannot spin.** Progress is monotone by construction: an absent or un-servable id advances the cursor by itself, and the byte budget can only bind once a frame is already in, so a truthful `TRUNCATED` always has `cursor >= 1`. A peer that answers `TRUNCATED` having consumed nothing is refused as `MALFORMED`, and `max_continuations` bounds the rounds regardless.

`carrier_net.hpp` additionally refuses a targeted send over the ceiling and counts it (`sends_refused_oversize()`) rather than writing a frame the receiver must break on. That path is unreachable from the serve side now; it is a backstop so a regression surfaces as a counter instead of as a mysterious disconnect.

### CS-9 (new, in the same registered KAT)

- **(a) arithmetic** — the largest reply the encoder can build (64 un-servable ids **and** a full payload) encodes under `kMaxCarrierFrame` and round-trips; `encoded_size()` predicts `encode()` exactly; the same for a maximal `ORDER`; and an option set to 4 MiB is clamped.
- **(b) completion** — 20 carriers with ~64 KiB frames, whose bytes total **more than the whole ceiling**, asked for in **one** `request_frames` call: all 20 arrive verified, across chunks, `fetches_completed == 1`, `truncated_replies >= 1`, and — the assertion that fails on `f1286d18` — **`n_peers` is still 1 on both sides**.
- **(c) un-servable** — a retained frame alone over the budget, asked for alongside a normal one: the normal one is delivered and verified, the other comes back as `UNSERVABLE_ID` with its id, `missing_id == 0`, the fetch completes, no slot is left pinned, and neither peer is dropped.
- **(d) loop safety** — a peer that answers "truncated, served nothing" is failed as `MALFORMED` and is asked **exactly once**.

### Verification

- `v37_convergence_supply_kat` **453 checks / 0 failures**, plain and under **ASan+UBSan** (leak detection on).
- `v37_carrier_relay_robustness_kat` **145/0**; `v37_w3_wire_freeze_kat` **409/0 + 17/0** — the frozen `v0x01` / `v0x02` goldens are byte-identical.
- Post-fix probe on the same sockets: budget `1040111`; the over-ceiling frame is now **refused** by `send_to` and `n_peers` stays **1**.
- Consensus untouched: `git diff 22ef1a11..HEAD -- src/sharechain/v37` is **0 files**; opcodes stay `0x80..0x83` and absent from `kAcceptedVersions`; no apply, no fold, consumer tree only.

### Known limit added by this commit (honest)

The `0x83` body gained `status`, `cursor` and the `unservable` list while `ctrl_version` stays `0x01`. That is a layout change to an **unreleased** control channel (this branch is DRAFT and unmerged, and the opcodes are not a wire version), so no deployed peer can see it — but a node built from the earlier commit on this branch would read the new `FRAMES` as malformed. Anything that pins the control layout before merge should bump `kCtrlVersion` instead.


frstrtr:
Landed via #1658: the whole convergence stack (Stage-1 supply — unified FrameVault + GETORDER/GETFRAMES/FRAMES fetch — and Stage-2 apply — on-demand cut re-derivation) merged to master as 0f36fe31. This PR's content is entirely in master (0 files differ). Closing as superseded; the branch is left in place, not deleted.

PR #1658, opened 2026-09-17
[DO NOT MERGE] v37 Stage 2 APPLY: repair the cut a peer block names (2-node owed_digest convergence over a dropped share)

frstrtr:
**[DO NOT MERGE — DRAFT]** Stage 2 of the convergence track. Stacked: base is
`v37/convergence-supply` (frstrtr/c2pool#1656, the carrier supply channel),
which sits on `v37/carrier-relay-robustness` (frstrtr/c2pool#1655), off master
`22ef1a11`.

## The defect

S-1c folds the winner entitlement at the lane prefix `P` the winner names, read
back out of **our own** settlement ring. That works exactly as long as our order
at `P` is the winner order. Drop one share and it is not:

| what happened | what the arm sees |
|---|---|
| we dropped a share the winner admitted | our lane is a push short, `P` was never published here → `cut_miss` |
| we got it out of order | we publish `P` with other bytes → `cut_digest_mismatch` |

Both refusals were terminal. The block-winner descriptor is one-shot, the flood
echo is suppressed by the W2 dedup window, and nothing could re-drive it — so the
two owed ledgers stayed forked for good. That is root cause (2) of the 09-13
sustained-convergence finding.

## What this does

Both bits are the same question — *is my order at `P` the winner order?* — and
the answer is not in the owed ledger. It is on the wire.

1. **Repair at the arm** (`btc_node.hpp`). At the refused arm the node asks a
   repair source for a projection at exactly `(chain, P, spine_digest)`. If one
   exists it folds there; if not it refuses and counts **exactly as before**.
2. **Replay, digest-gated** (`carrier_repair.hpp`, new). The repair fetches the
   winner ordered carrier ids over `[0, P)` and their frame bytes over the
   Stage-1 supply channel (every frame hash-verified against the id asked for,
   fail-closed), then replays them into a **scratch** `V37Engine` through the
   existing w6 slow path (`ReplayDriver::canonical_records` →
   `PrefixResolver::resolve`), admitting each frame through **our own** W2 over
   **our own** mainchain index. The result is taken **only** when the digest
   reached at `P` is byte-equal to the winner own `cut_spine_digest`.
3. **S3 re-drive** (`block_event_driver.hpp`). A verified repair re-enters the
   same arm with the held descriptor, which now folds. A repair that cannot
   verify refuses again and the block stays uncredited.
4. **Daemon wiring** (`main_v37_btc_dash.cpp`). The daemon stands up both halves
   of the supply channel for the first time (they existed; nothing instantiated
   them), binds the #1655 `set_on_peer_connect` + `set_reoffer_dedup_probe` and
   the #1656 `set_on_unservable`, instantiates the FrameVault-backed
   `ISharechainReader` + `ReplayDriver`, and arms a repair on every repairable
   refusal.

### The structural pre-check

The ORDER answer carries lane positions, and a carrier occupies as many
positions as it pushes — so `pos[i+1] - pos[i]` must equal the push count we
derive locally. A server that reorders, omits or invents an entry is caught on
arithmetic, before any digest is computed.

### The bound, stated

A digest at `P` commits to the whole prefix, not to a suffix, so the replay
starts at lane position 0. A node can only be repaired back as far as some peer
still retains position 0 (the FrameVault window horizon). Past that the ORDER
answer is `BELOW_HORIZON` and the repair refuses rather than quietly repairing a
suffix. The stop line prints the horizon this node can serve.

## Proof — `v37_convergence_apply_kat` (new, 45 checks, 0 failures)

Two in-process nodes over **real sockets**, one missing share:

* **CA-1** the defect reproduced: `cut_miss` → REFUSED → `owed(A) != owed(B)`,
  B still at the empty `b4db1ded…` anchor.
* **CA-2** repair over `cut_miss`: 9/9 carriers fetched and verified, one replay,
  reaches the winner commitment at `P`, one S3 re-drive →
  **`owed_digest(A) == owed_digest(B)` byte-equal**, `E_b` equal key-for-key,
  both finalize at the same `bin_height == H_b + d_conf`.
* **CA-3** repair over `cut_digest_mismatch` (the same share delivered late):
  converges to the **same** digest CA-2 reaches.
* **CA-4** a commitment no honest replay reaches: still REFUSED, `owed(B)` moves
  by zero, counted.
* **CA-5** a server that cannot serve the order: refused, counted, no credit.
* **CA-6** the pre-check: lying position delta → `PUSH_COUNT`; missing bytes →
  `MISSING_FRAME`; a suffix → `ORDER_GAP`; wrong digest → `DIGEST_MISMATCH`.
* **CA-7 / CA-8** zero consensus movement, and the `first_eligible` arming
  question measured rather than assumed (see below).

Deterministic across 3 runs; ASan+UBSan+LSan clean — one real use-after-free was
found and fixed that way (the resolver destroys its engine seam on return).

Regression, all green on this head: `v37_s1c_convergence_kat` 51/0 (golden
`4e13dd3f…` unmoved), `v37_s1_live_wiring_kat` 38/0,
`v37_carrier_relay_robustness_kat` 145/0, `v37_convergence_supply_kat` 453/0,
`v37_w3_wire_freeze_kat` 17/0, `v37_btc_digest_drift_kat` 0 failures.
`main_v37_btc_dash.cpp` compiles (full semantic pass, g++-15).

## Zero consensus

* `git diff --name-only 22ef1a11..HEAD -- src/sharechain/v37` → **0 files**.
* `w4_settlement.hpp`, `w3_wire_freeze.hpp`, `v37_engine.hpp`, `w5_coinbase.hpp`,
  `owed_event_log.hpp` → **0 files changed** vs master, so `settle::fold_eb` and
  `OwedLedger::owed_digest()` are byte-identical and no frozen wire or owed
  golden moves.
* The repair changes **which** `SettlementView` `fold_eb` reads. It does not
  change the fold, the ledger, or when `on_block_finalized` fires.

## `first_eligible` arming — NOT applied, and why

`rearm_first_eligible` arms on `EffectiveOwed` (the unsynchronised half) and
`owed_digest` commits to `first_eligible`, so the two arming rules can disagree
when two nodes hold different **pending payout** maps across a settling block.
On this path they cannot: `btc_node.hpp` mirrors the winner burial gate exactly
on the receive side (`gate.canonical = false`, `confirmations = 0`), so both
sides register the FOUND with an **empty** payout map, `EffectiveOwed` equals
`finalW` on both, and the two rules coincide. CA-8 asserts this rather than
assuming it. The S1B arming turnkey is therefore **not required** for the repair
to converge and is **not** in this branch — it stays a canon flip, operator hand.

## Files

| file | |
|---|---|
| `src/c2pool/v37/carrier_repair.hpp` | new — reader/store/seam adapters, `replay_to_cut`, `RepairDriver` |
| `src/c2pool/v37/btc/btc_node.hpp` | the repair at the cut-miss arm + `set_repair_source` |
| `src/c2pool/v37/btc/block_event_driver.hpp` | S3 re-drive + held repairable descriptors (bounded) |
| `src/c2pool/v37/main_v37_btc_dash.cpp` | supply channel + repair driver instantiation |
| `src/c2pool/v37/test/v37_convergence_apply_kat.cpp` | new KAT |
| `src/c2pool/v37/test/CMakeLists.txt`, `.github/workflows/build.yml` | `add_test` + BOTH `--target` lists |

---

## Follow-up (821a90df): a liveness defect in the driver, fixed

The supply channel allows **one outstanding request per peer**, and on the
two-node shape there is only one peer — the very peer whose wins we are
refusing. So *a second distinct cut armed while a repair is in flight against
that peer* is the ordinary case, and it used to **destroy the first repair**:

1. `arm()` wrote `m_by_peer[peer]` **before** asking, so a second arm `k2`
   overwrote the binding of the live job `k1`;
2. `k2`'s own ask came back refused (one request per peer), and the `finish(k2)`
   that followed erased `m_by_peer` **by bare peer id**.

`k1`'s ORDER reply then had nowhere to land, `k1` sat in `m_jobs` forever, and
every later arm of that cut **coalesced into the dead job** — permanently
unrepairable, silently, no refusal logged, the two `owed_digest`s forked for
good. Fail-closed, and fatal to convergence.

**The invariant restored:** an in-flight repair never loses its peer binding
because another cut was armed or finished.

* `arm()` binds the peer **only after** `request_order()` accepted and
  dispatched — a locally refused ask neither inserts a binding nor erases one;
* a cut whose only peer is already serving is **QUEUED** by key (no binding, no
  job, no coalesce) and re-armed the moment the peer falls free. The queue is
  bounded (64); over the bound the cut is refused cleanly and stays re-armable
  by a later S3 re-offer, and a queued cut whose peer disconnects is dropped
  rather than left aimed at a dead peer;
* `finish()` releases the binding **only when it is still its own**, keyed by
  job identity rather than by peer.

New counters on `RepairStats`: `deferred`, `resumed`, `deferred_drop`. New
read-only accessors `deferred()` and `bindings()`. Existing fields, `arm()`'s
signature, and the daemon's stats line are unchanged (compile-proved against
the exact consumer surface `main_v37_btc_dash.cpp` uses at lines 726 and
1039-1062).

### CA-9 — the trigger, asserted

The winner commits to **two** prefixes (P=6, P=9) and wins a block at each; the
second node lost one share and matches **neither** (`cut_digest_mismatch` at the
first, `cut_miss` at the second); both refused wins are armed back-to-back
against the one peer with the first fetch **deliberately held in flight**. CA-9
asserts the live job keeps its binding, the second cut is queued rather than
coalesced or zombied, it is resumed exactly once, both replays reach the
winner's own digest, both blocks credit, and `owed_digest(A)` equals
`owed_digest(B)` byte-for-byte over both settled blocks.

| run | result |
|---|---|
| `v37_convergence_apply_kat` (this branch) | **64 checks, 0 failures** — deterministic over 3 runs |
| same KAT, both halves of the fix ablated | **64 checks, 13 failures** — incl. CA-9f (binding lost), CA-9k (zombie survives), CA-9l (0 repaired), CA-9r (owed forked) — deterministic over 3 runs |

CA-9 runs the supply channel over a hand-pumped `SendFn` instead of the socket
transport, so the in-flight window can be held open on purpose. Every object —
`SupplyService`, `SupplyRequester`, `RepairDriver`, both `XbtcNode`s — is the
real one; only the wire's *timing* belongs to the test. CA-1..CA-8 still run
over real sockets, unchanged.

### Regression set (workstation-191, g++-15, all green)

`v37_convergence_apply_kat` 64/0 · `v37_convergence_supply_kat` 453/0 ·
`v37_carrier_relay_robustness_kat` 145/0 · `v37_s1c_convergence_kat` 51/0 ·
`v37_owed_event_mmr_kat` 174/0 · `v37_a2_multinode_test` 50/0 ·
`v37_w4_settlement_test` 899/0. **1836 checks, 0 failures.**

### Zero consensus movement

* `git diff 22ef1a11..821a90df -- src/sharechain/v37` → **0 files**;
* `src/c2pool/v37/w4_settlement.hpp` (`settle::fold_eb`,
  `OwedLedger::owed_digest`) md5 `e69e668b7083` on master **and** on head —
  byte-identical;
* `src/c2pool/v37/w3_wire_freeze.hpp` md5 `526240fed597` on both —
  byte-identical, and CA-7a's frozen v0x01+v0x02 selfcheck is green;
* the empty-fold anchor `b4db1ded…` is unmoved (CA-7d);
* the fix changes only *when a repair is allowed to start*. It does not change
  which bytes a verified repair folds, or the fold.

### Working-tree scope of this commit

`src/c2pool/v37/carrier_repair.hpp` ·
`src/c2pool/v37/test/v37_convergence_apply_kat.cpp` ·
`src/c2pool/v37/test/CMakeLists.txt` (comment only — the `add_test` and both
`build.yml --target` lists were already in place and are unchanged).

---

## `a7e8450d` — the cross-thread bind window, and the timeout nobody drove

The re-verify of `821a90df` found that the fix there was only half of the
liveness story. Both halves are closed here. Consumer tree only.

### 1. `arm()` binds `m_by_peer[peer]` **before** `request_order()`

`821a90df` moved the binding to *after* the ask returned, on the argument that
the reply is delivered later by the same thread that armed. That argument does
not hold: `arm()` is also reached from the block-event path, from
`drain_deferred()` on a completing repair, and from the daemon's S3 re-drive,
none of which is the transport reader thread. Between the dispatch and the bind
an ORDER reply for that very job can land, find no binding for the peer in
`on_order()`, and be **dropped** — the job then sits in `m_jobs` with nothing
left that can ever answer it. The same zombie as before, entered from the other
side.

The binding is now written inside the arm's critical section, before the ask.
Both other rules are **kept**, and they are what make that safe:

* **the deferred-queue gate** — a second cut whose only peer is already serving
  (ours per `m_by_peer`, or the channel's per the requester's slot) is queued in
  `m_deferred` *without* a binding and *without* a job, so the new write can
  never land on a live job's binding;
* **the identity-guarded `finish()` erase** — `finish()` erases
  `m_by_peer[peer]` only if it still maps to **this** job key, so if
  `request_order()` then refuses, the `finish()` it triggers rolls back exactly
  the binding this arm wrote and touches nobody else's. That is the rollback.

### 2. `SupplyRequester::tick()` is driven from the daemon

`tick()` expires an outstanding fetch slot on deadline and fires
`on_fail` → `RepairDriver::on_fail` → `finish()` → `drain_deferred()`. Nothing
in `main_v37_btc_dash.cpp` or `carrier_send.hpp` called it, so **no deadline was
ever read**: a repair whose reply never arrived stayed in flight forever. Fail
closed — owed never moved — but permanently, silently, and with every cut queued
behind that peer stuck behind it.

`CarrierSendQueue` gains `set_idle_tick()`, called on the same idle cadence the
worker already uses to drive the relay's re-offer sweep (`reoffer_tick_interval`,
default 1 s), outside the relay mutex and never on the emit path. The daemon
binds `supply_fetch->tick()` there. No new thread, and `carrier_send->stop()`
joins that worker before anything the hook touches is torn down.

### 3. The `[v37-dash] repair stop:` line

now carries `queue{deferred, resumed, deferred_drop, queue_depth, in_flight,
bindings}`. A non-zero `deferred_drop` or a non-empty `queue_depth` at stop is a
cut this node wanted repaired and never got.

### 4. KAT — `v37_convergence_apply_kat`, **85 / 0** (was 76 / 0)

Same target, same `add_test`, same two `build.yml --target` lists; no
registration changed.

* **CA-10 — the reply beats the bind.** The control `SendFn` delivers each frame
  into the other node *immediately*, so the whole ORDER / GETFRAMES round trip
  runs **inside** `B.repair->arm()`'s `request_order()` call. Ablation (binding
  restored to after the ask, everything else identical): **7 failures**
  — CA-10c/d/e/f/g/h/i — the reply is dropped, the job is a zombie, the block is
  never registered and the two `owed_digest`s stay forked. With the shipped
  ordering: green, and `owed(A) == owed(B)` byte-equal.
* **CA-11 — the repair nobody answers.** A repair armed against a peer that
  never replies, with a second cut queued behind it. A tick *before* the deadline
  expires nothing (this is the state a stranded repair stays in forever when
  nothing drives `tick()`); past the deadline the slot expires, the repair is
  counted REFUSED, the queued cut is released and re-armed, the resumed repair
  times out in its turn, and **no job, no binding and no queue entry is left**.
  The cut stays re-armable by a later S3 re-offer, and `owed_digest(B)` is
  unmoved by three timed-out repairs.
* **CA-9 stays green**, in the shipped build *and in the ablation* — confirming
  the original orphan is closed by the queue gate and the identity erase, not by
  the bind position, so moving the bind did not re-open it.
* `v37_convergence_supply_kat` **453 / 0**, `v37_carrier_relay_robustness_kat`
  **145 / 0**, g++-15. `main_v37_btc_dash.cpp` compiles clean.

### 5. Zero consensus movement (re-checked at `a7e8450d`)

* `git rev-parse <base>:src/sharechain/v37` == `git rev-parse HEAD:src/sharechain/v37`
  → `fcc0bba339104fe819801093d29bb850112886ac` — the whole canon subtree is the
  **same tree object**; `git diff 22ef1a11..a7e8450d -- src/sharechain/v37` → 0 files;
* `src/c2pool/v37/w4_settlement.hpp` (`settle::fold_eb`,
  `OwedLedger::owed_digest`) md5 `e69e668b70833da9b4249de7561bfba5` on master
  **and** on head — byte-identical;
* `src/c2pool/v37/w3_wire_freeze.hpp` md5 `526240fed5971e9c150551c254d4c671` on
  both — byte-identical; CA-7a's frozen selfcheck green;
* the empty-fold anchor `b4db1ded…` is unmoved (CA-7d, CA-11i);
* the change decides only *when a repair may start* and *when one is allowed to
  die*. It does not change which bytes a verified repair folds, or the fold.

### 6. Files touched by `a7e8450d`

`src/c2pool/v37/carrier_repair.hpp` · `src/c2pool/v37/carrier_send.hpp` ·
`src/c2pool/v37/main_v37_btc_dash.cpp` ·
`src/c2pool/v37/test/v37_convergence_apply_kat.cpp`.

### 7. Still the operator's hand — not in this PR

The `first_eligible` arming rule (root cause 1 of the 09-13 sustained-convergence
finding) is canon and is **not** touched here; CA-8 still only *measures* whether
it bites on this path. `--network testnet` soak on VM100 is unchanged. This PR
stays **DRAFT / DO NOT MERGE**.


frstrtr:
### CI link failure — diagnosed, fixed, and it is a pre-existing `master` break

`Build tests` was red on both Linux legs (and in `coin-matrix / dash-daemonless
smoke`, which builds the `all` target) with:

```
config_endpoint.cpp:(.text+0x74d5): undefined reference to
  `c2pool::address::BlockchainAddressValidator::BlockchainAddressValidator()'
config_endpoint.cpp:(.text+0x74ec): undefined reference to
  `c2pool::address::BlockchainAddressValidator::validate_address_multi(...)'
collect2: error: ld returned 1 exit status
```

**Failing target:** `test_config_endpoint` (`src/c2pool/CMakeLists.txt`). It is
the only target that compiles `src/core/config_endpoint.cpp` directly rather
than through the `core` object library.

**Origin: not this stack.** The same undefined reference fails identically on
`master` — verified both in CI and by a local build of `test_config_endpoint`
at `28fac44b`:

| commit | Linux x86_64 | same ld error |
|---|---|---|
| `5636d8f1` (head of frstrtr/c2pool#1655) | failure | yes |
| `28fac44b` (merge of frstrtr/c2pool#1655) | failure | yes |
| `f9468da4` (current `master`) | failure | yes |
| `a7e8450d` (head of this PR) | failure | yes |

So `master` is red today and frstrtr/c2pool#1655 did not in fact merge green on
that leg.

**Root cause.** `test_config_endpoint` compiles `config_endpoint.cpp` directly
on purpose, keeping a lean config-only link shape. That was sound while
`validate_apply_batch` was a dormant oracle. `894a0b79` ("control-plane: make
POST /api/config/apply live and money-gated", frstrtr/c2pool#1616, merged at
`85a3ed45` on 09-13) added a server-side address check to the money gate that
constructs a `BlockchainAddressValidator` and calls `validate_address_multi()`.
That definition lives in `src/core/address_validator.cpp`, which was never on
this target's link line. The coin binaries link `core`, so they resolved it; the
break surfaced only here.

**Fix** (`68fe18c6`, build wiring only): add the two TUs the call actually needs
— `address_validator.cpp`, plus `log.cpp` for the `Logger::m_categories` that
the `LOG_DEBUG_*` category gate inside the validator reads — and the libraries
they need: `btclibs` for `DecodeBase58Check` on the base58check paths, and
Boost `log`/`log_setup`/`thread`/`filesystem` for the logger. No source file is
touched.

**Verification** (workstation-191, g++-15, Release, `COIN_DGB=ON`,
`XMR_BUILD_RANDOMX=ON`):

- full `cmake --build` of the **`all`** target: exit 0, zero `undefined
  reference`, zero `ld returned` in the whole log
- `test_config_endpoint`, `v37_convergence_apply_kat`,
  `v37_convergence_supply_kat`, `v37_carrier_relay_robustness_kat`: 4/4 pass
- canon regression set `v37_w3_wire_freeze_kat`, `v37_owed_event_mmr_kat`,
  `v37_s1c_convergence_kat`, `v37_s1_live_wiring_kat`, `v37_w4_settlement_test`,
  `v37_w4_owed_incremental_test`, `v37_1_activation_test`: 7/7 pass

**Zero consensus movement.** `src/sharechain/v37` tree object is
`fcc0bba339104fe819801093d29bb850112886ac` at `28fac44b`, at `a7e8450d` and at
`68fe18c6`; `git diff` master..head touches 0 files under it. `v37_lane.hpp`,
`w3_wire_freeze.hpp`, `settle_finalize_driver.hpp` and `record_log.hpp` blobs
are byte-identical to master.

Still **[DO NOT MERGE] DRAFT**.

Issue #1680, opened 2026-09-19
xmr native node: RESPONSE_FLUFFY_MISSING_TX dropped by P2P DoS block bucket strands a bodiless block, stalls node one block behind until restart


frstrtr:
### Summary
In the embedded XMR native node, `RESPONSE_FLUFFY_MISSING_TX` replies from a peer are dropped under the per-peer P2P DoS token bucket, which permanently strands a bodiless fluffy block and stalls the node one block behind until restart. Found while working on the good-citizen template path (PR #1678); it is **pre-existing** (outside that PR's diff) and does not affect consensus or block validity.

### Root cause
`xmr_p2p_dos.hpp` gives block-push frames a small token bucket (`block_capacity = 8`, `block_refill = 0.1/s`, tuned for "honest ~1 block per 120 s"). Every levin `2008` frame is counted against it. But `monerod`'s `RESPONSE_FLUFFY_MISSING_TX` (the reply to our missing-tx request) is *itself* a `2008` frame. At a fast block cadence (regtest, or any burst) the bucket drains, the missing-tx reply is dropped (`DosAction::Drop`, silently), and the bodiless fluffy park stays in `ChainIndex` `alt_` forever: `on_chain_entry` skips ids `alt_` already holds and `refetch_wanted()` filters by `have_body_locked_()`, so the block is never re-requested. The node sits one block behind and every later share submits at a taken height (`submit: last_error=rpc: Block not accepted`) until restart, after which the block connects fine via `GET_OBJECTS`.

### Reproduction
Reproduced twice at regtest cadence (stalled at h=93 and h=111), root-caused with local (reverted, uncommitted) instrumentation. Mainnet block cadence is far below the bucket, so this is unlikely on mainnet, but a lost reply is not recovered.

### Impact
- No empty-block-over-non-empty-pool and no invalid block — block validity and the good-citizen rule are unaffected.
- The node silently falls one block behind and rejects its own found shares until restarted.

### Suggested fix (either or both)
1. A dropped/lost missing-tx reply should be re-asked — time out the `NeedsBodies` park and re-issue the request (or fall back to `GET_OBJECTS`) rather than stranding it.
2. A `2009`-request's `2008`-reply should not spend a block token (only unsolicited block pushes should).

PR #1682, opened 2026-09-19
[DO NOT MERGE] xmr native: CLSAG ring-sig verify + global double-spend + reproducible monerod-parity harness

frstrtr:
## Input consensus, made load-bearing on the live node

CLSAG ring-signature verification and the global double-spend check (landed
earlier on this branch) were monerod-parity-exact **at the admission level** but
were **not load-bearing on a running node**: the producer never populated
`BlockTxEvent.outputs`, so every ring was `RingUnresolved`, and the default
select policy still mined a `RingUnresolved` (unverified-ring) transaction. A
forged-ring tx would still have been mined. This stack closes that.

### What changed (5 commits atop the parity harness)

1. **contracts** — `BlockTxEvent` gains the coinbase (amount, one-time key)
   pairs + unlock_time; `TxpoolSelectPolicy` gains `unresolved_rings`
   (Exclude default). Contracts-family amendment (types.hpp + txpool.hpp only).
2. **producer output capture** — `rct::zero_commit(amount)` (the v2 coinbase's
   `zeroCommit`, `zeroCommit(0)==G`); `outPk` masks captured into
   `out_commitments`; coinbase outputs captured; `block_eval` builds every v2
   body's outputs; `chain_view` is the single numbering authority (stamps
   `first_output_index`, advances/rolls back the counter, seeds from the
   anchor's `rct_output_count`). The chain layer stays free of the rct curve
   code, so the C2 chain KATs are untouched.
3. **authenticated output+spent set on the v37 MRR-Merkle/MMR model** — the
   ad-hoc in-memory `ChainOutputSet` is replaced by two append-only logs, one
   leaf per connected Monero block (its origin bin = the committed height),
   whose roots are the **shipped v37 lane MMR** (`::v37::Lane::mmr_*`, leaf =
   `sha256d(0x00||payload)`, byte-identical to `src/c2pool/v37/record_log.hpp`)
   — one MMR implementation, **zero edits under `src/sharechain/v37`**. Log-size
   membership proofs, a set digest for the anchor snapshot, reorg-safe bit-exact
   peak restore, and serialize/deserialize with re-derivation self-check.
4. **select policy** — a resolvable ring is verified-or-rejected at admission, so
   "require InputConsensus when the ring resolves" == "exclude the ring-
   unresolved from selection". A forged/unverifiable tx is now never
   selectable/mined; a valid tx whose ring resolves stays selectable (good
   citizen; blocks non-empty on a covered node). `excluded_unresolved` sensor.
5. **KATs + must-fix** — output-set KAT extended to **80 checks** (coinbase
   zeroCommit + ordering, RecordLog root parity on the set's own leaves,
   membership proof verify + flip-fail, bit-exact reorg restore, serialize
   round-trip, and the select-policy exclusion/good-citizen/Include cases). The
   false `--replay path` claim in the CLSAG KAT comment is corrected.

### Verification

Built on the `c2pool-build` self-hosted runner (gcc-13, `XMR_BUILD_RANDOMX=ON`,
-j20): the KAT targets **and the full `c2pool-v37-xmr` node binary** compile and
link. KATs green: `output_set` 80/0, `clsag_verify` 207/0, `txpool` 124/0,
`rct_verify` 130/0, `node` 88/0.

### Anchor-snapshot codec — OPERATOR SEAM (ruling R1)

The numbering-base seeding uses the **existing** `AnchorBundle::rct_output_count`
field (0 on a format-1 bundle), so it forces no format change and is implemented
here. Committing the output/spent **roots** into the anchor (so a daemonless
node boots the historical set instead of being blind to pre-anchor outputs) DOES
change the AnchorBundle format + digest, so it is left as a decision-ready seam,
not unilaterally applied. Decision-ready format-2 proposal (append after
`already_generated_coins`, writer emits it only when set, so every existing
format-1 `.inc` stays byte-identical):

```
rct_output_count       <u64>   # numbering base (already a field; now committed)
output_set_base_height <u64>   # 0 = full history
output_set_root        <64hex>
output_set_leaves      <u64>
spent_set_root         <64hex>
spent_set_leaves       <u64>
```

`anchor_self_check` gains `leaves == H_a + 1 - base_height`. Generator fills
`rct_output_count` from `get_output_distribution{amounts:[0],cumulative:true}`
and the roots from a genesis→H_a walk through `evaluate_block` + `ChainOutputSet`
(the same walk writes the backfill journal). Rejected alternative: an
unreferenced sidecar (no binding between anchor id and roots).

### Honest gaps

- **Live from-genesis regtest end-to-end proof is not yet run on this build.**
  The mechanism is proven at KAT level (forged/unresolved → `selectable=0`,
  valid → `selectable=1`, coinbase zeroCommit, reorg, MMR). Driving the running
  `c2pool-v37-xmr` binary against a real regtest monerod (genuine tx mined,
  forged-ring not mined, monerod `submit_block` accepts, double-spend not mined)
  is the remaining end-to-end step.
- **Mainnet O-backfill** depends on the anchor-snapshot ruling (R1): loading the
  pre-anchor output/spent set (walk-at-mint, journal distribution, mmap'd table
  + disk spent index, RAM budget) is a seam. Until then a mainnet anchor node's
  rings below the base are `RingUnresolved` (fail-closed).
- **Select-policy default (ruling R2):** Exclude is the recommended default
  (never mine an unverified ring). A mainnet node pre-backfill needs Include to
  produce non-empty blocks; the `--xmr-unresolved-rings` CLI flag is a small
  follow-up (the config field + Exclude default are in).
- **Committed in-tree regtest parity harness** (monerod-side mutant rejection +
  `submit_block` parity, option-gated + loud SKIP): a follow-up; the rig that
  records this parity exists out of tree.
- Built with gcc-13 (the runner's compiler), not gcc-15.


frstrtr:
Superseded by #1688.

This PR's tree is proven and load-bearing, but its commit-message trailers trip the required attribution-gate, and force-push is forbidden here. Per operator ruling, #1688 is a clean recommit of the exact same tree — the 8 commits (head 3d4c4b82) replayed onto current master with operator-signed, attribution-clean messages (verified), plus the two must-fixes folded in:

- CodeQL py/illegal-raise in gen_input_consensus_golden.py is fixed;
- the committed, option-gated monerod-side parity harness (§5b) is landed and live-green.

`git patch-id` of `3d4c4b82..#1688-head` equals master's own #1680 fix exactly, so every line of input-consensus code is byte-identical to what was reviewed here. attribution-gate is green on #1688.

Closing this in favour of #1688. Branch retained (no deletion).

PR #1689, opened 2026-09-19
[DO NOT MERGE] xmr native: anchor format-2 (output-set O-backfill) — ruling R1

frstrtr:
**Stacked on #1688** (base `v37/xmr-input-consensus-v2`). Do not merge before it.

Implements **operator ruling R1**: anchor **format-2** so a daemonless XMR node booting from a mainnet `AnchorBundle` can **O-backfill** the historical amount-0 output set and the spent-key-image set, and therefore resolve + CLSAG-verify a ring whose members are **below the anchor's numbering base**, and reject a below-base double-spend. On #1688 the CLSAG + double-spend verify is live-enforced on regtest via an authenticated output/spent set on the shipped v37 Lane MMR, but that set is built only from blocks connected since boot — a mainnet anchor node cannot resolve a ring whose members are pre-anchor. Format-2 fixes that by committing the set roots into the anchor.

### The format-2 fields
Appended after `already_generated_coins`, emitted **only when a set is present** (`has_output_set()`):

```
rct_output_count N
output_set_base_height N
output_set_leaves N
output_set_root <64hex>
spent_set_leaves N
spent_set_root <64hex>
```

`AnchorStatus::OutputSetShape` checks the committed set is one leaf per block over `base..H_a`, both logs equal length, non-zero frontier; format-1 bundles must carry all six as zero.

### Format-1 byte-identity (non-negotiable)
The writer emits the six lines *only* alongside a set, so a bundle without one produces **exactly today's bytes** — same `format 1` line, same digest. Proven by the anchor self-check KAT: the embedded stagenet `.inc` is untouched, the C++ canonical writer still reproduces its digest, and a new `test_format2` asserts a format-1 bundle writes `format 1` and none of the six lines. A format-1 reader parses a format-1 bundle bit-for-bit; the reader accepts `format 1|2`, refuses a format-2 key under a `format 1` line, and requires all six under `format 2`.

### Generator fill
`tools/xmr-anchor-gen/xmr_anchor_gen.py --output-set` walks genesis→H_a from the operator's own unrestricted monerod: `rct_output_count` from the walk, `(pubkey, commitment)` straight from `get_outs` (the coinbase zeroCommit is the daemon's own value — no ed25519 in Python), key images from the tx bodies. It builds the **same lane MMR and leaf discipline** the C++ `ChainOutputSet` uses and writes the roots plus the `serialize()` snapshot (`--output-set-out`). Its ported MMR + leaf encoding are **pinned byte-for-byte** to the C++ producer by the output-set KAT PART E selftest vector (`--selftest`).

### Load / seed path
`ChainBoot` keeps the loaded bundle; `ChainOutputSet::seed_from_snapshot` re-derives the operator snapshot and adopts it **only when it matches the anchor's committed tip / frontier / roots** — the anchor roots, not the blob, are the trust; it fails closed on any mismatch. The node seeds at boot behind `--native-output-set` (c2pool-v37-xmr) / `--output-set` (native node), so below-base offsets satisfy `resolve()` and CLSAG runs and the seeded spent set catches a below-base double-spend. A format-2 anchor **without** a snapshot boots fine with a loud status line and pre-anchor rings stay `RingUnresolved` (today's fail-closed).

### Young-chain window rule (the one semantic touch)
`anchor_expected_rows(H_a)` = `min(W, H_a+1)` per window (genesis included). At mainnet/stagenet heights this is exactly the pinned `735/100/100000`, so full bundles are unchanged; on a young chain it carries precisely what a genesis-booted `ConsensusState` holds, which is what makes minting a **regtest** anchor possible. Pinned by a from-genesis parity assertion in the anchor KAT.

### Tests (all green)
- **anchor self-check KAT 188/0** — format-1 byte-identity, format-2 round-trip, `OutputSetShape` cases, young-chain windows.
- **output-set KAT 150/0** — PART D: a seeded set resolves a below-base output, proves its membership against the committed root, catches a below-base double-spend, and fails closed on any root/count/tip mismatch; PART E: Python↔C++ leaf/MMR pin.
- txpool 124/0, node 88/0, clsag 207/0.

PART B (#1688) already proves that once an offset resolves to the authentic member the CLSAG verifies (Accepted+InputConsensus) and a wrong member is `RingSigFail`; PART D proves the seed is what makes a below-base offset resolve to that authentic member. Together they are the (a) below-base ring verifies / (b) forged below-base ring rejected / (c) below-base double-spend rejected proof at the consensus layer, deterministically and without a daemon.

### Constraints
- **0 files under `src/sharechain/v37`** (only `::v37::Lane` statics are called).
- Format-1 `.inc` byte-identical (presence rule; pinned by the embedded-stagenet KAT).
- BSD-only / stdlib (Python `hashlib`), no new crypto, no GPL/xmrig.

### Honest gaps
- **Mainnet mint + O-backfill test pends the operator's synced Mac** — `get_coinbase_tx_sum` / `get_outs` at scale need the operator's own unrestricted monerod.
- The live regtest **mint → fresh-node boot → below-base tx** integration run is the remaining cherry on top of the deterministic KAT proof above; status noted in the thread.
- Mainnet-scale disk-backed record table, peer O-backfill (walk 1..H_a to the same roots), and gate-4 network-confirm wiring are documented follow-on seams.

### Format-1 fail-closed (follow-up commit c776af2)
A format-1 bundle carries no rct_output_count (the six format-2 fields are all zero), so a node seeding from it would number the output set from base 0. With the input-consensus leg active an honest ring whose real member sits below the real (unknown) base collides with a post-anchor output, resolves to the WRONG member, and is RingSigFail'd with its peer scored a drop offender -- every real ring on the embedded stagenet format-1 anchor, until re-minted as format-2. This commit makes the format-1 path do what this PR body describes: on a BootMode::Anchor boot with no committed set, ChainOutputSet::disable_resolution() turns resolution off, so every ring stays RingUnresolved (fail-closed) and no honest peer is mis-scored. Format-2 anchors resolve and verify unchanged. Covered by output-set KAT PART F.


frstrtr:
## Live regtest evidence (vm905, own rig, fresh ports 49100/49101/49103)

Real regtest chain to height 321 (2 wallet transfers mined <260, so pre-anchor blocks carry non-coinbase outputs + key images), own monerod v0.18.5.1.

**1. Generator mints a real format-2 anchor from monerod** — the full RPC walk works end to end:
```
fetching headers 0..260
coinbase decomposition OK over 60 blocks
output-set walk 260/260 (rct_output_count=264)
wrote set_260.bin (40049 bytes) -- ChainOutputSet snapshot
wrote anchor_regtest_260.inc (7470 bytes), H_a=260
format 2 / rct_output_count 264 / output_set_base_height 1 / output_set_leaves 260
output_set_root c62082a5… / spent_set_leaves 260 / spent_set_root 67b1b9c3…
```

**2. Format-1 vs format-2 differ only by the format line + the six appended lines** — the format-1 control `.inc` is otherwise identical (presence rule, live).

**3. A fresh node boots from the format-2 anchor + snapshot and seeds it** (no genesis replay):
```
[output-set] seeded from format-2 anchor snapshot: base=0 frontier=264 outputs=264 spent=4 -- pre-anchor rings now RESOLVE
native: … verified_frontier=260
```
This also cross-validates the generator against C++: the node's `seed_from_snapshot` deserialized the Python-produced snapshot, re-derived the MMR with the C++ lane statics, and its roots matched the Python-committed anchor roots — otherwise it would fail closed. (Independently, the KAT PART E pins the Python leaf/MMR encoding to the C++ producer byte-for-byte.)

**4. Format-1 control boots unchanged** — no output-set line (a format-1 anchor commits no set).

**5. Fail-closed, live** — a snapshot tampered in the **leaf** region is refused and the node will not start:
```
REFUSED: output-set snapshot rejected: snapshot output-set root does not match the anchor's committed root
```

### Honest scope of the fail-closed guarantee (corrected from the live run)
Tampering the snapshot's **leaf** region is caught (root mismatch, above). Tampering only the **resolve table / flat spent set** while leaving every leaf intact is **not** caught: the `serialize()` form inherited from #1688 stores those tables separately from the leaves and carries no per-block `block_id`, so the anchor roots authenticate the snapshot's **leaf structure** (roots, leaf counts, tip, frontier), not the query tables byte-for-byte. The operator-supplied snapshot is therefore a **trusted input whose structure the anchor verifies** — bulk/random corruption always reaches a leaf and fails closed. Full trustless binding of the tables is the **peer O-backfill** follow-on (walk 1..H_a, re-derive to the same roots). Claim scoped precisely in `seed_from_snapshot` and here.

### Not yet done live (honest gaps)
- **Mainnet mint + O-backfill** pends the operator's synced Mac (`get_coinbase_tx_sum`/`get_outs` at scale need the operator's own unrestricted daemon).
- The **genuine below-base-ring tx → verified → mined → monerod-accepts** admission run (test a) needs the anchor-booted node's levin p2p to sync 261→321 and receive relayed txs (the #1688 live-proof machinery; peers=0 on this quick run). The **(a)/(b)/(c)** correctness is proven deterministically by the output-set KAT PART D (seed → below-base resolve + membership proof + double-spend caught + fail-closed) composed with PART B (#1688: authentic member ⇒ CLSAG passes, wrong member ⇒ RingSigFail, spent ⇒ KeyImageSpent).

Rig torn down after capture.

frstrtr:
Superseded by #1690: the identical proven format-2 tree re-committed onto current master (post-#1688), all three commits operator-Verified (this branch's tip `c776af21` was Unverified/unknown_key). Closing per no-force-push policy; branch retained, not deleted.

